PR #18, opened 2023-12-29
Add support for `var` globals

antocuni:
global variables are by default const and blue, but this PR makes it possible to declare them `var`, and thus they become red.

The hard part was to add support to it to the parser, see the long docstring inside `magic_py_parser.py`.


antocuni:
global variables are by default const and blue, but this PR makes it possible to declare them `var`, and thus they become red.

The hard part was to add support to it to the parser, see the long docstring inside `magic_py_parser.py`.

Issue #73, opened 2024-09-26
The `make.sh` script in `examples/` should use relative paths


b-long:
For example:

https://github.com/spylang/spy/blob/main/examples/jsffi/make.sh#L5

PR #74, opened 2024-09-26
Replace absolute path

b-long:
Hi @antocuni , awesome project! 🙇 🥇 🎉 

I'm wondering if you'd be willing to accept this PR?  I'd also wonder about other changes to this script, such as:

After:

```sh
emcc $CFLAGS raw_c_demo.c -o raw_c_demo.js
```

What if we add the following?

```sh
cd "$SCRIPT_DIR"
```

That way, the relative paths below the `emcc` would work as expected (I assume).

Fixes: https://github.com/spylang/spy/issues/73

### DIFF
--- a/examples/jsffi/make.sh
+++ b/examples/jsffi/make.sh
@@ -1,8 +1,13 @@
+#!/usr/bin/env bash
+
+# Based on: https://stackoverflow.com/a/246128
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
 set -e
 
 # use ./make.sh -O3 to enable optimizations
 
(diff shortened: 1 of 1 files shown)


antocuni:
> Hi @antocuni , awesome project! 🙇 🥇 🎉
> 
> I'm wondering if you'd be willing to accept this PR? 

sure, thank you!

> I'd also wonder about other changes to this script, such as:
> 
> After:
> 
> ```shell
> emcc $CFLAGS raw_c_demo.c -o raw_c_demo.js
> ```
> 
> What if we add the following?
> 
> ```shell
> cd "$SCRIPT_DIR"
> ```
> 
> That way, the relative paths below the `emcc` would work as expected (I assume).

ah yes, I think that so far I always run it from the `examples/jsffi` directory, but +1 to make it more robust

Issue #70, opened 2024-09-20
Guide How to Use Spy-lang


gunungpw:
Hello, very cool project.
I Hope this spylang can be python subset that prune and make the language more strict, also make it fast :)
Is there any documentation to use/build spylang in local machine (linux Ubuntu). I try to use it, but stuck when build the `libspy`.
Thankyou

antocuni:
Hello!
Sorry for the late answer, had too many things going on.
I'm totally aware that doc is lacking, and I'd actually like some help with it :).

On linux, the prerequisite are:
  - python >=3.10 (I'm testing with 3.10 but should work also with more recent versions)
  - [emscripten](https://emscripten.org/docs/getting_started/downloads.html#installation-instructions-using-the-emsdk-recommended) should be installed and `emcc` should be available

Then the basic setup should be something like this:
```
$ python -m venv venv --prompt spy
$ . venv/bin/activate
$ pip install -r requirements.txt
# make sure that the zig binary is available
$ ln -s $(realpath venv/lib/python3.*/site-packages/ziglang/zig) venv/bin/
$ make -C spy/libspy
```

Then, you should be good to go:
```
# execute the script in interpreted mode
$ ./spy.sh --run examples/hello.spy 
Hello world!

# compile to an executable
$ ./spy.sh -t native  examples/hello.spy
$ ./examples/hello
Hello world!

# see what happens after redshifting
$ ./spy.sh --redshift --no-pretty examples/hello.spy 
def main() -> `builtins::void`:
    `builtins::print_str`('Hello world!')

# run tests
$ pytest
```

This should be enough to get you started.
In case you are interesting in contributing, there are many small areas to improve which should be good as first time issues:
1. ideally the `zig` ln dance should not be needed and the makefile should be able to find `zig` by itself
2. currently `emcc` is a hard requirement but it shouldn't. If you don't want to target emscripten, it should not be necessary to have it installed
3. `spy/tests/test_cbuild.py::TestToolchain::test_c2wasm[clang]` currently fails if `clang` is not installed, but it should just be skipped in that case

antocuni:
I just realized that PR #76 contains a small guide to get started, so it might be what you are looking for.
Still, it would be nice to have "proper docs"

gunungpw:
Thanks for the answer,
Unfortunately, I don't have experience on build system and compiler.
But, I can help with documentation PR #78 address.

guangrei:
[...]

PR #76, opened 2024-10-03
Added one install and run example

aptakhin:
Hello @antocuni, thank you for the excellent talk in EuroPython. I just played around, recalling some of my compiler attempts. 

I drafted an example for installing, compiling, and running something simple with Spy. It might be useful for other people to start with, too.

### DIFF
--- a/EXAMPLE.md
+++ b/EXAMPLE.md
@@ -1,0 +1,65 @@
+Let's make install spy and generate something we can compile with C and run.
+
+Requires:
+- Emscripten
+- Zig
+- Fresh Python 3.12 possibly less, but Python 3.9 doesn't suit
+- gcc/clang
+
+For Mac OS:
+
+```bash
+brew install zig && brew install emscripten
+```
+
+Build native libraries for spy:
+
+```bash
+(cd spy/libspy && make TARGET= && make TARGET=native && make TARGET=wasi)
+```
+
+Setup one example:
+
+```
+cat > examples/generics.spy << EOF
+# filename: generics.spy
+
+@blue
+def make_fn(T):
+  def fn(x: T, y: T) -> T:
+    return x + y * 2
+  return fn
+
+fn_i32 = make_fn(i32)
+fn_f64 = make_fn(f64)
+
+def main() -> void:
+  if True:
+     print(fn_f64(1, 2))
+  else:
+     print(fn_f64(1, 3))
+EOF
+```
+
+Setup spy in the virtual env:
+
+```bash
+python -m venv .venv
+source .venv/bin/activate
+python -m ensurepip
+python -m install -r requirements.txt
+
+python setup.py install
+```
+
+Run generation:
+
+```bash
+python -m spy examples/generics.spy
+
+gcc -I`pwd`/spy/libspy/include -D SPY_TARGET_NATIVE \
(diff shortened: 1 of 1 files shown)


EXAMPLE.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+brew install zig && brew install emscripten
+```
+
+Build native libraries for spy:
+
+```bash
+(cd spy/libspy && make TARGET= && make TARGET=native && make TARGET=wasi)
+```
```
antocuni: I think that `make -C spy/libspy` should be enough, shouldn't it?
It automatically builds all the targets

EXAMPLE.md
```diff
@@ -0,0 +1,65 @@
+Let's make install spy and generate something we can compile with C and run.
+
+Requires:
+- Emscripten
+- Zig
```
antocuni: this is not really a prerequisite, since we are installing `ziglang` through `requirements.txt`.
The only annoying thing is that the `zig` executable is not in the `PATH` by default, see my comment here: https://github.com/spylang/spy/issues/70#issuecomment-2407473537

EXAMPLE.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+Run generation:
+
+```bash
+python -m spy examples/generics.spy
+
+gcc -I`pwd`/spy/libspy/include -D SPY_TARGET_NATIVE \
+  examples/generics.c `pwd`/spy/libspy/build/native/libspy.a
+
```
antocuni: this is not needed. You should be able to just do:
`python -m spy -t native examples/generics.spy`

In the meantime, I've also add an `hello.spy` example, so maybe you can also include it in this doc.

antocuni (changes requested):
Hello, thank you for the PR.
I've left some comments.

I think that once everything is setup, there are other interesting things to try. E.g.:

1. running a script in interpreter mode: `python -m spy --run hello.spy`
2. compile to a native exectutable: `python -m spy -t native hello.spy`
3. see the AST: `python -m spy --parse hello.spy`
4. see what happens after redshifting: `python -m spy --redshift hello.spy`
5. same as above but without pretty print, so make more evident some of the underlying structure: `python -m spy --redshift --no-pretty hello.spy`

EXAMPLE.md
aptakhin: Thank you! I fixed this

EXAMPLE.md
aptakhin: Nice! I didn't know. Fixed

EXAMPLE.md
aptakhin: Also added the solution

examples/generics.spy
```diff
@@ -0,0 +1,16 @@
...
+  return fn
+
+fn_i32 = make_fn(i32)
+fn_f64 = make_fn(f64)
+
+def main() -> void:
+  if True:
+     print(fn_f64(1, 2))
+  else:
+     print(fn_f64(1, 3))
```
antocuni: sorry for not seeing it earlier, but I suppose that the example makes more sense if you call `fn_32` in one branch and `fn_64` in the other branch

aptakhin: Oups, yes, it is. Fixed

aptakhin:
Hey @antocuni no worries! Thank you!

PR #128, opened 2025-02-26
Use Pyodide to drive Emscripten tests

hoodmane:
This PR makes it possible to run spy under pyodide.
The main highlights are:

1. adapt `llwasm` to work also on emscripten: in normal environment, `llwasm` uses `wasmtime` to load WASM binaries (in particular, `libspy.wasm`). On pyodie, `llwasm` loads WASM binaries by js-importing the corresponding emscripten `.mjs` module

2. make it possible to instantiate `SPyVM` async: this is needed under pyodide, where the functions to load WASM modules are async

3. adapt the CLI to run on pyodide


antocuni:
@hoodmane I rebased all the commits on top of `origin/main`

spy/tests/test_llwasm.py
```diff
@@ -1,32 +1,93 @@
...
+        #
+        # It is provided by pytest_pyodide and its valude is "node". Ideally,
+        # we would like to request it ONLY when llwasm_backend=='pyodide', but
+        # we couldn't find a way to do that.
+        #
+        # If we don't require "runtime" in the function signature, the
+        # "getfixturevalue('selenium')" below fails with the following
+        # message:
+        # The requested fixture has no parameter defined for test:
+        #     spy/tests/test_llwasm.py::TestLLWasm::test_call[pyodide]
```
antocuni, line 41: @hoodmane do you have an idea what causes this?
For more context, `init_llwasm` is a fixture which is automatically used by all tests in `TestLLWasm`. Ideally, for each of these tests I'd like two versions:
1. `test_xxx[wasmtime]`
2. `test_xxx[node-pyodide]`

Note that the `runtime` fixture is completely unused in the `[wasmtime]` case, but try to comment it out and see the error that you get.

It's not a big deal, but it's a bit annoying because now I get the `[node-*]` prefix also in wasmtime tests, which looks a bit confusing:
  - `test_xxx[node-wasmtime]`

spy/cbuild.py
```diff
@@ -222,22 +226,25 @@ def CC(self) -> list[str]:
...
 
     def c2exe(self, file_c: py.path.local, file_exe: py.path.local, *,
               opt_level: int,
               debug_symbols: bool,
               ) -> py.path.local:
-
+        post_js = spy.libspy.SRC.join('emscripten_extern_post.js')
+        EXTRA_LDFLAGS = [
+            f"--extern-post-js={post_js}",
+        ]
```
antocuni, line 242: @hoodmane I need your emscripten expertise here.
In our work together we removed these two LDFLAGS:
  - `-sEXPORTED_FUNCTIONS=['_main']"`
  - `--extern-post-js={post_js}`

Now, I had to reintroduce `--extern-post-js` again in order to make `test_c2exe` to pass:
https://github.com/hoodmane/spy/blob/8b7f91c844c7f3f957c523ace21001c5293d6f4a/spy/tests/test_cbuild.py#L22-L42

The content of post_js is this:
```
export let emscriptenModule = Module();
```

In other words, this seems to be needed to tell emscripten to actually start the `main()` function when we execute the output using node.

Is it correct?

PR #130, opened 2025-02-28
Add Dockerfile

guangrei:
Docker can make spylang easy to install and use, see https://github.com/guangrei/spy_docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM python:3.12
+
+RUN apt -qq update && \
+    apt-get install -y wget cmake ninja-build make autoconf autogen automake libtool expect && \
+    rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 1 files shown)


antocuni:
Thank you for the idea.
Honestly, I'm not sure that the spy main repo is the best place where to keep the dockerfile though, for various reasons:
1. it increases maintenance burden
2. we should setup building the image in CI and test that spy actually works
3. most of all, I don't think it's actually necessary.

The current instructions involve a lot of manual steps because I didn't put too much effort in making the installation process more straightforward, but eventually it should be possible to just do `pip install spylang` and have it working out of the box.

So, I'm closing this PR without merging.
That said, if you are willing to maintain a dockerfile which is regularly updated and tested, I'll be happy to link it from the README. Thanks!

guangrei:
@antocuni thanks for your explanation, spylang working out of the box with pip installation is great news but in some case docker could be an alternative.

so i will continue to maintain my docker image and move to workflows & github packages for transparency and automation.

glad if you can link it in spylang README 🙏

PR #170, opened 2025-05-20
chore: Enable MacOS tests on CI github workflow.

aisipos:
Add macos-latest as a platform the github tests workflow runs on. 

See here for a list of runners:
https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,14 +16,16 @@ on:
         required: false
         default: false
 
-
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


aisipos:
This is working now, we get separately named workflows for all combinations of OS x Python, although right now we just test on Python 3.12: 
image

antocuni:
that's wonderful! Thank you!

PR #172, opened 2025-06-04
Add dict to stdlib (WIP)

dpdani:


stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+        index = array1d[i32](1 << log_size)
+        # index = array1d[i32](2 ** log_size)  # [Antonio]: missing ** operator
+        # also, the error was super confusing:
+        #   ...
+        #     w_OP = OP_from_token(binop.op) # e.g., w_ADD, w_MUL, etc.
+        #   File "/Users/dp/repos/spy/spy/vm/modules/operator/__init__.py", line 108, in OP_from_token
+        #     return _from_token[token]
+        #            ~~~~~~~~~~~^^^^^^^
+        # KeyError: '**'
+        data.index = index
```
dpdani: So I guess I need to change a pointer here, but how do I express that in spy?
Python certainly doesn't allow that, so I don't know the syntax off the top of my head

stdlib/array.spy
```diff
@@ -0,0 +1 @@
+../examples/multifile/array.spy
```
antocuni: I think it's fine to copy `array.spy` to stdlib for now.

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
+from array import array1d
+# from hash import hash  # [Antonio]: I feel like this is supposed to work but doesn't
```
antocuni: yes, it should work, but you called it `hash.py` instead of `hash.spy` ;).
That's actually a good point, we should probably detect this common mistake and give a nicer error.
https://github.com/spylang/spy/issues/173

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+    def hash_i32(v: i32) -> i32:
+        # this mimics the implementation of CPython's hash(int)
+        return v
+    if T is i32:
+        return hash_i32
+
+    raise TypeError("unsupported type for hash()")
+
+
+DKIX_EMPTY: i32 = 0
```
antocuni: note that here `: i32` is not needed: the assignment operator automatically perform type inference in this case.

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+# DKIX_DUMMY: i32 = 2 ** 32 - 1
+DKIX_DUMMY: i32 = 1 << 32 - 1
+# DKIX_ERROR: i32 = 2 ** 32 - 2
+DKIX_ERROR: i32 = 1 << 32 - 2
+# DKIX_KEY_CHANGED: i32 = 2 ** 32 - 3
+DKIX_KEY_CHANGED: i32 = 1 << 32 - 3
+
+MIN_LOG_SIZE: i32 = 6
+MAX_LOG_SIZE: i32 = 31
+MAX_FILL_RATIO: float = 2 / 3  # [Antonio]: should I use something other than float?
```
antocuni: `float` is an alias to `f64` (and `int` is an alias to `i32` for now).
I think that if you use `i32` everywhere, it's probably better to use `f64` here (but again, it's not necessary to add type annotation at all)

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+    @struct
+    class Entry:
+        # empty: bool  # [Antonio]: a super-confusing error here:
+        #   ...
+        #   File "/Users/dp/repos/spy/spy/vm/modules/unsafe/struct.py", line 42, in calc_layout
+        #     field_size = sizeof(w_type)
+        #   File "/Users/dp/repos/spy/spy/vm/modules/unsafe/misc.py", line 22, in sizeof
+        #     assert False
+        #            ^^^^^
+        # AssertionError
```
antocuni: good point!
https://github.com/spylang/spy/issues/174

stdlib/dict.spy
antocuni: you should image `@typelif` as a C typedef. In this case `array1d[i32]` is a "typedef" to `ptr[ArrayData]`, so it's _already_ a pointer, you can just `array1d[Entry]`, I think.

If the question is "how do I *dereference* a pointer", the answer is to use `[0]`. So, if you have:
```
p: ptr[i32] = gc_alloc(...)
p[0] = 42   # equivalent to *p = 42
print(p[0])
```

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+        log_size: i32  # capacity
+        length: i32  # number of items stored
+        entries: ptr[array1d[Entry]]
+
+
+    def _new_data(log_size: i32) -> DictData:
+        # assert MIN_LOG_SIZE <= log_size <= MAX_LOG_SIZE
+        data = gc_alloc(DictData)(1)
+        index = array1d[i32](1 << log_size)
+        # index = array1d[i32](2 ** log_size)  # [Antonio]: missing ** operator
```
antocuni: thanks
https://github.com/spylang/spy/issues/175

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+        # also, the error was super confusing:
+        #   ...
+        #     w_OP = OP_from_token(binop.op) # e.g., w_ADD, w_MUL, etc.
+        #   File "/Users/dp/repos/spy/spy/vm/modules/operator/__init__.py", line 108, in OP_from_token
+        #     return _from_token[token]
+        #            ~~~~~~~~~~~^^^^^^^
+        # KeyError: '**'
```
antocuni: https://github.com/spylang/spy/issues/176

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+        data.index = index
+        # [Antonio]: ^ unsure what to do about this one:
+        # expected `unsafe::ptr[array::array1d[i32]::ndarray]`, got `array::array1d[i32]::ndarray`
+        entries = array1d[Entry](1 << log_size)
```
antocuni: this is exactly what it says: you are trying to mix a type `T` with `ptr[T]` (where T in this case is `array::array1d[i32]::ndarray`, see also my comment above.

If you have suggestions for how to emit a better error message, I'm all ears.

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+        data.entries = entries
+        data.log_size = 1 << log_size
+        data.length = 0
+        # [Antonio]: do I need to initialize with zeros here, or is it handled
+        # transparently?
```
antocuni: `@struct`s allocated via `unsafe::gc_alloc` are not initialized, so you should do it by yourself.

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+                    # continue # [Antonio]: no continue? :(
+                    pass
+                else:
+                    entry = self.__get_entry_at(ix)
+                    if not entry.empty:
+                        # continue (I'd rather remove the `not` and use `continue` here)
+
+                        # [Antonio]: it's nice I can actually bubble up exceptions here
+                        # so that effectively I don't need DKIX_ERROR, unlike CPython's
+                        # dict.
```
antocuni: I'm sure what you mean here.
Note that currently exceptions are not really implemented, because `raise` always panics

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+                        hash_entry = hash(entry.key)  # might raise exception
+                        hash_key = hash(key)  # might raise exception
```
antocuni: note that you will need `hash[Key](entry.key)` (maybe sooner or later SPy will be able to deduce it by itself, but not now).

Also, I don't think that it can raise: `hash[Key]` can raise at _blue time_, but if we arrive at runtime then the actual call is guaranteed not to raise (another nice property of SPy :)).

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+            ix = self.__lookup(key)
+            if ix == DKIX_EMPTY:
+                raise KeyError(key)
+            return data.entries[ix].value
+
+        def __insert(self: _dict, key: Key, value: Value) -> void:
+            # [Antonio]: I guess here we might like stronger guarantees for private
+            # methods than Python allows? If a user calls this method incorrectly,
+            # they will corrupt the dict state.
+            # I guess CPython does have actually-private methods (in C).
```
antocuni: yes, that's something that we need to think about.
In the meantime, if you want to hide the method you can just turn it into a function Something like this:

```
@blue.generic
def dict(Key, Value):
    @struct
    class DictData:
        ...

    def insert_entry(data: ptr[DictData], key: Key, value: Value) -> void:
        ...

    class _dict:
        def __setitem__(self: _dict, ...):
            data = self.__ll__
            ...
            insert_entry(data, key, value)
```

stdlib/dict.spy
```diff
@@ -0,0 +1,202 @@
...
+            data: DictData = self.__ll__
+            entry = data.entries[data.length]
+            # data.length += 1  # [Antonio]: ParseError: not implemented yet: assign to complex expressions :(
```
antocuni: at least this one gives a nice message :)
https://github.com/spylang/spy/issues/177

stdlib/hash.py
```diff
@@ -0,0 +1,9 @@
+@blue.generic
+def hash(T):
+    def hash_i32(v: i32):
+        # this mimics the implementation of CPython's hash(int)
+        return v
+    if T is i32:
+        return hash_i32
```
antocuni: The `is` operator is not implemented yet, but you can use `T == i32`

stdlib/test_dict.spy
```diff
@@ -0,0 +1,26 @@
+from dict import dict
+
+
+def test() -> void:
+    # d = dict[int, int]()
+    d = dict[i32, i32]()
+    d[1] = 1
+    # assert d[1] == 1  # [Antonio]: ParseError: not implemented yet: Assert :(
```
antocuni: https://github.com/spylang/spy/issues/179

stdlib/test_dict.spy
```diff
@@ -0,0 +1,26 @@
...
+    else:
+        print("❌")
+
+# [Antonio]: have also seen this error.
+# it's quite confusing not to have a reference to my spy file and
+# having just "<string>" instead :/
+#   File "<string>", line 22
+#                 value: Value
+#                             ^
+# TabError: inconsistent use of tabs and spaces in indentation
```
antocuni: uh, this shouldn't happen, indeed. I cannot image where the `<string>` comes from, because IIRC I don't use `exec` or `eval` anywhere.
Can you show me a way to reproduce?

stdlib/test_dict.spy
```diff
@@ -0,0 +1,26 @@
...
+
+# [Antonio]: it's quite annoying in general when I get ParseErrors one at a
+# time. I'd rather see them all and resolve them in one shot.
```
antocuni: that's a fair point, but not easy to solve due to the way the current parser is done (which is: first get the CPython ast and then convert it to `spy.ast`).
Eventually we will write our own parser in spy, but that's low priority.

antocuni (commented):
Thank you a lot for this PR, I love it.

I especially appreciate that you highlighted all the various "pain point": I agree we should do better, and I opened an issue for each of them (some are easy to fix/implement, others are harder).

stdlib/dict.spy
dpdani: ahh ok. than this changes things quite a bit.

but does it not raise even if my function is something dumb like this?
```python
def hash(v):
    raise Exception
```

as in, this would always cause a compilation error? or always a panic?

stdlib/test_dict.spy
dpdani: sure:
1. open the file in a simple editor (in PyCharm the tab character gets replaced by spaces no matter what I do)
2. replace four white spaces on line 22 of dict.spy with a tab character
3. `spy dict.spy` shows the error

antocuni: ok, I managed to reproduce and created an issue:
https://github.com/spylang/spy/issues/180

stdlib/dict.spy
antocuni: I must admit that some of these details are still not fully decided yet, but I *think* I'm leaning towards the following rules (inspired by [this essay](https://joeduffyblog.com/2016/02/07/the-error-model/)):

1. There are runtime exceptions (which can be caught in theory, although `except` has not been implemented yet) and panics (which cannot be caught)
2. Exceptions are for those cases in which an operation can fail even if it's not a bug (e.g., `FileNoutFoundError`)
3. panics are for "bugs"
4. `TypeError`s are panics, because they are "bugs". Also, normally they should be raises during redshift, so I don't expect them survive until runtime anyway.

That said, I'm open to modify/tweak them after we start to actually use the language and see which patterns make sense and which don't.

For the specific case of `hash`, I expect the `TypeError` to happen in the "blue" part, e.g.:

```
@blue.generic
def hash(T):
    if T == i32:
        return ...
    raise TypeError(...)

@blue.generic
def dict(Key, Value):
    def getitem(...):
        ...
        h = hash[Key](x)
```

This could be rewritten in this way:

```
@blue.generic
def dict(Key, Value):
    hash_key = hash[Key] # might raise TypeError

    def getitem(...):
        ...
        h = hash_key(x)
```

If you write like in the second example, it becomes very clear that `hash_key` cannot raise at runtime.

dpdani: I haven't had the time to checkout the article yet, but it looks super interesting, will definitely read it 🙏

I think the error model you have described is sound. It makes a lot of sense to me.
It's one of those things that will need to be taught to newcomers.

stdlib/dict.spy
dpdani: I think I understand better what I was doing wrong now.

Tell me if this sounds consistent to you:
1. I was "typedefing" a _dict as a pointer to DictData
2. I wanted to swap DictData when doing a resize
3. I can't do it because I can't change the thing that's holding the ptr[DictData]
4. _dict must be a typedef for a struct that holds a ptr[DictData], so that later I can swap the two structs

I got this after reading this comment: https://github.com/spylang/spy/pull/172#discussion_r2131813701
Thinking in terms of C constructs definitely helped 🙏

stdlib/dict.spy
```diff
@@ -0,0 +1,187 @@
...
+    def insert(self: _dict, key: Key, value: Value) -> void:
+        data: DictData = self.__ll__.data[0]
+        # [Antonio]: another confusing error
+        #   File "/Users/dp/repos/spy/spy/vm/function.py", line 265, in raw_call
+        #     w_res = self._pyfunc(vm, *args_w)
+        #   File "/Users/dp/repos/spy/spy/vm/modules/unsafe/mem.py", line 47, in w_mem_read_T
+        #     assert False
+        #            ^^^^^
+        # AssertionError
+        # does this mean that structs cannot be dereferenced?
```
antocuni: thanks to PR #182 and the additional commit https://github.com/spylang/spy/commit/52ae2dde69f508e85b3780dbfeb9c1d96a92543e this now results in a better error (after you rebase your branch to `main`):
```
WIP: Cannot read struct by value: `dict::dict[i32, i32]::DictData`
   --> /home/antocuni/anaconda/spy/stdlib/dict.spy:102:26
102 |         data: DictData = self.__ll__.data[0]
    |                          |_________________| called from here
```

This is another instance of confusing structs vs ptr-to-structs. In C terms, what you are trying to do is this:
```
typedef struct { ... } DictData;
typedef struct {
    DictData *data;
} _dict;
typedef struct {
   __ll__ *_dict;
} __dict;

__dict self;
DictData data = self.__ll__.data[0]; // or *(self.__ll__.data)
```

In C, this is permitted but it probably doesn't do what you want, because it *copies* the content of the struct into the tmp variable `data`. In SPy, eventually we will probably support this, but at the moment copying struct-by-value is not supported.

antocuni: If you just want to have an easy access to `data` for shorter typing, you can this (which I guess it's what you originally wanted):
```
data: ptr[DictData] = self.__ll__.data
```

stdlib/_dict.spy
```diff
@@ -0,0 +1,187 @@
...
+
+    @struct
+    class __dict:
+        data: ptr[DictData]
+
+
+    @typelift
+    class _dict:
+        __ll__: ptr[__dict]
+
```
antocuni, line 208: Is this double indirection actually needed?
I seem to understand that with the current system, when you do a resize you allocate a new `DictData` and swap the pointer to it.
But I suppose you could achieve the same effect by keeping the same `DictData` and just update its `.index` and `.entries` arrays?

I *think* this is more similar to how things actually work in CPython, but I admit that I didn't look at the code too closely and I might be mistaken

stdlib/dict.spy
```diff
@@ -0,0 +1,187 @@
...
+        distance = -1
+        while distance < capacity(self):
+            distance += 1
+            ix = data.index[(d0 + distance) & capacity(self)]
+            # [Antonio]: another confusing error:
+            #   File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/fixedint/base.py", line 391, in _f
+            #     return nt(intfunc(int(self), int(other)))
+            #               ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
+            # ZeroDivisionError: integer modulo by zero
+            # I guess this is saying that capacity() might return 0?
```
antocuni: I think it says that someone tried to do `x % 0` (not that it *might* do modulo 0... someone actually DID that).
You get the same tb with this toy program:
```
def main() -> void:
    print(2 % 0)
```

We should detect the case and raise a proper exception/panic, see #183.

That said, I cannot really understand where it comes from, because I don't see any modulo operator in your source code, and I cannot reproduce it.
Can you got this error with a different/previous version of the code?
Or, can you show me how to reproduce?

dpdani: Sure! If you substitute `&` → `%` on line 74 you should see that error

stdlib/_dict.spy
dpdani: I'll remove that 👍

stdlib/dict.spy
```diff
@@ -0,0 +1,196 @@
...
+        entries = gc_alloc(Entry)(1 << log_size)
+        i = 0
+        while i < 1 << log_size:
+            entries[i].empty = True
+            # ^ WIP: Cannot read struct by value: `dict::dict[i32, i32]::Entry`
+            # [Antonio]: not sure what to do here, I'm leaning towards removing
+            # the Entry struct in favor of having three different arrays.
+            # It degrades cache friendliness, but that's fine.
+            # Do you think there's a way around this WIP error now?
+            # I also tried blueing the function, but it didn't help.
```
dpdani: @antocuni what do you think about this one? Should I just split the struct array into two arrays? Or is there a way around it?

antocuni: I think you hit another missing case which I didn't think to implement :).
Yes, this code is _supposed_ to work, but clearly it doesn't.
I opened a new issue: https://github.com/spylang/spy/issues/184

I'm on vacation in these days (greetings from Puglia :beach_umbrella: ) so probably the easiest thing to unlock you is to use separate arrays for now, and switch back to `Entry` when the issue is fixed... sorry for that.

And btw, the work you are doing is _extremely_ helpful in finding concrete pain points of SPy, thank you a lot for doing that :pray:

dpdani: Buone burrate! 🏖️

antocuni:
@dpdani I have some news which should be relevant to you:
1. I merged #195: this should make it possible to use an array of `Entry` as you wanted to do
2. I marged #187: it officially introduces the `stdlib` directory and you can now seamlessly import modules from it
3. I have a WIP #196: I think the most interesting bit for you is to see how I write tests for `array.spy`, which you could use as inspiration for testing `dict.spy`. As you can see currently everything is very explicit, although eventually I would like to have something more straightforward, so that people don't have to call `self.compile()` manually... but I suppose that for now it's good enough

antocuni:
> hey there @antocuni 👋
> 
> big merge from main coming through 🚂 along with tests (shamelessly written by Junie) and array of `Entry`
> 
> I'm seeing this error in all doppler and C tests, but not in the interpreter tests. How do I deal with it?
> 
> ```
> E       spy.errors.SPyError: TypeError: type `unsafe::ptr[dict::dict[i32, i32]::DictData]` does not support assignment to attribute 'data'
> E          --> /Users/dp/repos/spy/stdlib/dict.spy:161:9
> E       161 |         self.__ll__.data = new
> E           |         |_________| this is `unsafe::ptr[dict::dict[i32, i32]::DictData]`
> E       
> E          --> /Users/dp/repos/spy/stdlib/dict.spy:161:9
> E       161 |         self.__ll__.data = new
> E           |         |____________________| operator::SETATTR called here
> ```

this is actually SPy's static typing doing its job :)

In the method `resize`, `self.__ll__` is of type `DictData`, which does **not** have a `data` field:
```
    @struct
    class DictData:
        index: ptr[i32]
        log_size: i32  # capacity
        length: i32  # number of items stored
        entries: ptr[Entry]
```
So, the error you get it's actually correct.

Why the `[interp]` tests pass then? Because in interp mode, type errors are raised only if/when we actually try to execute that line (like in Python): evidently the method `resize` is never called by any test.

However, when we do redshift we check ALL functions for TypeErrors, and that's why compilation fails. This is called "eager error mode", and it's the default (because it's what you expect in a statically typed language, after all).

There is also the "lazy error mode", in which `TypeError`s don't cause compilation to fail, but they are translated into `raise` statements.

E.g., consider this program:
```python
def main() -> None:
    print('hello')
    if 0:
        1 + 'xxx'
    print('world')
```

```

❯ spy /tmp/x.spy 
hello
world

❯ spy --redshift /tmp/x.spy    # eager errors
TypeError: cannot do `i32` + `str`
   --> /tmp/x.spy:4:9
  4 |         1 + 'xxx'
    |         ^ this is `i32`

   --> /tmp/x.spy:4:13
  4 |         1 + 'xxx'
    |             |___| this is `str`

   --> /tmp/x.spy:4:9
  4 |         1 + 'xxx'
    |         |_______| operator::ADD called here


❯ spy --redshift --error-mode=lazy /tmp/x.spy 

def main() -> None:
    print_str('hello')
    if `operator::i32_to_bool`(0):
        raise TypeError('cannot do `i32` + `str`') # /.../astframe.py:583
    print_str('world')
```

In tests, you can turn on lazy errors by using `self.compile(src, error_mode='lazy')`, and they pass.
However, the right thing to do is to fix `resize`, of course :)

stdlib/dict.spy
dpdani: three arrays merged back into an array of structs 👍

dpdani:
Agreed 👍

I think I want to tackle deletes and the iterator in this PR, and the rest can be separated.
What do you think?

dpdani:
I'm happy with the state of this PR. 

@antocuni want to do a review?

stdlib/dict.spy
```diff
@@ -0,0 +1,283 @@
...
+
+    @struct
+    class DictData:
+        index: ptr[i32]
+        log_size: i32  # capacity
+        length: i32  # number of items stored
+        entries: ptr[Entry]
+
+    def new_index(log_size: i32) -> ptr[i32]:
+        # assert MIN_LOG_SIZE <= log_size <= MAX_LOG_SIZE
```
antocuni: thanks to #240 we now have `assert`

stdlib/dict.spy
```diff
@@ -0,0 +1,283 @@
...
+        def __getitem__(self: _dict, key: Key) -> Value:
+            data: ptr[DictData] = self.__ll__
+            ix = lookup(data, key)
+            if ix == DKIX_EMPTY:
+                # raise KeyError(key)
+                # [Antonio]: no KeyError? :(
+                # NameError: name `KeyError` is not defined
+                #    --> /Users/dp/repos/spy/stdlib/dict.spy:184:23
+                # 184 |                 raise KeyError(key)
+                #     |                       |______| not found in this scope
```
antocuni: you can add `KeyError` here:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/spy/vm/exc.py#L127-L174

stdlib/_dict.spy
```diff
@@ -0,0 +1,283 @@
...
+                #     |                       |______| not found in this scope
+                raise Exception("KeyError")
+            return data.entries[ix].value
+
+        # def get(self, key: Key, default: Value = None) -> Value | None:
+        # [Antonio]: no default args? :(
+        # E       spy.errors.SPyError: ParseError: default arguments are not supported yet
+        # E          --> /Users/dp/repos/spy/stdlib/dict.spy:191:50
+        # E       191 |         def get(self, key: Key, default: Value = None) -> Value | None:
+        # E           |                                                  |__| this is not supported
```
antocuni: yes, this is a current limitation of SPy.
You can emulate them with metafunctions, like I do here for `range`:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/stdlib/_range.spy#L9-L34

However, in this case we have an additional problem, because SPy doesn't support `T | None` yet.
So, I suppose that for now it's fine to just declare that you must call `.get` with two arguments.

stdlib/_dict.spy
```diff
@@ -0,0 +1,283 @@
...
+        def get(self: _dict, key: Key, default_: Value) -> Value:
+            # _ in default_: see https://github.com/spylang/spy/issues/246
+            # [Antonio]: no try-except? :(
+            # E       spy.errors.SPyError: ParseError: not implemented yet: Try
+            # E          --> /Users/dp/repos/spy/stdlib/dict.spy:198:13
+            # E       198 |             try:
+            # E           |             |________________| this is not supported
```
antocuni: yes, exceptions don't work yet 🤷‍♂️ 
Sorry :)

stdlib/_dict.spy
```diff
@@ -0,0 +1,283 @@
...
+        def __delitem__(self: _dict, key: Key) -> None:
+            data: ptr[DictData] = self.__ll__
+            delete(data, key)
+
+        def __len__(self: _dict) -> i32:
+            data: ptr[DictData] = self.__ll__
+            return data.length
+
+        def __fastiter__(self: _dict) -> dict_iterator:
+            return dict_iterator(self.__ll__, 0)
```
antocuni, line 287: can you also add a `keys()` method while you are at it? It should just work out of the box, shouldn't it?

stdlib/_dict.spy
```diff
@@ -0,0 +1,283 @@
...
+        i: i32
+
+        def __next__(self: dict_iterator) -> dict_iterator:
+            return dict_iterator(self.data, self.i + 1)
+
+        def __item__(self: dict_iterator) -> Key:
+            return self.data.entries[self.i].key
+
+        def __continue_iteration__(self: dict_iterator) -> bool:
+            return self.i <= self.data.length
```
antocuni, line 203: I think this is good enough for now. However, eventually we probably want to introduce the same check as in CPython and detect whether the dict has changed size during the iteration.
So maybe we can add a `# TODO` comment?

stdlib/hash.spy
```diff
@@ -0,0 +1,24 @@
+@blue.generic
+def hash(T):
```
antocuni: we have a better way to deal with `hash` nowadays 🎉.
Look how we implement `len`:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/spy/vm/modules/builtins.py#L112-L125

We should do the same for `hash`.
We can do it in this PR, but I'm also happy to do it in a follow-up PR if you prefer.

spy/tests/stdlib/test_dict.py
```diff
@@ -0,0 +1,226 @@
...
+        src = """
+        from dict import dict
+
+        def test() -> int:
+            d = dict[i32, i32]()
+            return d[99]
+        """
+        mod = self.compile(src)
+        with pytest.raises(Exception, match="KeyError"):
+            mod.test()
```
antocuni, line 60: you should use `SPyError.raises`, see e.g. here:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/spy/tests/stdlib/test__list.py#L67-L78

stdlib/dict.spy
dpdani: but the chained comparison doesn't work, and using `and` doesn't work either:

```
self = <spy.parser.Parser object at 0x106fbcf50>
primary = 'not implemented yet: BoolOp', secondary = 'this is not supported'
loc = <Loc: '/Users/dp/repos/spy/stdlib/_dict.spy 37:15 37:68'>

    def error(self, primary: str, secondary: str, loc: Loc) -> NoReturn:
>       raise SPyError.simple("W_ParseError", primary, secondary, loc)
E       spy.errors.SPyError: ParseError: not implemented yet: BoolOp
E          --> /Users/dp/repos/spy/stdlib/_dict.spy:37:16
E        37 |         assert MIN_LOG_SIZE <= log_size and log_size <= MAX_LOG_SIZE
E           |                |___________________________________________________| this is not supported
```

I'll split it into two asserts

stdlib/dict.spy
dpdani: I don't quite see a way around this error 🤔
I guess, I'll just leave it as `raise KeyError()` without an argument?

```text
E               spy.errors.SPyError: TypeError: cannot call blue function with red arguments
E                  --> /Users/dp/repos/spy/stdlib/_dict.spy:161:28
E               161 |             raise KeyError(str(key))
E                   |                            |______| this is red
E               
E                  --> /Users/dp/repos/spy/spy/vm/exc.py:67:1
E                67 |         def w_new(vm: 'SPyVM', w_cls: W_Type, w_message: W_Str) -> T:
E                   | |___________________________________________________________________| function defined here
E               
E                  --> /Users/dp/repos/spy/stdlib/_dict.spy:161:19
E               161 |             raise KeyError(str(key))
E                   |                   |________________| operator::CALL called here
```

stdlib/_dict.spy
dpdani: np 😉

stdlib/_dict.spy
dpdani: sure 👍

stdlib/dict.spy
dpdani: By which I mean `raise KeyError("")` with an empty string because an argument is required

dpdani:
I'll leave the `hash` rework and the design comment to tomorrow

stdlib/dict.spy
antocuni: oh right, I always forget about missing features.
I opened #248 and #249 so we don't forget.
Feel free to open issues whenever you find missing features.

stdlib/dict.spy
antocuni: yes, current support for exceptions is VERY limited:
1. `raise` always causes a panic and exceptions cannot be caught
2. the error message must be `blue` (this is the error that you are seeing).

Note that you don't need `raise KeyError("")`. You can just do `raise KeyError`

antocuni: > because an argument is required

why required?
The following example works for me:
```
def main() -> None:
    raise KeyError
```

```
❯ spy /tmp/x.spy 
KeyError: 
   --> /tmp/x.spy:2:2
  2 |     raise KeyError
    |  |_______________|
```

dpdani: mm right, but if I write `raise KeyError()` with the parens then it doesn't work.

antocuni: right, opened #250 for it

spy/vm/modules/builtins.py
```diff
@@ -141,6 +141,56 @@ def w_repr(vm: 'SPyVM', wam_obj: W_MetaArg) -> W_OpSpec:
         wam_obj.loc
     )
 
+@BUILTINS.builtin_func
+def w_hash_i8(vm: 'SPyVM', w_x: W_I8) -> W_I32:
```
antocuni, line 145: this is good for now, but ideally I'd like these functions to be in `W_I8.w_hash` etc, similarly to what we do for `__str__`, `__repr__` and `__len__`.
It's not a merge blocker though, we can do it in a follow up PR

Issue #180, opened 2025-06-09
Don't crash if Python's parser cannot parse the file


antocuni:
At the moment, SPy's parser is based on CPython's one:
1. first CPython converts source code into an `ast` tree
2. then `spy/parser.py` converts `ast` into `spy.ast`

If there is an error and/or an unsupported construct in phase (2), we emit a nice error message, e.g.:

```
def main() -> void:
    a.x += 1
```

```
❯ spy -p /tmp/x.spy 
ParseError: not implemented yet: assign to complex expressions
   --> /tmp/x.spy:2:5
  2 |     a.x += 1
    |     |_| this is not supported
```

However, if there is a parsing error during phase (1), the interpreter crashes with a Python-level traceback, and moreover we show either `<string>` or `<unknown>` as the source code location.
E.g.:

```
def main() -> void:
    if 1:
```

```
❯ spy -p /tmp/x.spy
Traceback (most recent call last):
  File "/home/antocuni/anaconda/spy/venv/bin/spy", line 8, in <module>
    sys.exit(app())
[...]

  File "/home/antocuni/anaconda/spy/spy/magic_py_parse.py", line 51, in magic_py_parse
    py_mod = py_ast.parse(src2)
             ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ast.py", line 52, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<unknown>", line 2
    if 1:
         ^
IndentationError: expected an indented block after 'if' statement on line 2
```

If we mix tab and spaces, we get a slightly different error:
```
def main() -> void:
        print('hello')  # 8 spaces
	print('world')  # 1 tab
```

```
❯ spy -p /tmp/mix_tab_spaces.spy
[...]
  File "/usr/lib/python3.12/tokenize.py", line 576, in _generate_tokens_from_c_tokenizer
    for info in it:
  File "<string>", line 3
    	print('world')
    	             ^
TabError: inconsistent use of tabs and spaces in indentation
```

Most editors automatically convert tabs into spaces and so it's not easy to actually write `mix_tab_spaces.spy` to disk. It can be downloaded here:
https://antocuni.eu/files/misc/mix_tab_spaces.spy

I think that in both cases, the culprit is the `magic_py_parse.py` module, which does a dance of to tokenize/modify the token stream/untokenize/parse. We should be more robust and handle those cases better.

[...]

PR #181, opened 2025-06-10
Correction of tabs and indentation

PatrickRedStar:
Correction of tabs and indentation #180

### DIFF
--- a/spy/magic_py_parse.py
+++ b/spy/magic_py_parse.py
@@ -62,9 +62,9 @@ def magic_py_parse(src: str, filename: str = "<string>") -> py_ast.Module:
         if isinstance(node, py_ast.Name):
             assert node.end_lineno is not None
             assert node.end_col_offset is not None
-            loc = LocInfo(node.lineno, node.end_lineno,
+            loc_info = LocInfo(node.lineno, node.end_lineno,
                           node.col_offset, node.end_col_offset)
-            node.is_var = loc in var_locs # type: ignore
+            node.is_var = loc_info in var_locs # type: ignore
 
     return py_mod
 
@@ -113,13 +113,13 @@ def preprocess(src: str, filename: str = "<string>") -> tuple[str, set[LocInfo]]
                                tok0.start, tok1.end, tok1.line)
             newtokens.append(newtok)
             # compute the location info of the future ast.Name
-            loc = LocInfo(
+            loc_info = LocInfo(
                 lineno = var_l0,
                 end_lineno = var_l0,
                 col_offset = var_c0,
                 end_col_offset = var_c0 + len(tok1.string)
             )
-            var_locs.add(loc)
+            var_locs.add(loc_info)
             i += 1
         else:
             newtokens.append(tok0)
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -22,7 +22,7 @@ class AST:
     stubs/_ast.pyi
     """
 
-    _loc = None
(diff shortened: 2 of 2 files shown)


spy/magic_py_parse.py
```diff
@@ -42,13 +45,18 @@ class LocInfo:
...
     """
-    src2, var_locs = preprocess(src)
-    py_mod = py_ast.parse(src2)
+    src2, var_locs = preprocess(src, filename)
+    try:
+        py_mod = py_ast.parse(src2, filename=filename)
+    except SyntaxError as e:
+        lineno = e.lineno or 1
+        loc = Loc(filename, lineno, lineno, 0, -1)
+        raise SPyError.simple("W_ParseError", e.msg, "", loc)
```
antocuni, line 59: LGTM, but it would be nice to add a comment which explain _when_ we enter this error path (as opposed as the one in `preprocess`). E.g., "this is triggered by e.g. mixing tabs and spaces", or something like that.

spy/magic_py_parse.py
```diff
@@ -64,8 +72,17 @@ def get_tokens(src: str) -> list[TokenInfo]:
...
+    try:
+        tokens = get_tokens(src)
+    except (SyntaxError, TokenError) as e:
+        lineno = getattr(e, "lineno", None)
+        if lineno is None and isinstance(e, TokenError):
+            lineno = e.args[1][0] if e.args and isinstance(e.args[1], tuple) else 1
+        if lineno is None:
+            lineno = 1
+        loc = Loc(filename, lineno, lineno, 0, -1)
+        raise SPyError.simple("W_ParseError", str(e), "", loc)
```
antocuni, line 86: same as above, would be nice to explain _when_ we enter this error path.
Moreover, by running the two new tests I noticed that none of them actually raises `TokenError`.
Ideally, I would like to have a test which exercises this code path as well.

spy/tests/test_magic_py_parse.py
```diff
@@ -49,3 +54,27 @@ def test_magic_py_parse():
...
+    src = "def main() -> void:\n    if 1:\n"
+    f = tmpdir.join("test.spy")
+    f.write(src)
+    parser = Parser(src, str(f))
+    with expect_errors(
+        "expected an indented block after 'if' statement on line 2",
+        ("", "    if 1:")
+    ):
+        parser.parse()
+
```
antocuni, line 73: these are nitpicks, but I think it's a good idea to keep consistent style across tests:
1. could you use `textwrap.dedent()` and a multi-line string instead of embedding `\n` and spaces in a single line string?

2. All the other tests follow this pattern:
```
errors = expect_errors(...)
with errors:
    ...
```

spy/tests/test_magic_py_parse.py
```diff
@@ -49,3 +54,27 @@ def test_magic_py_parse():
...
+def test_magic_py_parse_tabs(tmpdir):
+    src = "def main() -> void:\n        print('hello')\n\tprint('world')\n"
+    f = tmpdir.join("test.spy")
+    f.write(src)
+    parser = Parser(src, str(f))
+    with expect_errors(
+        "inconsistent use of tabs and spaces in indentation (<string>, line 3)",
+        ("", "\tprint('world')")
+    ):
+        parser.parse()
```
antocuni, line 104: same as above

antocuni (approved):

PR #191, opened 2025-06-30
Replace usage of unbuffered with pexpect for colored ninja messages

seibert:
This will make it easier to package spy since it won't depend on an external tool (`expect`) that isn't available via pip or conda.

### DIFF
--- a/spy/util.py
+++ b/spy/util.py
@@ -133,10 +133,9 @@ def unbuffer_run(cmdline_s: Sequence[str]) -> subprocess.CompletedProcess:
         child.expect(pexpect.EOF)
         child.wait()  # avoid a race condition on child.exitstatus
 
-        # child.exitstatus should never be None if child.wait() finished, but
-        # mypy doesn't know that.  Will treat that situation as a command
-        # error.
-        returncode = -1 if child.exitstatus is None else child.exitstatus
+        # child.exitstatus is never None if child.wait() finished
+        assert child.exitstatus is not None
+        returncode = child.exitstatus
 
         return subprocess.CompletedProcess(
             args=cmdline_s,


spy/util.py
```diff
@@ -118,17 +119,56 @@ def shortrepr(s: str, n: int) -> str:
         s = s[:n-2] + '...'
     return repr(s)
 
+def unbuffer_run(cmdline_s: Sequence[str]) -> subprocess.CompletedProcess:
+    """
+    Simulate the behavior of the unbuffer command from the expect package.
+    
+    Like unbuffer, this assumes the command only outputs to stderr.
```
antocuni: This comment doesn't seem true. On my machine, `unbuffer` seems to be able to handle programs which outputs to both stdout and stderr, and writes everything to stdout.
I had to write two helper scripts to test it:
```
#!/usr/bin/env python3
# myprint.py
import sys

def main():
    if len(sys.argv) != 2:
        print("Usage: myprint.py [1|2|1+2]", file=sys.stderr)
        sys.exit(1)

    arg = sys.argv[1]

    if arg == "1":
        print("this is stdout")
    elif arg == "2":
        print("this is stderr", file=sys.stderr)
    elif arg == "1+2":
        print("this is stdout")
        print("this is stderr", file=sys.stderr)
    else:
        print(f"Unknown argument: {arg}", file=sys.stderr)
        sys.exit(1)

if __name__ == "__main__":
    main()
```

```
#!/usr/bin/env python3
# runproc.py

import sys
import subprocess

def run_subprocess(command):
    # Run the command and capture stdout and stderr
    result = subprocess.run(
        command,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        text=True  # ensures output is returned as string instead of bytes
    )

    print("STDOUT:")
    print(result.stdout)

    print("\nSTDERR:")
    print(result.stderr)

# Example usage:
if __name__ == "__main__":
    run_subprocess(sys.argv[1:])
```

Here are some examples:
```

❯ ./runproc.py unbuffer echo ciao
STDOUT:
ciao


STDERR:


/tmp 
❯ ./runproc.py unbuffer ls --error
STDOUT:
ls: unrecognized option '--error'
Try 'ls --help' for more information.


STDERR:


/tmp 
❯ ./runproc.py unbuffer ./myprint.py 1+2
STDOUT:
this is stdout
this is stderr


STDERR:
```

spy/util.py
```diff
@@ -118,17 +119,56 @@ def shortrepr(s: str, n: int) -> str:
...
+    try:
+        cmd = cmdline_s[0]
+        args = list(cmdline_s[1:])
+        child = pexpect.spawn(command=cmd, args=args)
+        child.expect(pexpect.EOF)
+        child.wait()  # this is needed to avoid a race condition on child.exitstatus
+
+        # child.exitstatus should never be None if child.wait() finished, but
+        # mypy doesn't know that.  Will treat that situation as a command error.
+        returncode = -1 if child.exitstatus is None else child.exitstatus
```
antocuni: if I understand correctly this line is just to make mypy happy?
Then I think I'd prefer something like `assert child.exitstatus is not None`

antocuni (commented):
LGTM, apart stdout vs stderr in `unbuffer_run`.

spy/util.py
antocuni: fixed by d79d8c57

Issue #173, opened 2025-06-06
Better error message if you try to import `foo.py`


antocuni:
If you do `import foo` and by mistake you have `foo.py` instead of `foo.spy`, you get an `ImportError`.

I think this mistake will be common enough to warrant a dedicated error message.

PR #197, opened 2025-07-19
Pretty print AST with color information 

### LINKED ISSUE #168: Pretty print AST with color information
It would be nice to have a way to:

1. redshift a module
2. pretty print the ORIGINAL ast in a way where each expr has the color computed by the redshifting

So e.g.:
```python
def foo(x: i32) -> i32:
    return x + 2 * 3
```

Currently, you get this:
```
$ spy -p x.spy
[...]
                    Return(
                        value=BinOp(
                            op='+',
                            left=Name(id='x'),
                            right=BinOp(
                                op='*',
                                left=Constant(value=2),
                                right=Constant(value=3),
                            ),
                        ),
                    ),
[...]
```

I would like something like thise:

```
$ spy --colorize x.spy
                    Return(                                    # red
                        value=BinOp(                           # red
                            op='+',
                            left=Name(id='x'),                 # red
                            right=BinOp(                       # blue
                                op='*',
                                left=Constant(value=2),        # blue
                                right=Constant(value=3),       # blue
                            ),
                        ),
                    ),
```

(instead of `#red` and `#blue` we should have ANSI colors)

aisipos:
Fixes #168 

I should say first, this is probably a larger PR than you were expecting. I'm totally fine if we abandon this approach and go in a different direction, but I wanted to see what directions I could come up with and if any are useful.

The tl;dr currently is, if you have `foo.spy`:
```
def main(x: i32) -> i32:
    return x + 2 * 3
```
Then run:
```
spy --redshift --colorize foo.spy
```
You will see:
image

This is probably not quite what we were looking for, for one I suspect we don't want the delemiters like () [] colored, or things like the top level Module colored, but the approach is close enough that I wanted to get some feedback first before proceeding in this direction.

I tried to make small atomic commits introducing single concepts at a time:
- In 8c5022b0929af6791bff67cdba394d0fda06bd11 I tried to follow your advice in saving the pre-redshift color in `DopplerFrame` instances. The problem I ran into here is that `DopplerFrame` instances are essentially ephemeral, they are not saved anywhere that is easily retrieved after redshifting. 
- So instead in cbd96b46b724d10a3359fb0eaedc9b2037cc4c51 I added a `color` attribute to `Expr` nodes, and save the original color during the redshifting process. These are the instances that are used in the AST itself that gets pretty printed, so it seemed like the natural place to add them for this use case. I realize this attribute is not used in the actual compilation pipeline, so if we feel this means the methodology here is no good I'm fine with figuring out an alternate method.
- The way that we want to do syntax coloring in this new "mode" is different than before, so to make this work I introduced a concept of `color_mode` in ddc30180766dad413f5c9d0630c1bd3314f7f4ae. This way we can use different coloring logic depending on the `color_mode`.
- To avoid conflation of names and concepts, I separated out the concept of `text_color` from `color` in 1151a7d9700e1f670bf2f430be620eea3cec8d19. When I think "color" I think "red or blue", but "text_color" can be anything, even including "green" or "turquoise".
- I have several commits that made several new ways of setting a new `text_color` that I ultimately all abandoned in favor of 0e9e4596acc6eb6d072a7eee24eee0cf5a9fc6b3, which introduces an `active_text_color` attribute to `TextBuilder`. This feature gets used by the logic now enabled in the new `redshift` `color_mode`, so that multiple nodes and children can automatically get the parent color, as implemented in 0e9e4596acc6eb6d072a7eee24eee0cf5a9fc6b3. However, as noted above, this method might be too broad, so we can work on this before we possibly merge.

This PR will need at the very least some cleanup, squashing, and tweaking before it is ready to merge, which I'm happy to do. Or we might decide the approach is not useful and do something else entirely. But since it does at least come close to the original intent of #168 I wanted to put it up to get some feedback first.

### DIFF
--- a/spy/ast_dump.py
+++ b/spy/ast_dump.py
@@ -61,7 +61,7 @@ class Dumper(TextBuilder):
         name = node.__class__.__name__
         fields = list(node.__class__.__dataclass_fields__)
         fields = [f for f in fields if f not in self.fields_to_ignore]
-        self._dump_node(node, name, fields, color='blue')
+        self._dump_node(node, name, fields, text_color='blue')
 
     def dump_py_node(self, node: py_ast.AST) -> None:
         name = 'py:' + node.__class__.__name__
@@ -69,12 +69,12 @@ class Dumper(TextBuilder):
         fields = [f for f in fields if f not in self.fields_to_ignore]
         if isinstance(node, py_ast.Name):
             fields.append('is_var')
-        self._dump_node(node, name, fields, color='turquoise')
+        self._dump_node(node, name, fields, text_color='turquoise')
 
     def dump_Symbol(self, sym: Symbol) -> None:
         self.write(f'Symbol({sym.name!r}, {sym.color!r}, {sym.varkind!r}, {sym.storage!r})')
 
-    def _dump_node(self, node: Any, name: str, fields: list[str], color: str) -> None:
+    def _dump_node(self, node: Any, name: str, fields: list[str], text_color: Optional[str]) -> None:
         def is_complex(obj: Any) -> bool:
             return (isinstance(obj, (spy.ast.Node, py_ast.AST, list, Symbol)) and
                     not isinstance(obj, py_ast.expr_context))
@@ -83,8 +83,8 @@ class Dumper(TextBuilder):
         multiline = any(is_complex_field)
         #
         if node is self.highlight:
-            color = 'red'
-        self.write(name, color=color)
+            text_color = 'red'
+        self.write(name, color=text_color)
         self.write('(')
         if multiline:
             self.writeline('')
--- a/spy/ast_dump.py
+++ b/spy/ast_dump.py
@@ -61,7 +61,8 @@ class Dumper(TextBuilder):
         name = node.__class__.__name__
         fields = list(node.__class__.__dataclass_fields__)
         fields = [f for f in fields if f not in self.fields_to_ignore]
-        self._dump_node(node, name, fields, text_color='blue')
+        # Use turquoise text_color to distinguish from blue in --colorize
+        self._dump_node(node, name, fields, text_color='turquoise')
 
     def dump_py_node(self, node: py_ast.AST) -> None:
         name = 'py:' + node.__class__.__name__
@@ -69,6 +70,7 @@ class Dumper(TextBuilder):
         fields = [f for f in fields if f not in self.fields_to_ignore]
         if isinstance(node, py_ast.Name):
             fields.append('is_var')
+        # Use turquoise text_color to distinguish from blue in --colorize
         self._dump_node(node, name, fields, text_color='turquoise')
 
     def dump_Symbol(self, sym: Symbol) -> None:
--- a/spy/ast_dump.py
+++ b/spy/ast_dump.py
@@ -1,23 +1,27 @@
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any, Optional
 import ast as py_ast
 import spy.ast
 from spy.analyze.symtable import Symbol
 from spy.textbuilder import TextBuilder
+if TYPE_CHECKING:
+    from spy.vm.vm import SPyVM
 
 def dump(node: Any,
          *,
          use_colors: bool = True,
          fields_to_ignore: Any = (),
          hl: Any = None,
+         vm: Optional['SPyVM'] = None,
          ) -> str:
-    dumper = Dumper(use_colors=use_colors, highlight=hl)
+    dumper = Dumper(use_colors=use_colors, highlight=hl, vm=vm)
     dumper.fields_to_ignore += fields_to_ignore
     dumper.dump_anything(node)
     return dumper.build()
 
 def pprint(node: Any, *, copy_to_clipboard: bool = False,
-           hl: Optional[spy.ast.Node]=None) -> None:
-    print(dump(node, hl=hl))
+           hl: Optional[spy.ast.Node]=None,
+           vm: Optional['SPyVM'] = None) -> None:
+    print(dump(node, hl=hl, vm=vm))
     if copy_to_clipboard:
         import pyperclip  # type: ignore
         out = dump(node, use_colors=False)
@@ -26,15 +30,18 @@ def pprint(node: Any, *, copy_to_clipboard: bool = False,
 
 class Dumper(TextBuilder):
     fields_to_ignore: tuple[str, ...]
+    vm: 'SPyVM | None'
 
     def __init__(self, *,
                  use_colors: bool,
                  highlight: Optional[spy.ast.Node] = None,
+                 vm: Optional['SPyVM'] = None
                  ) -> None:
         super().__init__(use_colors=use_colors)
         self.highlight = highlight
         self.fields_to_ignore = ('loc', 'target_loc', 'target_locs',
                                  'loc_asname')
+        self.vm = vm
 
     def dump_anything(self, obj: Any) -> None:
         if isinstance(obj, spy.ast.Node):
--- a/spy/ast_dump.py
+++ b/spy/ast_dump.py
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING, Any, Optional
 import ast as py_ast
+
+from spy.analyze.symtable import Color
 import spy.ast
 from spy.analyze.symtable import Symbol
 from spy.textbuilder import TextBuilder
@@ -86,7 +88,15 @@ class Dumper(TextBuilder):
         #
         if node is self.highlight:
             text_color = 'red'
-        self.write(name, color=text_color)
+        # If the vm contains an expr_color_map, use the expression's color
+        # as the text background color, and use the default text_color
+        bg_color = None
+        if self.vm and self.vm.expr_color_map:
+            color: Optional[Color] = self.vm.expr_color_map.get(node, None)
+            if color:
+                bg_color = color
+                text_color = None
+        self.write(name, color=text_color, bg=bg_color)
         self.write('(')
         if multiline:
             self.writeline('')
--- a/spy/cli.py
+++ b/spy/cli.py
@@ -55,6 +55,14 @@ class Arguments:
         )
     ] = False
 
+    colorize: Annotated[
+        bool,
+        Option(
+            "-C", "--colorize",
+            help="Output the pre-redshifted AST with blue / red text colors."
+        )
+    ] = False
+
     imports: Annotated[
         bool,
         Option(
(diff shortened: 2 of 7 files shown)


antocuni:
Thank you a lot for your work, I love it! I especially appreciate the thoughtfulness you put in explaining things commit-by-commit, it made it much easier to review the code.

Let's deal with things one by one.

---

### `ast.Expr.color`

I can see why it looks appealing to do that, but I'd highly prefer not to, for two reasons:
1. `--colorize` is something that people will do only rarely, and I don't want to pay the memory cost for an extra field which is completely unused 99% of the time

2. I'm trying to treat `ast.*` classes are immutable. This will make it easier to switch to a more compact representation later (where e.g. all the AST is binary-encoded in a cache-friendly contiguous buffer), which will be important for the performance of the interpreter

I see the problem with DopplerFrame being ephemeral. So, here is a new suggestion: we could introduce a new field `vm.expr_color_map`, which maps `ast.Expr` to their `Color`.
In normal mode, `vm.expr_color_map` is None, but when `--colorize` is passed, we set it to a dictionary, and then DopplerFrame knows when/where to store the info.

Finally, we will pass `expr_color_map` to `ast_dump.Dumper` so that it knows the color of the nodes.

---

### TextBuilder improvements

I think that your approach of using just one `active_text_color` doesn't work in case of nested `color_block()`. Moreover, I'm not sure I like the fact that just entering the color_block writes to the terminal, because e.g. it breaks if you then call `write(s, color='...')` (I think, I admit I didn't try).

I tried a slightly different approach in PR #200 which I think is more general:
  - you can have nested `color()` blocks
  - you now have support for bg colors
  - you can always temporarily override color and bg inside individual `write` calls, so that e.g. the following works:
https://github.com/spylang/spy/blob/9adbd94ca966320a1fa8dca54cc4138842903537/spy/textbuilder.py#L265-L275

---

### Appearance of the colored AST

When it comes to aesthetic things are very personal and opinionated of course, but I find the "all blue all red" screenshot a bit "too much" :).

Some random notes:
  - not all nodes are expressions, and thus not all should be colored
  - the current pretty-print logic prints node names in blue, but that's just an accident. It probably makes more sense to use a different color (such as turquoise, same as py_ast) just to avoid confusions

I randomly experimented with two "low-key" approaches. I pushed my code in the branch `tmp-anto-ast-color`, which is based on your commit 743a008d.

Consider this example:
```python
def fn(x: int, y: int) -> int:
    return x + y + 2*3
```

The first is to colorize *only the node name* in red/blue (and use turquoise for "default" color for non-expr nodes). You can find the code in commit c010af25):
image

The second is to keep all foreground colors normal, and use background color to distinguish red/blue. You can find the code in commit a07f5ad4:
image

I think I have a slight preference for the first, but I'm happy to hear opinions.

aisipos:
No worries, thanks for all the feedback! I review many PRs myself as well, so I also appreciate the value of making readable and reviewable PRs. Gotta write code to be readable! 🙂

Regarding the points you raised:
- OK, I'll abandon `ast.Expr.color` method as you noted. I was suspicious it would introduce some overhead, but since it did seem like the most direct route I thought I'd try it first to see how well it worked.  I'll instead try the `vm.expr_color_map` you suggested. 
- I saw that you merged a separate `TextBuilder` color PR. I'll rebase this PR on top of main and make use of the new features, but I might still try to preserve the naming of `text_color`. That way, if we grep in the codebase for `text_color` we'll see all the parts of the code that are trying to produce colored output. If we just use `color`, it becomes conflated with the code that detect the different node types.
- Regarding the proposed two styles of colorization you proposed, my initial reactions are
  - In favor of (2), 2 does draw the eye more than 1, in the additional use of background color as well as foreground color.
  - In favor of (1), 2 feels a bit "louder" than 1, in that more of the text is colored than 1.
- With that in mind, I might propose that we use a hybrid of the two, in that like (1) we only color the node's name, but we use the background color to highlight it like we do in (2). The nice part of this approach is that in this case foreground text_color corresponds to a node's "type", and the background text_color refers to a node's "color", so we are using a more consistent conceptual presentation. 
- We might not need the concept of `color_mode` any more, since if we have a `vm.expr_color_map` instance we can use it in the coloring, or if it is None we do nothing. The only real difference then is that the cli `--colorize` argument tells us to print the pre-redshifted AST, not the post redshift AST, as it currently does in this PR as of this writing.

Let me know what you think, I'll give this a try but probably will have a look at #173 first.

antocuni:
> I saw that you merged a separate TextBuilder color PR. I'll rebase this PR on top of main and make use of the new features, but I might still try to preserve the naming of text_color. That way, if we grep in the codebase for text_color we'll see all the parts of the code that are trying to produce colored output. If we just use color, it becomes conflated with the code that detects the different node types.

yes, that's a good idea. Other possible choices are `fgcolor` or `fg` (since we already have a `bg` attribute). But `text_color` is also good.

> With that in mind, I might propose that we use a hybrid of the two, in that like (1) we only color the node's name, but we use the background color to highlight it like we do in (2). The nice part of this approach is that in this case foreground text_color corresponds to a node's "type", and the background text_color refers to a node's "color", so we are using a more consistent conceptual presentation.

sounds good!
Also, eventually I'd like to be able to colorize also the source code (not only the AST), so this style plays well with that as well (e.g., we will be able to pretty-print the source code with red/blue background, and it will be consistent with the AST view).

> We might not need the concept of color_mode any more, since if we have a vm.expr_color_map instance we can use it in the coloring, or if it is None we do nothing. The only real difference then is that the cli --colorize argument tells us to print the pre-redshifted AST, not the post redshift AST, as it currently does in this PR as of this writing.

+1

aisipos:
OK, I essentially redid the PR with the new approach you suggested. I also made the cli work with just the `--colorize` option, there's no need now to run `--redshift --colorize`. I reverted all the earlier work in commit 879eb0926d9fd7d401b7a47678596adc429ddc7a, the commits after that add the new approach step by step. If we decide we like this approach, I'll delete all the old commits before we merge.

I added a new variable `expr_color_map` to `SPyVM`.  The main refactor was making this available to all the pretty print functions in `ast_dump.py`. I did this by adding an optional `vm` argument to `Dumper` and making it an instance attribute. I had to also add an optional `vm` argument to several functions that end up creating a Dumper instance, in particular `Node.pp`.

I tried my idea of using Blue / Red as the background color, and the current `text_color` concept as the foreground color, which partly looks like this:
image

I have slightly mixed feelings about it now that I can see it working in a real output, the contrast between the foreground color and background color clashes a bit and lowers the legibility somewhat. However, the red and blue colors do stand out more than if they were just foreground colors. Perhaps if we just ignored` text_color` when `bg_color` is specified it would look better, but then we do lose some information. 

Let me know what you think about the current implementation. Happy to make more changes if it's not a good fit.

spy/ast_dump.py
```diff
@@ -1,23 +1,29 @@
...
 from spy.textbuilder import TextBuilder
+if TYPE_CHECKING:
+    from spy.vm.vm import SPyVM
 
 def dump(node: Any,
          *,
          use_colors: bool = True,
          fields_to_ignore: Any = (),
          hl: Any = None,
+         vm: 'SPyVM | None' = None,
```
antocuni: nitpick: so far I always used `Optional[T]` instead of `T | None`.
It's a very minor thing but I think there is value in being consistent.
(I _might_ be convinced that it's better to use `| None` instead of `Optional`, but then it's a topic for a different PR).

spy/ast_dump.py
```diff
@@ -76,8 +87,15 @@ def is_complex(obj: Any) -> bool:
...
         #
         if node is self.highlight:
-            color = 'red'
-        self.write(name, color=color)
+            text_color = 'red'
+        # If the vm contains an expr_color_map, use the expression's color
+        # as the text background color
+        bg_color = None
+        if self.vm and self.vm.expr_color_map:
+            color: Color | None = self.vm.expr_color_map.get(node, None)
```
antocuni: +0.5 to pass around `vm` if it helps, but note that we could also pass just `expr_color_map`, and cut the circular dependency between `ast.py` and `vm.py.`
Anyway, not a big deal.

spy/cli.py
```diff
@@ -55,6 +55,14 @@ class Arguments:
         )
     ] = False
 
+    colorize: Annotated[
+        bool,
+        Option(
+            "-C", "--colorize",
+            help="Output the pre-redshifted AST with blue / red text colors."
+        )
+    ] = False
```
antocuni, line 64: `-C` is already used as an alias for `--cwrite`.
I think that `--colorize` is more important, so +1 to use `-C` here, but we should remove `-C` from the `cwrite` option.

spy/tests/test_doppler.py
```diff
@@ -294,3 +294,41 @@ def foo() -> None:
             raise TypeError('foo') # /.../test.spy:3
             raise ValueError # /.../test.spy:4
         """)
+
+    def test_expr_color_map_populated(self):
+        # Verify that when vm.expr_color_map is not None, redshift populates it
+        from unittest.mock import patch
```
antocuni: I'd use pytest's `monkeypatch` fixture if possible.

antocuni (approved):
I like it, thank you!
I have commented a couple of nitpicks, but nothing fundamental.
For the color yes, I agree that turquoise-on-blue and turquoise-on-red is not very visible, +1 to just use `default-on-bg` is bg is set.

spy/ast_dump.py
aisipos: I changed the PR to use `Optional[T]`. I agree that consistency is important. I don't have too strong an opinion on `Optional` vs `| None`, but I'll admit to a slight preference to `| None` in new projects. In theory it is "built in" to the language and doesn't require an import. But since Spy is currently using `Optional`, `Optional` it is.

spy/ast_dump.py
aisipos: Yeah, I decided to pass in `vm` since as more features get added, I didn't want to pass more and more arguments to the various pretty print functions. Then again, we probably don't want `vm` to become a "dumping ground" of global variables. But at least for now, passing in `vm` seemed like a good compromise.

spy/cli.py
aisipos: Done.

spy/tests/test_doppler.py
aisipos: Done.

aisipos:
@antocuni OK, I addressed all of your feedback, deleted all the old commits, and rebased on top of current `main`. The output now only sets the background color for red / blue:
image
Hopefully this fixes all our remaining issues.

antocuni:
@aisipos merged, thank you :pray:

PR #198, opened 2025-07-22
Give an explanatory error if user passes a .py file to cli.

aisipos:
Old habits die hard, more than once in development I've mistakenly passed a `.py` file to the cli, which will then spit out a long traceback. For the benefit of new users and old, write a helpful message to stderr and exit with an error code if the user asks to process a `.py` file.

Example:

```
$  spy --redshift foo.py 
Error: foo.py is a .py file, not a .spy file.
```

### DIFF
--- a/spy/cli.py
+++ b/spy/cli.py
@@ -307,6 +307,10 @@ async def inner_main(args: Arguments) -> None:
         do_pyparse(str(args.filename))
         return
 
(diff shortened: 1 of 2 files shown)


antocuni (approved):
That's a good idea, thank you!
Related issue: #173, which is a more general problem, but for now this is good enough.

PR #208, opened 2025-08-09
Better error messages when mistakenly importing .py files

### LINKED ISSUE #173: Better error message if you try to import `foo.py`

aisipos:
Fixes #173 

Introduces new ImportError messages if you mistakenly try to import a `.py` file. Example:
```
ImportError: cannot import `foo`
   --> /Users/anton/code/spy/examples/multifile/main.spy:3:1
  3 | import foo
    | |________| file `foo.py` exists, but py files cannot be imported

```

The two main bits of code needed to produce this error are in the method `get_filename`, currently a member of `ImportAnalizyer`, and `declare_Import` in `ScopeAnalyzer`. The main difficulty was to make the logic in `get_filename` available while running `declare_Import`. I implemented this by moving the `get_filename` method out of `ImportAnalizyer` and into `SPyVM`, and adding a `check_py` flag to it to detect matching `.py` files. This function already had a comment in it, suggesting an alternate implementation:
```
    # XXX maybe THIS is the right place where to raise SPyImportError?
```
I looked into this briefly, but decided it would be a bigger refactor to handle catching this error rather than moving `get_filename` to `SPyVM`, where we can use it in the `ScopeAnalyzer`. Hopefully this refactor is acceptable.


spy/vm/vm.py
```diff
@@ -106,6 +107,24 @@ def import_(self, modname: str) -> W_Module:
         w_mod = self.modules_w[modname]
         return w_mod
 
+    def get_filename(self, modname: str, check_py: bool = False) -> Optional[py.path.local]:
```
antocuni: A couple of naming suggestions:
  - I think that `get_filename` is a bit too generic for being a method in VM.
What about `find_files_to_import`?
  - I'm always wary of `check_*` boolean flags, because often it's not obvious what they do. What about very explicit names such as `allow_py_files` or maybe even `search_also_for_py_files`?

aisipos: I agree that `get_filename` is too general. I renamed it to `find_file_on_path`, which I thought was even more specific  than `find_files_to_import`, but I can use the other name if you prefer it. I took your suggestion for the flag argument though, and named it `allow_py_files`. I pushed a new commit in a69d32177b8252822ffec28804914c830a494205

antocuni: `find_file_on_path` is perfect! Merging :)

PR #235, opened 2025-10-01
Raises ZeroDivisionError Exception when appropriate

### LINKED ISSUE #183: Division and modulo should raise DivisionError when appropriate
Currently, if you try to divide (or modulo) by 0, you get a crash in the interpreter:
```
def main() -> void:
    print(2 / 0)
```

```
❯ spy /tmp/x.spy
[...]
  File "/home/antocuni/anaconda/spy/spy/vm/modules/operator/opimpl_int.py", line 47, in w_div
    return _binop(vm, w_a, w_b, lambda a, b: a / b)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/antocuni/anaconda/spy/spy/vm/modules/operator/opimpl_int.py", line 20, in _binop
    res = fn(a, b)
          ^^^^^^^^
  File "/home/antocuni/anaconda/spy/spy/vm/modules/operator/opimpl_int.py", line 47, in <lambda>
    return _binop(vm, w_a, w_b, lambda a, b: a / b)
                                             ~~^~~
ZeroDivisionError: division by zero
```

We should detect this case. Probably the right fix is something along these lines:
```
# in opimpl_int.py

    @OP.builtin_func(f'{T}_div')
    def w_div(vm: 'SPyVM', w_a: WT, w_b: WT) -> W_F64:
        if w_b.value == 0:
            raise ZeroDivisionError
        return _binop(vm, w_a, w_b, lambda a, b: a / b)
```

But it should be done with the proper tests, and for all division/modulo operations (including e.g. floordiv, float division, etc.)

Viriathus1:
This PR fixes the behavior of integer and float division, floor division and modulo operators so they correctly raise ZeroDivisionError when the divisor is zero, across all compiler backends (interp, doppler, wasm).

Changes include:
- Added explicit zero checks to opimpl_int.py and opimpl_float.py for operators /, // and %.
- Updated spy/libspy/include/spy/operator.h to include guards for division, floor division and modulo operations so that the c/wasm backend no longer traps on divide-by-zero.
- Added new test cases in test_int.py and test_float.py to confirm correct behaviour.

This closes #183.


spy/libspy/include/spy/operator.h
```diff
@@ -53,18 +53,74 @@ static inline void spy_operator$raise(spy_Str *etype,
 }
 
 static inline double spy_operator$i8_div(int8_t x, int8_t y) {
+    if (y == 0) {
+        spy_panic("ZeroDivisionError", "Raised when the second argument of a division or modulo operation is zero",
+                  __FILE__, __LINE__);
+        return nan("");
```
antocuni: I think it's better to use the same error messages as on CPython:
```
>>> 1/0
[...]
ZeroDivisionError: division by zero
>>> 1.0/0.0
[...]
ZeroDivisionError: float division by zero
```

antocuni: also, I suppose you had to insert the `return nan("")` to silence a warning from the compiler, but it looks like the "wrong" fix.
The point is that `spy_panic` is never supposed to return, so the compiler should be able to understand it and don't complain.

I *think* that the correct fix is to mark `spy_panic` as `noreturn`.
ChatGPT suggests to use the following pattern:
```
#if defined(__GNUC__) || defined(__clang__)
#define NORETURN __attribute__((noreturn))
#elif defined(_MSC_VER)
#define NORETURN __declspec(noreturn)
#else
#define NORETURN _Noreturn  // C11
#endif

NORETURN
void spy_panic(const char *err, const char *msg,
               const char *file, int line);
```

Make sure to put the `#define NORETURN` in `spy.h`, where we already define `WASM_IMPORT`.

spy/vm/modules/operator/opimpl_f64.py
```diff
@@ -35,10 +36,14 @@ def w_f64_mul(vm: 'SPyVM', w_a: W_F64, w_b: W_F64) -> W_F64:
 
 @OP.builtin_func
 def w_f64_div(vm: 'SPyVM', w_a: W_F64, w_b: W_F64) -> W_F64:
+    if w_b.value == 0:
+        raise SPyError("W_ZeroDivisionError", "Raised when the second argument of a division or modulo operation is zero")
```
antocuni: same as above, better to use the same error message as CPython

antocuni (changes requested):
Very good job :muscle: 
The PR looks very good, apart from the minor tweaks about error messages and `NORETURN`.

spy/tests/compiler/test_int.py
```diff
@@ -64,11 +64,12 @@ def floordiv(x: T, y: T) -> T: return x // y
...
         assert mod.div(11, 2) == 5.5
         assert mod.floordiv(11, 2) == 5
-        with SPyError.raises("W_ZeroDivisionError"):
+        with SPyError.raises("W_ZeroDivisionError", match="integer modulo by zero"):
             mod.mod(10, 0)
-        with SPyError.raises("W_ZeroDivisionError"):
+        with SPyError.raises("W_ZeroDivisionError", match="division by zero"):
             mod.div(11, 0)
         with SPyError.raises("W_ZeroDivisionError"):
+            # TODO: add appropriate floor div handling for c backend
```
antocuni, line 72: is this TODO still relevant or we should remove the comment?

Viriathus1: yes it's just to remind me to match the error message in that test case as well once I complete this issue #238

PR #240, opened 2025-10-02
Add assert implementation

sbundaz:
Relates to #179 

Assert statement implemented

### DIFF
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -484,7 +484,7 @@ class Raise(Stmt):
 @dataclass(eq=False)
 class Assert(Stmt):
     test: Expr
-    msg: Expr | None
+    msg: Optional[Expr]
 
 
 # ====== IR-specific nodes ======
--- a/spy/tests/compiler/test_assert.py
+++ b/spy/tests/compiler/test_assert.py
@@ -1,11 +1,9 @@
-import pytest
 from spy.errors import SPyError
-from spy.tests.support import CompilerTest
+from spy.tests.support import CompilerTest, expect_errors
 
 
 class TestAssert(CompilerTest):
     def test_assert_true(self):
-        """Test that assert True passes without error"""
         mod = self.compile(
             """
             def test() -> None:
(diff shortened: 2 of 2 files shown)


spy/ast.py
```diff
@@ -481,6 +481,11 @@ class While(Stmt):
 class Raise(Stmt):
     exc: Expr
 
+@dataclass(eq=False)
+class Assert(Stmt):
+    test: Expr
+    msg: Expr | None
```
antocuni:
```suggestion
    msg: Optional[Expr]
```

just for consistency with all the rest of the code

spy/tests/compiler/test_assert.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_assert_false(self):
+        """Test that assert False raises AssertionError"""
+        mod = self.compile(
+            """
+            def test() -> None:
+                assert False
+            """
+        )
+
+        with pytest.raises(SPyError) as exc_info:
```
antocuni: you can use `with SPyError.raises('W_AsserionError')` here

spy/tests/compiler/test_assert.py
```diff
@@ -0,0 +1,92 @@
...
+            self.compile(
+                """
+                def test() -> None:
+                    assert False, 42
+                """
+            ).test()
+
+        assert exc_info.value.etype == "W_TypeError"
+        assert "expected `str`, got `i32`" in str(exc_info.value)
```
antocuni: here you should use the `expect_errors` pattern. See this for example:
https://github.com/spylang/spy/blob/aaca3644490068ca2a0393edd9e50cfb5c41ea3e/spy/tests/compiler/test_basic.py#L76-L86

antocuni (changes requested):
Thank you @AndreBonda !
Very solid PR :muscle: 
I requested just a couple of minor changes, mostly around how to test errors, for consistency with the rest of the testsuite.

PR #251, opened 2025-10-06
Conform floor division and modulo to Python semantics

### LINKED ISSUE #238: Improve division tests and add float modulo
Python and C has slightly different rules for rounding of int division.
We should:
  - write tests to check that spy behaves like python
  - possibly fix the C backend if needed

Moreover, we should add `float % float`, as I think that it's currently not implemented in SPy.

---

The following table has been generated by ChatGPT:

| Expression | Python 3 (`//` = floor division) | C (C99+, `/` with ints = trunc toward zero) |
|------------|----------------------------------|---------------------------------------------|
| `7 / 3`    | `2`                              | `2`                                         |
| `-7 / 3`   | **`-3`**                         | **`-2`**                                    |
| `7 / -3`   | **`-3`**                         | **`-2`**                                    |
| `-7 / -3`  | `2`                              | `2`                                         |

**Notes:**
- With same-sign operands, results match.  
- With mixed signs, Python floors (toward −∞), while C truncates (toward 0).

Viriathus1:
This PR correctly implements floor division (//) and modulo (%) operators, fully matching Python’s semantics for both integers and floats.

Changes include:
- Signed Integer floor, eg:
```c
int32_t q = x / y;
int32_t r = x % y;

if ((r != 0) && ((x ^ y) < 0)) {
    q -= 1;
}

return q;
```
Now it uses pure integer arithmetic which matches CPython's internal implementation and ensures negative values round correctly toward negative infinity.
- Signed Integer modulo, eg:
```c
int32_t r = x % y;

if ((r != 0) && ((x ^ y) < 0)) {
    r += y;
}

return r;
```
Guarantees that the result of `x % y` always has the same sign as `y` consistent with Python.
- Added float modulo operator.
- Added new unit tests to cover all combinations of positive and negative dividends and divisors.

This closes #238

### DIFF
--- a/spy/tests/compiler/test_float.py
+++ b/spy/tests/compiler/test_float.py
@@ -60,10 +60,10 @@ class TestFloat(CompilerTest):
         assert mod.mod(3.5, -1.5) == -1.0
         assert mod.mod(-3.5, 1.5) == 1.0
         assert mod.mod(-3.5, -1.5) == -0.5
-        assert 5.0 % float('inf') == 5.0
-        assert -5.0 % float('inf') == float('inf')
-        assert 5.0 % float('-inf') == float('-inf')
-        assert -5.0 % float('-inf') == -5.0
+        assert mod.mod(5.0, float('inf')) == 5.0
+        assert mod.mod(-5.0, float('inf')) == float('inf')
+        assert mod.mod(5.0, float('-inf')) == float('-inf')
+        assert mod.mod(-5.0, float('-inf')) == -5.0
 
 
     def test_CompareOp(self):


spy/vm/modules/operator/opimpl_int.py
```diff
@@ -13,15 +13,14 @@ class W_IntLike(Protocol):
 
 def make_ops(T: str, pyclass: type[W_Object]) -> None:
     w_T = pyclass._w  # e.g. B.w_i32
-    WT = Annotated[W_IntLike, w_T]
```
antocuni, line 16: do you have any specific reason for killing this?
I personally find having `Annotated[W_IntLike, w_T]` everywhere hard to read, and this patterns is used a bit everywhere, e.g. here:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/spy/vm/member.py#L61-L78

Viriathus1: @antocuni I was running up against Pylance Type error `Variable not allowed in type expression` which made it hard to work with. I've just changed the language server to Jedi and it works ok with that pattern so I'm happy to revert it.  

It could be good to have the recommended static type checker (mypy) and language server on the readme for local development because they're all slightly different and the default language server on vscode is Pylance.

antocuni: I recommend mypy: this is what the CI tests and it's the only one "officially" supported.
I'm not opposed to switch to another type checker, but to do so we would need to understand pros&cons and to test it in CI.

Ironically, the fact that in Python "type checking" is not a well defined concept and that every typechecker implements slightly different rules is exactly one of the things that SPy aims to solve :).
But for that, we need to wait until SPy is powerful enough to compile itself :)

spy/tests/compiler/test_float.py
```diff
@@ -21,18 +21,50 @@ def sub(x: f64, y: f64) -> f64:      return x - y
...
+        assert mod.floordiv(-3.5, 1.5) == -3.0
+        assert mod.floordiv(-3.5, -1.5) == 2.0
+        assert mod.mod(3.5, 1.5) == 0.5
+        assert mod.mod(3.5, -1.5) == -1.0
+        assert mod.mod(-3.5, 1.5) == 1.0
+        assert mod.mod(-3.5, -1.5) == -0.5
+        assert 5.0 % float('inf') == 5.0
+        assert -5.0 % float('inf') == float('inf')
+        assert 5.0 % float('-inf') == float('-inf')
+        assert -5.0 % float('-inf') == -5.0
```
antocuni: I don't understand what is the purpose of the last 4 asserts: they are testing *CPython* behavior, not spy. You probably wanted to write `assert mod.mod(5.0, float('inf')) == 5.0` ?

Viriathus1: @antocuni I don't know how I missed this, fixed now.

PR #254, opened 2025-10-07
allow raise Exception()

### LINKED ISSUE #250: Allow "raise Exception()"
This works:
```python
raise Exception
raise Exception("hello")
```

This doesn't:
```python
raise Exception()
```

We should probably allow it and make it equivalent to `raise Exception("")`.
We can do it by checking number of arguments in `W_Exception.w_NEW`:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/spy/vm/exc.py#L54-L72

_Originally posted by @dpdani in https://github.com/spylang/spy/pull/172#discussion_r2404408596_

ernestoarbitrio:
fixes #250

### DIFF
--- a/spy/tests/compiler/test_exception.py
+++ b/spy/tests/compiler/test_exception.py
@@ -15,6 +15,8 @@ class TestException(CompilerTest):
                 raise Exception("hello")   # <-- line 6
             elif x == 2:
                 raise ValueError("world")
+            elif x == 10:
+                raise ValueError()
             else:
                 raise IndexError
         """)
@@ -29,6 +31,8 @@ class TestException(CompilerTest):
             mod.foo(2)
         with SPyError.raises('W_IndexError'):
             mod.foo(3)
+        with SPyError.raises('W_ValueError', match="") as excinfo:
+            mod.foo(10)
 
     def test_cannot_raise_red(self):
         src = """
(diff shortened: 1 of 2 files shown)


spy/tests/compiler/test_exception.py
```diff
@@ -15,6 +15,8 @@ def foo(x: i32) -> i32:
                 raise Exception("hello")   # <-- line 6
             elif x == 2:
                 raise ValueError("world")
+            elif x == 10:
+                raise ValueError()
```
antocuni, line 19: `mod.foo(3)` vs `mod.foo(10)` now is a bit confusing, it took a while to me to understand what was going on.
I suggest to use `x == 3` for `ValueError()`, and later call `mod.foo(4)` for the `IndexError` case

spy/vm/exc.py
```diff
@@ -55,6 +55,15 @@ def __repr__(self) -> str:
...
         # we cannot use the default __new__ because we want to pass w_cls
+        _args_wam: list[W_MetaArg] = list(args_wam) or [
+            W_MetaArg(
+                loc=Loc.here(),
+                w_static_T=BUILTINS.w_str,
+                vm=vm,
+                color="blue",
+                w_val=W_Str(s="", vm=vm),
+            )
+        ]
```
antocuni, line 60: You can use `W_MetaArg.from_w_obj` instead:
```
W_MetaArg.from_w_obj(vm, vm.wrap(""))
```

antocuni (changes requested):
Good job 💪 
The two requested changes are somewhat nitpicking just to improve a bit the readability, but overall the PR looks very good 👍

antocuni (approved):
LGTM 👍

Issue #244, opened 2025-10-03
Implement the `not` unary operator


antocuni:
This probably can be done by adding a new `w_NOT` operator to unaryop.py:
https://github.com/spylang/spy/blob/94a3999ccbd2f33e09a075e1c02a157f07fd5e18/spy/vm/modules/operator/unaryop.py#L1-L27

The trick is that it works differently than other operators. I don't think we need a multimethod for NOT, because the impl it's always "do bool(arg), return the opposite".

So, I THINK that this implementation should be fine (pseudocode):
```
def w_NOT(vm, wam_arg):
    def impl(vm, w_arg: W_Bool) -> W_Bool:
        return vm.wrap(not vm.unwrap_bool(w_arg))
    return W_OpSpec(impl)
```

the trick is that `impl` takes a bool, so the typechecking machinery should take care of inserting the appropriate conversion, or raise the appropriate TypeError error if not possible.
There should also be a test for the TypeError case.

PR #258, opened 2025-10-09
Add not unary operator

sbundaz:
Relates to #244

### DIFF
--- a/spy/vm/modules/operator/__init__.py
+++ b/spy/vm/modules/operator/__init__.py
@@ -91,10 +91,7 @@ _from_token: dict[str, W_Func] = {
     "<universal_ne>": OP.w_UNIVERSAL_NE,
 }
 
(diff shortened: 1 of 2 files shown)


spy/tests/compiler/operator/test_unaryop.py
```diff
@@ -47,3 +47,12 @@ def foo(x: i32) -> i32:
         assert mod.foo(5) == -5
         assert mod.foo(-10) == 10
         assert mod.foo(0) == 0
+
+    def test_not(self):
+        src = """
+        def foo(x: bool) -> bool:
+            return not x
+        """
```
antocuni, line 58:
```suggestion
        src = """
        def foo(x: bool) -> bool:
            return not x
            
        def bar(x: i32) -> bool:
            return not x
        """
```

can you also check this function? This way we ensure that it works also with implicit conversions

PR #263, opened 2025-10-11
Fix \x C literal generation for clang

### LINKED ISSUE #255: Wrong string literals emitted by the C backend
```python
def main() -> None:
    print('\nBall')
```

`spy --compile` fails when we use clang (e.g. on macosx). This happens because the C backend generates this code:

```c
static spy_Str SPY_g_str0 = {5, "\x0aBall"};
```

GCC gives this warning (which is not shown by default, you need to manually invoke `ninja` to see it:
```
[1/1] CC x
src/x.c:9:43: warning: hex escape sequence out of range
    9 | static spy_Str SPY_g_str0 = {5, "\x0aBall"};
      |                                           ^
```

clang errors out:
```
clang-20 src/x.c -o x --std=c99 -Werror=implicit-function-declaration -Wfatal-errors -fdiagnostics-color=always -I /home/antocuni/anaconda/spy/spy/libspy/include -DSPY_TARGET_NATIVE -DSPY_DEBUG -O0 -g -L /home/antocuni/anaconda/spy/spy/libspy/build/native/debug -lspy -DSPY_DEBUG -O0 -g
src/x.c:9:34: fatal error: hex escape sequence out of range
    9 | static spy_Str SPY_g_str0 = {5, "\x0aBall"};
      |                                  ^~~~~~
```

Apparently, the C standard mandates that `"\x"` consumes "as many hex digits as possible" :face_with_rolling_eyes: .
So, the correct semantics is to consume `\0x0aBa` and then complain it's too large.
According to chatgpt the "fix" is to split the string literal into two parts:
"\0x0a" "Ball..."

The code which emits C literals is here:
https://github.com/spylang/spy/blob/c40ae1471b0dc471d927939049ff664a21a87dfd/spy/backend/c/c_ast.py#L86-L101

we should adapt it to handle this corner case. Ideally, it should split the string ONLY if it's strictly necessary, to make sure that the "default" case is more readable.

aisipos:
Fixes #255 

The C backend will turn the string `'\nball'` into the C literal string `"\x0aball"`. clang will parse the beginning escape into a two byte hex string followed by `'ll'`, which is not what we want, we want a single byte escape sequence to encode the newline.

The solution implemented here is to use C string concatenation to ensure that a `\x` only escapes one hex byte. The algorithm used here is to add `""` to the beginning and end of `\x` escapes. The corner cases though of `\x` at the start or end of strings is then "fixed" by turning `""""`  into `"`. This is a quick fix but a little clunky, I can use a little smarter algorithm if we prefer.

### DIFF
--- a/spy/backend/c/c_ast.py
+++ b/spy/backend/c/c_ast.py
@@ -34,6 +34,7 @@ PREC  CATEGORY         OPERATOR            ASSOCIATIVITY
 
 import re
 from dataclasses import dataclass
+from enum import Enum
 from typing import ClassVar
 
 from .context import C_Type
@@ -99,17 +100,17 @@ class Literal(Expr):
                 return rf"\{ch}"
             elif 32 <= val < 127:
                 return ch
-            # \x in C will consume as many legal hex digits as possible.
-            # To prevent this from consuming following characters, put it
-            # in its own quotes and rely on string concatenation.
-            return rf'""\x{val:02x}""'  # :x is "hex format"
+            return rf"\x{val:02x}"  # :x is "hex format"
 
         lit = "".join([char_repr(val) for val in b])
-        # Remove any empty string concatenations at the front or back,
-        # then surround with quotes
-        lit = f'"{lit}"'
-        lit = lit.replace('"""', '"')
-        return Literal(lit)
+
+        # The C standard mandates that `\x` consumes as many hex digits as possible, but
+        # what we want is that each of them is followed by exactly TWO hex digits. The
+        # following regex finds occurences of `\x` followed by THREE hex digits, and
+        # inserts a literal "" after the 2nd. E.g. "\x0aBall" -> "\x0a""Ball"
+        lit = re.sub(r"(\\x[0-9A-Fa-f]{2})(?=[0-9A-Fa-f])", r'\1""', lit)
+
+        return Literal(f'"{lit}"')
 
 
 @dataclass
--- a/spy/tests/test_backend_c.py
+++ b/spy/tests/test_backend_c.py
@@ -72,4 +72,5 @@ class TestExpr:
         assert cstr(b"--\x00--\n--\xff--") == r'"--\x00--\x0a--\xff--"'
         assert cstr(b"\nball") == r'"\x0a""ball"'
         assert cstr(b"ball\n") == r'"ball\x0a"'
-        assert cstr(b"ball\n\nball") == r'"ball""\x0a\x0a""ball"'
+        assert cstr(b"ball\n\nball") == r'"ball\x0a\x0a""ball"'
+        assert cstr(b"\x00\x01\x02") == r'"\x00\x01\x02"'


spy/tests/test_cli.py
```diff
@@ -174,3 +186,27 @@ def test_execute_pyodide(self):
...
+        """
+        self.main_spy.write(textwrap.dedent(src))
+        res, stdout = self._run_compilation(target)
+        cmd = self._check_build_outputs(target)
+
+        status, out = getstatusoutput(cmd)
+        if status != 0:
+            print(out)
+            assert False, f"command failed: {cmd}"
+        assert out == "\nBall"
```
antocuni: Is there any specific reason why you put this test in `test_cli`? I naively expected to find it in `tests/compiler/test_str.py`.

aisipos: I agree that `test_cli` seems like an odd place for this test, but this was the only existing place that I knew for sure called the native C backend. I wasn't sure if the WASI or emscripten backends exhibited the bug. I can move the test into `test_str.py` if it makes sense.

spy/backend/c/c_ast.py
```diff
@@ -92,17 +93,41 @@ def from_bytes(cls, b: bytes) -> "Literal":
...
-            return rf"\x{val:02x}"  # :x is "hex format"
-
-        lit = "".join([char_repr(val) for val in b])
-        return Literal(f'"{lit}"')
+        # First tokenize into lists of tokens
+        pos = 0
+        tokens = []
+        while pos < len(b):
+            # Escape backslash and double quote
+            if match := re.match(rb"[\\\"]", b[pos:]):
```
antocuni: I think that `b[pos:]` makes the algorithm quadratic, and thus potentially very slow for large strings:
- the `while pos < len(b)` is O(N) on the length of b
- `b[pos:]` is also O(N), so it becomes O(N**2)

spy/tests/test_backend_c.py
```diff
@@ -70,3 +70,7 @@ def cstr(b: bytes) -> str:
         assert cstr(b'--"hello"--') == r'"--\"hello\"--"'
         assert cstr(rb"--aa\bb--") == r'"--aa\\bb--"'
         assert cstr(b"--\x00--\n--\xff--") == r'"--\x00--\x0a--\xff--"'
+        assert cstr(b"\nball") == r'"\x0a""ball"'
+        assert cstr(b"ball\n") == r'"ball\x0a"'
+        assert cstr(b"ball\n\nball") == r'"ball\x0a\x0a""ball"'
+        assert cstr(b"\x00\x01\x02") == r'"\x00\x01\x02"'
```
antocuni, line 76: love the new tests :)

spy/tests/test_cli.py
antocuni: > but this was the only existing place that I knew for sure called the native C backend. 

all the tests in `tests/compiler` are executed 3 times: in `[interp]`, `[doppler]` and `[C]` mode (you can see it with `pytest -v`). The `[C]` tests use the C backend and then use `zig cc` to compile the generated C code to WASM. And `zig cc` is just a wrapper around clang, so I expect the bug to be totally visible there.


> I move the test into `test_str.py` if it makes sense.

yes, I think it's better to move them there, if you don't mind.

aisipos: Ok, I force pushed a fix in d31af602e0a480ec54d4746544b35caa410a51bf, and verified that the test fails without the fix on clang (my primary dev environment is MacOS at the moment), and works with the fix.

```
E           python -m ziglang cc src/test.c -o test.wasm --std=c99 -Werror=implicit-function-declaration -Wfatal-errors -fdiagnostics-color=always -I /Users/anton/code/spy/spy/libspy/include -DSPY_TARGET_WASI -DSPY_DEBUG -O0 -g -mmultivalue -Xclang -target-abi -Xclang experimental-mv --target=wasm32-wasi-musl -lm -L /Users/anton/code/spy/spy/libspy/build/wasi/debug -lspy -DSPY_DEBUG -O0 -g --target=wasm32-wasi-musl -mexec-model=reactor '-Wl,--export=spy_test$foo'
E           src/test.c:9:34: fatal error: hex escape sequence out of range
E               9 | static spy_Str SPY_g_str0 = {5, "\x0aBall"};
E                 |                                  ^~~~~~
E           1 error generated.
E           ninja: build stopped: subcommand failed.
```

spy/backend/c/c_ast.py
aisipos: Ah yes, silly me, I'd forgotten that Python string slices are copies and not "views" (a bad thing to forget), so yes indeed that algorithm would be quadratic. I like your regex solution, it's not that complicated a solution (only one lookahead assertion), and only a minor change to the last version, so let's go with it. I force pushed a fix in a21fbcdeebb5c2724602a84e16a8c074e76ec4db

antocuni (approved):
LGTM, thank you!

Issue #248, opened 2025-10-04
Implement `and` and `or`


antocuni:
We should support `expr and expr` and `expr or expr`.

This involves:
  - adding `And` and `Or` nodes to `ast.py`
  - adding support for `and` and `or` in the parser (make sure to add tests!)
  - write tests in `test_basic.py`: the tests should make sure to exercise the "short-circuit" behavior of the two operators. The easiest way to test the presence or absence of a side effect is to use a function call which mutates a `var`, see e.g. `test_void_return:
https://github.com/spylang/spy/blob/ead6d958393b66e69582606021ec24b32b4b2882/spy/tests/compiler/test_basic.py#L229-L245
  - add the relevant methods to ASTFrame and DopplerFrame
  - add the relevant tests and code to the SPy backend (`spy/backend/spy.py` and `tests/test_backend_spy.py`)
  - fix the C backend

sbundaz:
Hi @Jemeljanov,
yeah I've been busy lately. Feel free to pick it up if you want :)

PR #264, opened 2025-10-15
Add `or` and `and` expressions

sbundaz:
Relates to #248

### DIFF
--- a/spy/vm/astframe.py
+++ b/spy/vm/astframe.py
@@ -145,7 +145,7 @@ class AbstractFrame:
             exp_loc = self.symtable.lookup(varname).type_loc
             if varname == "@return":
                 because = " because of return type"
-            elif varname in ("@if", "@while", "@assert"):
+            elif varname in ("@if", "@while", "@assert", "@and", "@or"):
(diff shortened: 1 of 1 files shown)


spy/tests/test_parser.py
```diff
@@ -39,10 +39,12 @@ def assert_dump(self, node: ast.Node, expected: str):
             pytest.fail("assert_dump failed")
 
     def test_Module(self):
-        mod = self.parse("""
+        mod = self.parse(
+            """
```
antocuni, line 43: something wrong happened here, it seems that the automatic formatter changed all these invocations?
With ruff 0.14.0 it doesn't happen:
```
spybranch on  main [$] 🐍(spy) ⚠️env
❯ ruff version
ruff 0.14.0

spybranch on  main [$] 🐍(spy) ⚠️env
❯ ruff format spy/tests/test_parser.py 
1 file left unchanged
```

maybe you are using a different ruff version?

sbundaz: mmm IDK, I'll have a look

PR #266, opened 2025-10-16
Colorize source code

### LINKED ISSUE #265: Negative consts have the wrong `loc`
Consider this test:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/tests/test_parser.py#L715-L732

If we add the following extra check, it fails:
```
        self.assert_dump(stmt, expected)
        #
        const = stmt.value.left  # this is the node for "-123"
        assert const.value == -123
        assert const.loc.get_src() == "-123"
```

Image

This is the code which treats the case `-NUM` as a special case. Probably it needs to be fixed with the right `.loc`:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/parser.py#L690-L701

ernestoarbitrio:
Also fixes #265

### DIFF
--- a/spy/parser.py
+++ b/spy/parser.py
@@ -1,6 +1,4 @@
 import ast as py_ast
-import linecache
-import re
 from types import NoneType
 from typing import NoReturn, Optional
 
@@ -699,23 +697,9 @@ class Parser:
             and isinstance(value, spy.ast.Constant)
             and isinstance(value.value, (int, float))
         ):
-
-            def unary_offset(line: str, value: str) -> int:
-                # Match a unary operator (+, -, ~) followed by spaces and the exact value
-                pattern = rf"([+\-~])(\s+){re.escape(value)}(?!\w)"
-                m = re.search(pattern, line)
-                return len(m.group(2)) + 1 if m else 1
-
             c_loc = value.loc
-            srcline = linecache.getline(c_loc.filename, c_loc.line_start)
-            offset = unary_offset(srcline, str(value.value))
-            new_loc = Loc(
-                c_loc.filename,
-                c_loc.line_start,
-                c_loc.line_end,
-                c_loc.col_start - offset,  # handle the negative sign
-                c_loc.col_end,
-            )
+            op_loc = py_node.loc
+            new_loc = Loc.combine(op_loc, c_loc)
             return spy.ast.Constant(new_loc, -value.value)
         return spy.ast.UnaryOp(py_node.loc, op, value)
 
--- a/spy/tests/test_parser.py
+++ b/spy/tests/test_parser.py
@@ -712,36 +712,43 @@ class TestParser:
         """
         self.assert_dump(stmt, expected)
 
-    @pytest.mark.parametrize(
-        "srcline, expected_src, expected_value",
-        (
-            ("return -123 * -1.0", "-123", -123),
-            ("return -     123 * -1.0", "-     123", -123),
-            ("return (-         123) * -1.0", "-         123", -123),
-            ("return ((-   123) * -1.0)", "-   123", -123),
-        ),
-    )
-    def test_negative_const(self, srcline, expected_src, expected_value):
+    def test_negative_const(self):
         # special case -NUM, so that it's seen as a constant by the rest of
         # the code
         mod = self.parse(f"""
-        def foo() -> f64:
-            {srcline}
+        def foo() -> None:
+            -100
+            -  101
+            (-    102)
         """)
-        stmt = mod.get_funcdef("foo").body[0]
+        funcdef = mod.get_funcdef("foo")
         expected = """
-        Return(
-            value=BinOp(
-                op='*',
-                left=Constant(value=-123),
-                right=Constant(value=-1.0),
-            ),
+        FuncDef(
+            color='red',
+            kind='plain',
+            name='foo',
+            args=[],
+            vararg=None,
+            return_type=Constant(value=None),
+            docstring=None,
+            body=[
+                StmtExpr(
+                    value=Constant(value=-100),
+                ),
+                StmtExpr(
+                    value=Constant(value=-101),
+                ),
+                StmtExpr(
+                    value=Constant(value=-102),
+                ),
+            ],
+            decorators=[],
         )
         """
-        const = stmt.value.left  # type: ignore[attr-defined]
-        self.assert_dump(stmt, expected)
-        assert const.value == expected_value
-        assert const.loc.get_src() == expected_src
+        self.assert_dump(funcdef, expected)
+        assert funcdef.body[0].value.loc.get_src() == "-100"  # type: ignore
+        assert funcdef.body[1].value.loc.get_src() == "-  101"  # type: ignore
+        assert funcdef.body[2].value.loc.get_src() == "-    102"  # type: ignore
 
     @pytest.mark.parametrize("op", "== != < <= > >= is is_not in not_in".split())
     def test_CompareOp(self, op):


spy/util.py
```diff
@@ -231,6 +255,39 @@ def func_equals(f: Callable, g: Callable) -> bool:
     return True
 
 
+@typing.no_type_check
+def colors_coordinates(ast_module, expr_color_map) -> dict[int, list[tuple[str, str]]]:
```
ernestoarbitrio, line 248: This function is in `util.py` because it can be used by some external ide/plugins to get the color map and use that to colorize the code in the IDE itself

spy/tests/test_cli.py
```diff
@@ -116,10 +145,27 @@ def test_redshift_and_execute(self):
...
+
+        def factorial(n: i32) -> i32:
+            res = [B]1[/COLOR]
+            for i in [B]range[/COLOR][R](n)[/COLOR]:
+                res *= ([R]i+[/COLOR][B]1[/COLOR])
+            return [R]res[/COLOR]
+
+        def main() -> None:
+            [B]print[/COLOR][R]([/COLOR][B]factorial[/COLOR][R]([/COLOR][B]5[/COLOR][R]))[/COLOR]"""  # noqa
+        assert ansi_to_readable(stdout.strip()) == textwrap.dedent(expected_outout)
```
antocuni, line 167: I love this idea to test colors 👏 😍

spy/tests/test_cli.py
```diff
@@ -45,20 +59,35 @@ def init(self, tmpdir):
...
+                res *= (i+1)
+            return res
+
+        def main() -> None:
+            print(factorial(5))
+        """
+        self.main_spy.write(textwrap.dedent(main_src))
+        self.factorial_spy.write(textwrap.dedent(factorial_src))
+
+    def run(self, *args: Any, decolorize_stout=True) -> Any:
```
antocuni:
```suggestion
    def run(self, *args: Any, decolorize_stdout=True) -> Any:
```

spy/tests/test_cli.py
```diff
@@ -45,20 +59,35 @@ def init(self, tmpdir):
...
         args2 = [str(arg) for arg in args]
         print("run: spy %s" % " ".join(args2))
         res = self.runner.invoke(app, args2)
         print(res.stdout)
         if res.exit_code != 0:
             raise res.exception  # type: ignore
-        return res, decolorize(res.stdout)
+        return res, decolorize(res.stdout) if decolorize_stout else res.stdout
```
antocuni:
```suggestion
        return res, decolorize(res.stdout) if decolorize_stdout else res.stdout
```

spy/parser.py
```diff
@@ -697,7 +697,15 @@ def from_py_expr_UnaryOp(self, py_node: py_ast.UnaryOp) -> spy.ast.Expr:
             and isinstance(value, spy.ast.Constant)
             and isinstance(value.value, (int, float))
         ):
-            return spy.ast.Constant(value.loc, -value.value)
+            c_loc = value.loc
+            new_loc = Loc(
+                c_loc.filename,
+                c_loc.line_start,
+                c_loc.line_end,
+                c_loc.col_start - 1,  # handle the negative sign
```
antocuni: are we sure that -1 is enough?
What happens if I have e.g. `-    100` in the source code?

spy/util.py
```diff
@@ -118,6 +120,28 @@ def highlight_C_maybe(code: str | bytes) -> str:
     return highlight(code, CLexer(), TerminalFormatter())
 
 
+def highlight_spy_maybe(code: str | bytes) -> str:
```
antocuni: good idea to have syntax highlighting also for spy code!
  1. `highlight_C_maybe` and `highlight_spy_maybe` have a lot of duplicated code. Maybe we can turn them into a single `highlight_src_maybe("C", ...)`
  2. the spy version uses CLexer which is wrong. It should use PythonLexer? Also, it might be worth using the lexer for Python 2.7, because it automatically highlights text in backticks, which we emit a lot after redshifting
  3. we should call this function also when we do a plain `--redshift`

spy/util.py
```diff
@@ -118,6 +120,28 @@ def highlight_C_maybe(code: str | bytes) -> str:
...
+    from pygments.lexers import CLexer  # type: ignore
+
+    # Regex for ANSI escape sequences
+    regexp = re.compile(r"\033\[[0-9;]*m")  # \033 is \x1b
+
+    def has_ansi(text: str) -> bool:
+        return bool(regexp.search(text))
+
+    if not has_ansi(code):
+        return highlight(code, CLexer(), TerminalFormatter())
```
antocuni: I don't understand why this `has_ansi()` logic is needed?

antocuni (changes requested):
@ernestoarbitrio sorry for the late review, but I like this PR a lot.
I have left a few comments.

spy/util.py
ernestoarbitrio: because we prepend ansi while doing the src highlight and when we already have ansi on a line then pygments tries to re-apply the colors and it doesn't work anymore :D

spy/parser.py
ernestoarbitrio: you right, I've come with another solution, let me know what do you think

ernestoarbitrio: ae7de72d5ae744aeb8286d1da38d17a7dbdd3db5

antocuni: I think that the `re` based approach is a bit too complicated and probably brittle. You can just use the `py_node.loc` to precisely know the location of the `-` token. Something like this seems to work:
```python
            op_loc = py_node.loc
            c_loc = value.loc
            srcline = linecache.getline(c_loc.filename, c_loc.line_start)

            new_loc = Loc(
                c_loc.filename,
                op_loc.line_start,
                c_loc.line_end,
                op_loc.col_start,
                c_loc.col_end,
            )
```

But even better, we have `Loc.combine` which does exactly this work: take two locs and return a "bigger loc" which contains both:
```python
            new_loc = Loc.combine(op_loc, c_loc)
```

ernestoarbitrio: done thanks

spy/tests/test_parser.py
```diff
@@ -704,12 +712,21 @@ def foo() -> i32:
...
+    @pytest.mark.parametrize(
+        "srcline, expected_src, expected_value",
+        (
+            ("return -123 * -1.0", "-123", -123),
+            ("return -     123 * -1.0", "-     123", -123),
+            ("return (-         123) * -1.0", "-         123", -123),
+            ("return ((-   123) * -1.0)", "-   123", -123),
+        ),
+    )
+    def test_negative_const(self, srcline, expected_src, expected_value):
```
antocuni: I know this sounds a bit nitpicking, but I actually think it's very important for the long-term readability of tests.
I have seen this `parametrize` pattern multiple times, and in the long run it leads to unreadable test failures -- especially in CI -- because when you see a traceback it becomes very hard to understand what are the *values* you are operating on. In a case like this, you would see e.g. `assert const.value == expected_value` but then you look at the test ID to actually know what the test expects.

I tried to write an equivalent version which I find MUCH more readable:
```python
    def test_negative_const(self):
        # special case -NUM, so that it's seen as a constant by the rest of
        # the code
        mod = self.parse(f"""
        def foo() -> None:
            -100
            -  101
            (-    102)
        """)
        funcdef = mod.get_funcdef("foo")
        expected = """
        FuncDef(
            color='red',
            kind='plain',
            name='foo',
            args=[],
            vararg=None,
            return_type=Constant(value=None),
            docstring=None,
            body=[
                StmtExpr(
                    value=Constant(value=-100),
                ),
                StmtExpr(
                    value=Constant(value=-101),
                ),
                StmtExpr(
                    value=Constant(value=-102),
                ),
            ],
            decorators=[],
        )
        """
        self.assert_dump(funcdef, expected)
        assert funcdef.body[0].value.loc.get_src() == "-100"
        assert funcdef.body[1].value.loc.get_src() == "-  101"
        assert funcdef.body[2].value.loc.get_src() == "-    102"
```

antocuni (changes requested):
@ernestoarbitrio sorry, the other comment about `Loc.combine` was sent by mistake, it was supposed to be part of this review.

I think the PR is very good and ready to be merged, but see my comment about this test (the reason it came late is that I wanted to take my time to explain my reasoning but dinner was ready :)).

PR #268, opened 2025-10-22
Add unchecked division primitives

### LINKED ISSUE #237: Add unchecked division primitives
PR #235 adds "divide by zero" checks to all division operators.

We should introduce "unsafe unchecked divisions" for the rare cases in which they will be needed.
I propose to introduce `unchecked_div`, `unchecked_mod` etc. in the `unsafe` module.

Ideally, they should be overloaded so that they automatically work on any numerical type.
For inspiration on how to do an interp-level overload, we can look at the implementation of `print()`:
https://github.com/spylang/spy/blob/56795d6c3cfc06eb198bef52d8e26e920c1973ba/spy/vm/modules/builtins.py#L46-L81

Viriathus1:
resolves #237

### DIFF
--- a/spy/backend/c/cwriter.py
+++ b/spy/backend/c/cwriter.py
@@ -335,6 +335,7 @@ class CFuncWriter:
         #
         FQN("operator::f64_add"): "+",
         FQN("operator::f64_sub"): "-",
+        FQN("unsafe::f64_ieee754_div"): "/",
         FQN("operator::f64_mul"): "*",
         FQN("operator::f64_eq"): "==",
         FQN("operator::f64_ne"): "!=",
--- a/spy/vm/modules/unsafe/div.py
+++ b/spy/vm/modules/unsafe/div.py
@@ -147,3 +147,25 @@ def w_f64_unchecked_mod(vm: "SPyVM", w_a: W_F64, w_b: W_F64) -> W_F64:
     a = w_a.value
     b = w_b.value
     return vm.wrap(a % b)
+
+
+@UNSAFE.builtin_func(color="blue", kind="metafunc")
+def w_ieee754_div(vm: "SPyVM", wam_l: W_MetaArg, wam_r: W_MetaArg) -> W_OpSpec:
+    @vm.register_builtin_func("unsafe", "f64_ieee754_div")
+    def w_ieee754_div(vm: "SPyVM", w_a: W_F64, w_b: W_F64) -> W_F64:
+        a = w_a.value
+        b = w_b.value
+
+        if b == 0:
+            if a > 0:
+                result = float("inf")
+            elif a < 0:
+                result = float("-inf")
+            else:
+                result = float("nan")
+
+            return vm.wrap(result)
+
+        return vm.wrap(a / b)
+
+    return W_OpSpec(w_ieee754_div)


spy/tests/compiler/unsafe/test_div.py
```diff
@@ -0,0 +1,170 @@
...
+        from unsafe import unchecked_div, unchecked_floordiv, unchecked_mod
+        T = {int_type}
+        def div(x: T, y: T) -> T: return unchecked_div(x, y)
+        def floordiv(x: T, y: T) -> T: return unchecked_floordiv(x, y)
+        def mod(x: T, y: T) -> T: return unchecked_mod(x, y)
+        """)
+        try:
+            mod.div(11, 0)
+        except ZeroDivisionError:
+            pass
```
antocuni: I think that in interpreter mode, the unchecked division by 0 should still raise a `PanicError`.
The idea is this:
  - in interp mode, we always enable all the checks that we can do, even when using unsafe tools
  - in compiled DEBUG mode, we still do the check
  - in compiled RELEASE mode, it's actually unsafe.

For example, let's see what happens we `p` is an unsafe ptr and we  do `p[n]`.
In interp-mode we always check for the size of the buffer:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/vm/modules/unsafe/ptr.py#L186-L203

In compiled mode, it depends on `SPY_DEBUG`:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/libspy/include/spy/unsafe.h#L28-L32

The unchecked version is this:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/libspy/include/spy/unsafe.h#L45-L47

The checked version is this:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/libspy/include/spy/unsafe.h#L76-L81

(I realize now that we DO NOT check for NULL: we probably should, feel free to open an issue if you want :)).

So probably this test should be written similarly to this:
https://github.com/spylang/spy/blob/b88f8693da187567a58d25259ac9380bca07d2da/spy/tests/compiler/test_int.py#L67-L81

Sorry for giving extra work to do: I didn't think about this case at first, so thank you for bringing it to my attention 🙏

spy/tests/compiler/unsafe/test_div.py
```diff
@@ -0,0 +1,170 @@
...
+
+class TestUnsafeFloatDiv(CompilerTest):
+    def test_unchecked_div(self):
+        mod = self.compile("""
+        from unsafe import unchecked_div
+        def div(x: f64, y: f64) -> f64: return unchecked_div(x, y)
+        """)
+        assert mod.div(1.5, 2.0) == 0.75
+        assert mod.div(643.2, 51.2) == 12.5625
+        assert mod.div(500.000034, 45.000034) == 11.111103471610711
```
antocuni: this is a small nitpick, but did you choose these numbers for any particular reason?
I usually prefer tests which are "immediately obvious" (e.g. 11/2 == 5.5), and when reading this one I don't really know whether the result is correct without computing it somewhere else.

If it is to check that we have enough precision, I suggest to writ it as `assert mod.div(500.000034, 45.000034) == 500.000034 / 45.000034`.

If it's not about the precision, I think we should just pick easier numbers (same for 643.2 and 51.2)

spy/tests/compiler/unsafe/test_div.py
antocuni: unless there are good real life cases in which we actually *want* to define float division by 0 as returning `inf`. In that case, we should probably add the check that condition in the interpreter and explicitly return `inf`.
So maybe we might want also an `ieee754_div`? :thinking: 

E.g., this is how numpy works when operating on its own float64 numbers, but it also raises a warning, so probably they regretted doing this choice?
```
>>> np.float64(1) / np.float64(0)
<stdin>:1: RuntimeWarning: divide by zero encountered in scalar divide
np.float64(inf)
```

I am also fine to accept a PR which just panics for now, and leave this discussion for another issue/PR.

spy/tests/compiler/unsafe/test_div.py
Viriathus1: I selected those cases to test float precision but I get what you're saying about being immediately obvious. I've updated most of the test cases now in that spirit

spy/tests/compiler/unsafe/test_div.py
Viriathus1: @antocuni thanks that's a really clear write up 🙂 so all in all quite similar to the normal operators except it raises a `PanicError` at least for interp and compiled debug mode. 

It seems it's not uncommon to return `inf` in this context apparently Java and Javascript both do their version of this from a very brief online check. Given that this is the expected behaviour of numpy as well then it's probably worth providing such a function since it would effectively be opt in anyway. I'm happy to add that functionality too.

antocuni: yes exactly. Basically, they should be equal to the checked version, but the C code should have an `#ifdef SPY_DEBUG` around the `if (y == 0)`.

> Given that this is the expected behaviour of numpy as well then it's probably worth providing such a function since it would effectively be opt in anyway. I'm happy to add that functionality too.

that would be awesome, thank you

spy/tests/compiler/unsafe/test_div.py
antocuni: Have you by chance forgot to push? Because I don't seem to see any new commit since my last review?

Viriathus1: @antocuni I've done most of it now but I still have to implement `ieee754_div`

antocuni: @Viriathus1 ah ok, I was ready to merge the PR but then I'll wait for `ieee754_div`.
Note that I pushed a commit to add a small note to `TestUnsaveIntDiv` to explain the SPY_RELEASE/SPY_DEBUG distinction, so you might have to rebase/merge.
Thank you!

antocuni (approved):
perfect, thank you!

PR #270, opened 2025-10-25
Auto-publish Playground 

### LINKED ISSUE #169: Automatically build & publish the "SPy playground"
The SPy playground is a little toy which I wrote some time ago to run the SPy pipeline in the browser (using PyScript/pyodide) and see the various steps:
https://antocuni.pyscriptapps.com/spy-playground/latest/
https://pyscript.com/@antocuni/spy-playground/latest?files=main.py

It would be nice to have it automatically built at every commit, and published to some github page.
Any improvement on the UI/UX is also very welcome :).

To run SPy on pyodide you need two components:
1. a `spy-*.whl` wheel
2. `libspy.wasm` and `libspy.mjs` which are built by running `make -C spy/libspy`, and found in `spy/libspy/build/emscripten`.

aisipos:
Fixes #169 

This PR implements a Pyodide based "Playground" for SPy that is deployed to Github pages on each commit to main on this repo.

I followed these steps in the making of this PR:
-  Imported all the files from the existing "Playground" project at https://pyscript.com/@antocuni/spy-playground/latest?files=main.py.
- Created a Github workflow action to build and publish the playground to https://spylang.github.io/spy
- Fixed out of date examples in the old playground to work with current SPy. I also removed `smallpoint.spy` since it had a warning not to use it. We can add this back if we want to get it working.
- Fixed wheel build on emscripten to exclude non-pure python libraries: ninja, pexpect.
- Added a `--colorize` button to use the new feature added in #266. Works great even in the browser! 🎉 

You can preview what the playground looks like at https://aisipos.github.io/spy/ . Once merged here it will publish automatically to https://spylang.github.io/spy.

Some nice to haves that I'll leave for future PRs, since I don't want to blow scope for this PR:
- Symlink `playground/` files to `examples/` files if they are a match rather than keeping them separate.
- Integrate the playground examples into the test suite, so we ensure they don't break.
- UI enhancements.

### DIFF
--- a/playground/main.py
+++ b/playground/main.py
@@ -117,6 +117,7 @@ def main():
                 RunSPyButton("--redshift"),
                 RunSPyButton("--redshift --no-pretty"),
                 RunSPyButton("--cwrite"),
+                RunSPyButton("--colorize"),
             ).css(
                 {
                     "display": "flex",


aisipos:
> Wow, that's wonderful, thank you!
> 
> I was going to merge it but then I got a doubt: where are the github pages actually stored? Years ago they always ended up in the `gh-pages` branch of the repo, and I'm a bit worried that if we upload a new binary `spylang.whl` and `libspy.wasm` in every commit, then then spy repo will quickly become very big to clone.
> 
> But at the same time, I see that your repo `aisipos/spy` does not seem to contain a gh-pages branch, so maybe now things are different?

I don't know all the internals exactly of Github pages, but there is a separation of "build products" (which they call artifacts) of a Github action and the Git repository itself. So `spylang.whl` and `libspy.wasm` are built by Github actions but don't become part of the Git repo itself, they are artifacts and are stored separately. You can see the artifacts in the details page of the build-playground Github action: https://github.com/aisipos/spy/actions/runs/18891385245, although they get deleted after a short time. The deletion of the artifact doesn't affect your deployed Github page, once you go through the "publish" action the artifact goes I suppose to their CDN and persists until you do another deployment.

image


There are limits to the size and frequency of generating artifacts, see here https://docs.github.com/en/pages/getting-started-with-github-pages/github-pages-limits . Quoting:
```
- GitHub Pages source repositories have a recommended limit of 1 GB. 
- Published GitHub Pages sites may be no larger than 1 GB.
- GitHub Pages deployments will timeout if they take longer than 10 minutes.
- GitHub Pages sites have a soft bandwidth limit of 100 GB per month.
- GitHub Pages sites have a soft limit of 10 builds per hour. This limit does not apply if you build and publish your site with a custom GitHub Actions workflow.
```

I think the size limits / builds per hour is fine for us. I don't anticipate the bandwidth is a problem either, although if it ever does, assuming it was legitimate traffic, that would become a "nice problem to have" as they say. 😄

aisipos:
Another way of looking at this, the Github pages site as implemented by the workflow actions is a static website served by the "merging" of what is in the `playground/` directory of this repo, and the files generated by the `build_playground` workflow step. See : https://github.com/spylang/spy/blob/812fd328d3362f7872b419dac071e981498baac4/.github/workflows/playground.yml#L56-L58
These files are built on the fly and don't become part of the Git repo, they're considered "artifacts" in the Github sense, although they do have their own size quotas.

The deploy workflow step then is what pushes the new version to github pages, which will then be world visible: https://github.com/spylang/spy/blob/812fd328d3362f7872b419dac071e981498baac4/.github/workflows/playground.yml#L78-L81

antocuni:
@aisipos thank you for the detailed explanation, everything makes a lot of sense. Merging!

PR #271, opened 2025-10-29
Feature/add support for chained comparisons

### LINKED ISSUE #249: Add support for chained comparisons
```
def main() -> None:
    if 1 < 2 < 3:
        print('hello')
```

```
❯ spy /tmp/x.spy 
ParseError: not implemented yet: chained comparisons
   --> /tmp/x.spy:2:16
  2 |     if 1 < 2 < 3:
    |                ^ this is not supported
```

Jemeljanov:
Closes #249 

## Summary
- implement chained comparisons
- handle mixed red/blue operands in different cases
- implement the support in the C backend
- add tests

I would appreciate a review when you have a moment.

### DIFF
--- a/spy/tests/test_doppler.py
+++ b/spy/tests/test_doppler.py
@@ -206,6 +206,9 @@ class TestDoppler:
         self.redshift(src)
         self.assert_dump(src)
 
+    @pytest.mark.xfail(
+        reason="CmpChain truncation is temporary; will disappear once chained comparisons are desugared"
+    )
     def test_chained_comparison_red_operands(self):
         src = """
         def foo(x: i32, y: i32) -> bool:
@@ -214,6 +217,9 @@ class TestDoppler:
         self.redshift(src)
         self.assert_dump(src)
 
+    @pytest.mark.xfail(
+        reason="CmpChain truncation is temporary; will disappear once chained comparisons are desugared"
+    )
     def test_chained_comparison_red_then_blue(self):
         src = """
         def foo(x: i32) -> bool:
@@ -222,6 +228,9 @@ class TestDoppler:
         self.redshift(src)
         self.assert_dump(src)
 
+    @pytest.mark.xfail(
+        reason="CmpChain truncation is temporary; will disappear once chained comparisons are desugared"
+    )
     def test_chained_comparison_short_circuit_preserves_red(self):
         src = """
         def foo(x: i32) -> bool:
@@ -230,6 +239,9 @@ class TestDoppler:
         self.redshift(src)
         self.assert_dump(src)
 
+    @pytest.mark.xfail(
+        reason="CmpChain truncation is temporary; will disappear once chained comparisons are desugared"
+    )
     def test_chained_comparison_red_tail(self):
         src = """
         def foo(x: i32, y: i32) -> bool:
@@ -238,6 +250,9 @@ class TestDoppler:
         self.redshift(src)
         self.assert_dump(src)
 
+    @pytest.mark.xfail(
+        reason="CmpChain truncation is temporary; will disappear once chained comparisons are desugared"
+    )
     def test_chained_comparison_mixed_long(self):
         src = """
         def foo(a: i32, b: i32, c: i32) -> bool:
@@ -246,6 +261,9 @@ class TestDoppler:
         self.redshift(src)
         self.assert_dump(src)
 
+    @pytest.mark.xfail(
+        reason="CmpChain truncation is temporary; will disappear once chained comparisons are desugared"
+    )
     def test_chained_comparison_short_circuit_shift(self):
         src = """
         def foo(x: i32, y: i32) -> bool:


spy/backend/spy.py
```diff
@@ -367,6 +367,21 @@ def fmt_expr_CmpOp(self, op: ast.CmpOp) -> str:
...
+        if first.left.precedence < chain.precedence:
+            left_str = f"({left_str})"
+        pieces.append(left_str)
+        for cmp in chain.comparisons:
+            right_str = self.fmt_expr(cmp.right)
+            if cmp.right.precedence < chain.precedence:
+                right_str = f"({right_str})"
+            pieces.append(f"{cmp.op} {right_str}")
+        return " ".join(pieces)
+
```
antocuni, line 384: this should have a dedicated test in `test_backend_spy`.

spy/tests/test_cli.py
```diff
@@ -133,6 +133,64 @@ def test_execute(self):
         _, stdout = self.run(self.main_spy)
         assert stdout == "hello world\n"
 
+    def test_execute_chained_comparison(self):
```
antocuni: test_cli is not the right place where to put these tests.
They should be in `test_basic.py`: this way, each test is automatically run in three modes:
- `[interp]`: direct interpretation
- `[doppler]`: redshift and then interpret the result
- `[C]`: compile to C and execute

If you do so, you will also notice that the `[C]` tests fail because you need to add proper support to `ast.CmpChain` to the C backend.

antocuni: Moreover, I think you should explicitly check two important properties of chained comparisons:
  - each operand is evaluated only once
  - the chain is short circuit, so as soon as one comparison is `False`, the later operands are not evaluated at all

The easiest way to check for side effects to to increment a global variable inside a function call, as we do e.g. in `test_StmtExpr`:
https://github.com/spylang/spy/blob/f82ef21854940e4778c3dd9982e46133b6b5dd87/spy/tests/compiler/test_basic.py#L425-L436

spy/vm/astframe.py
```diff
@@ -803,6 +803,39 @@ def eval_expr_CmpOp(self, op: ast.CmpOp) -> W_MetaArg:
         w_opimpl = self.vm.call_OP(op.loc, w_OP, [wam_l, wam_r])
         return self.eval_opimpl(op, w_opimpl, [wam_l, wam_r])
 
+    def eval_expr_CmpChain(self, chain: ast.CmpChain) -> W_MetaArg:
+        assert chain.comparisons, "CmpChain must contain at least one comparison"
+        first_cmp = chain.comparisons[0]
+        prev_expr = first_cmp.left
+        wam_prev = self.eval_expr(prev_expr)
+        result: Optional[W_MetaArg] = None
```
antocuni: the naming convention is that all `W_MetaArg` variables start with `wam` (for Wrapped Argument Meta, I chose `wam` because it's much easier to "pronounce mentally" than `wma` :)).

So this should be `wam_result`

spy/vm/astframe.py
```diff
@@ -803,6 +803,39 @@ def eval_expr_CmpOp(self, op: ast.CmpOp) -> W_MetaArg:
...
+        saw_red = False
```
antocuni: I'm not sure to understand why you need `saw_red`.
If I understand correctly, what it does is to ensure that the final result is red if at least one of the individual operands is red. But I think this happens automatically because if you cal `eval_opimpl` with mixed-color operands, you get a red result anyway.

FWIW, I tried to comment out all the usage of `was_red` in the code and all your tests seem to pass.

spy/vm/astframe.py
```diff
@@ -803,6 +803,39 @@ def eval_expr_CmpOp(self, op: ast.CmpOp) -> W_MetaArg:
...
+
+        for cmp in chain.comparisons:
+            assert cmp.left is prev_expr, "chained comparisons must share operands"
+            wam_right = self.eval_expr(cmp.right)
+            w_OP = OP_from_token(cmp.op)
+            w_opimpl = self.vm.call_OP(cmp.loc, w_OP, [wam_prev, wam_right])
+            result = self.eval_opimpl(cmp, w_opimpl, [wam_prev, wam_right])
+            if result.color == "red":
+                saw_red = True
+            if result._w_val is not None:
```
antocuni: this check is also suspicious, I don't think that checking for `_w_val` is the right thing to do.
I **think** that the right logic is:
  - if we are redshifting, we want to evaluate all the operands without bothering to check their result (because we are doing abstrat interpretation and we want to ensure that we are evaluating all possible paths)
  - if we are not redshifting, then we *know for sure* that `result._w_val` is not None, and we can just access `result.w_val`.

antocuni (commented):
thank you @Jemeljanov!
This feature is not easy and what you achieved so far is impressive :clap: .

See the inline comments for more details, but I think this is a list of what we need before we can consider it mergeable:
1. better tests: you should write them in `test_basic.py`, and remove those from `test_cli.py`
2. fix/refactor `astframe.eval_expr_CmpChain` (see inline comments)
3. implement the support in the C backend.

Doing (3) is not straighforward because C doesn't have the right syntax to represent chained operations.
Imagine this:
```python
def inc(x: int) -> int:
    print('inc ' + str(x))
    return x + 1

def main() -> None:
    cond = inc(4) < inc(5) < inc(6)
    print(cond)
```
we want to print `inc 5` only ONCE.

One easy way to implement this in C is to save the result of cmp operands into tmp variables; something like this:
```c
_cmp0 = inc(4);
_cmp1 = inc(5);
_cmp_res = _cmp0 < _cmp1;
if (_cmp_res) {
    _cmp2 = inc(6);
    _cmp_res = _cmp1 < _cmp2;
}
```

Unfortunately, this approach doesn't work with the current state of the SPy C backend. The problem is that because of the way it's written, the C backend assumes that every `ast.Expr` becomes a C expression, and the code above is not an expression because it contains statements.

I *think* that the easiest way to implement without changing too much is to use the C comma `,` operator, which allows you to have multiple assignments in the same expression.
So the code above should become something like this:

```c
(_cmp0 = inc(4), _cmp1 = inc(5), _cmp0 < _cmp1) && (_cmp2 = inc(6), _cmp1 < _cmp2)
```

Jemeljanov:
@antocuni thank you for the earlier feedback, I’ve pushed another round of updates. 

This got trickier than I first expected so please let me know if you would like to make changes.

antocuni:
@Jemeljanov I MIGHT have had an idea, but it's 1am for me and I might be hallucinating, I'll need to think better tomorrow :).

We could make chained comparisons working in the following way:

1. first, we need #264 to be merged
2. then, we need to implement the walrus operator `:=` (which might be non-trivial, I need to think about it)
3. once we have that, we can use a "desugaring" logic similar to what we do for `for` loops, and desugar this expression:
```
a < b < c
```

into this:
```
(_cmp0 := a) < (_cmp1 := b) and _cmp1 < (_cmp2 := c)
```

i.e., the same trick that I proposed to use for the C backend, but done directly inside ASTFrame.

This way, chained comparisons should work out of the box in doppler and C.

But we need `and` and `:=` before we can do that.

antocuni:
@Jemeljanov so, after thinking a lot about this, I think it's better to close this PR and redo it later when we have the needed tooling, in particular:
- walrus operator: #313 
- `and`: #264 

Once we have those, implementing this will become much easier, because in theory it will be enough to desugar the chained comp in the ASTFrame and it will automatically work out of the box in `doppler` and `C`, so much less code will be needed.

PR #273, opened 2025-10-30
Colorize local assignment stmt

ernestoarbitrio:


spy/doppler.py
```diff
@@ -181,13 +182,20 @@ def shift_stmt_VarDef(self, vardef: ast.VarDef) -> list[ast.Stmt]:
         return [vardef.replace(type=newtype, value=newvalue)]
 
     def shift_stmt_Assign(self, assign: ast.Assign) -> list[ast.Stmt]:
+        def populate_color(color: Color) -> None:
+            if self.vm.ast_color_map is not None:
+                self.vm.ast_color_map[assign] = color
```
antocuni: a small nitpick, but since we need this logic in a few places (two here and later in `eval_expr`), it might be worth to introduce a method `DopplerFrame.record_node_color(node, color)` which does that.

spy/vm/vm.py
```diff
@@ -102,7 +102,7 @@ class SPyVM:
     bluecache: BlueCache
     emit_warning: Callable[[SPyError], None]
     # For use by --colorize to remember the red/blue color of each expr
-    expr_color_map: Optional[dict[ast.Expr, Color]]
+    ast_color_map: Optional[dict[ast.Expr | ast.Assign, Color]]
```
antocuni: I think we can just use `ast.Node` here, as it's the common superclass of both.
Also, I suspect that in the future we will add more `blue` statements (e.g. I plan to introduce `for x in UNROLL(some_blue_iterabale)` which will be unrolled during redshift).

spy/doppler.py
ernestoarbitrio: done

spy/vm/vm.py
ernestoarbitrio: done

antocuni:
@ernestoarbitrio merged, thank you!

PR #274, opened 2025-10-30
Fix Hardcoded Playground libspy location

aisipos:
Fix issues relating to a missed hardcoded location for libspy in the `playground/`.

## Use built version of libspy.mjs
Manually create an absolute URL for our built version of `libspy.mjs` based on the browsers URL. I tried first to use a relative link `./libspy.mjs`, but pyodide in those cases tries to load it from their own cdn: 
```
pyodide.ffi.JsException: TypeError: Failed to fetch dynamically imported module: https://cdn.jsdelivr.net/pyodide/v0.26.4/full/libspy.mjs
```
Creating an absolute URL manually from the browser's URL feels a little hacky, I didn't immediately see a more correct way to do this using PyOdide functionality, but I can look further if need be.

I verified this worked locally in the Chrome devtools network panel
image

## Fix compatibility with newer emscripten SDKs
Once I fixed the hardcoded `libspy.mjs` issue, I ran into a new issue:
```
  File "/lib/python3.12/site-packages/spy/llwasm/emscripten.py", line 68, in __init__
    self.mem = LLWasmMemory(self.instance.HEAP8)
                            ^^^^^^^^^^^^^^^^^^^
  File "http://localhost:8000/libspy.mjs", line 399, in Object.get
  File "http://localhost:8000/libspy.mjs", line 532, in abort
pyodide.ffi.JsException: RuntimeError: Aborted('HEAP8' was not exported. add it to EXPORTED_RUNTIME_METHODS (see the Emscripten FAQ))
```
This seems to be a break that happened in version 2.0.10 of emscripten, and was noted in this Google groups discussion: https://groups.google.com/g/emscripten-discuss/c/Qlob7yyD94U?pli=1.
I fixed it by adding `-sEXPORTED_RUNTIME_METHODS=HEAP8` to the `LDFLAGS` of  emscripten builds. This does fix the issue, but if there are more comprehensive solutions we can look into them.

### DIFF
--- a/spy/libspy/Makefile
+++ b/spy/libspy/Makefile
@@ -84,7 +84,8 @@ else ifeq ($(TARGET), emscripten)
 		-mbulk-memory
 
 	LDFLAGS := \
-		-sWASM_BIGINT
+		-sWASM_BIGINT \
+		-sEXPORTED_RUNTIME_METHODS=HEAP8
 
 	SRCS := \
 		$(SRCS) \


playground/Makefile
```diff
@@ -21,6 +21,9 @@ WHEEL_NAME := spylang-$(VERSION)-py3-none-any.whl
 # Build playground for local development
 # This mirrors the steps from .github/workflows/playground.yml
 local:
+	@echo "Cleaning libspy..."
+	$(MAKE) -C $(LIBSPY_DIR) clean
+
```
antocuni: I don't fully understand why a simple `make` is not enough.
In theory, `make` should automatically rebuild when necessary: if this doesn't happen, maybe it means that we should fix libspy makefile?

Note that I'm happy add the `make clean` as a temporary workaround if needed, but I'd like to understand why it's needed (and possibly insert a comment to explain)

aisipos: I agree that perhaps adding clean is overkill. It was really for my particular use case here, where I edited the `Makefile` for libspy, but changing the `Makefile` won't actually cause a rebuild. I think you're right, it's a rare enough case that the developer should manually need to clean. I removed this change and did a force push.

PR #279, opened 2025-11-05
Dont leak struct fields into symtables

### LINKED ISSUE #231: struct fields leak into the symtables of methods
Consider the following:

```python
@struct
class Point:
    x: int
    y: int

    def __new__(x: int, y: int) -> Point:
        return Point.__make__(x*2, y*2)


def main() -> None:
    print(Point(1, 2))
```

It fails because inside `__new__` it thinks that `x` has been declared in an outer scope, but struct fields should not be visible as normal declarations.

```
❯ spy  /tmp/x.spy 
ScopeError: variable `x` shadows a name declared in an outer scope
   --> /tmp/x.spy:6:17
  6 |     def __new__(x: int, y: int) -> Point:
    |                 |____| this is the new declaration

   --> /tmp/x.spy:3:5
  3 |     x: int
    |     |____| this is the previous declaration
```

aisipos:
Fixes #231 

I should say first that I while I think the approach currently in this PR appears to fix the problem in the ticket (struct fields should not clash with method argument names), my approach does feel wrong somehow to me.

I solved the issue for "structs" only, by adding a new `VarKindOrigin` of `struct-field`, and then treating them specially in `ScopeAnalyzer.define_name`, essentially ignoring name clashes in other levels with an origin struct-field.

My first question of course is, should we also treat all class fields the same, i.e. not treat structs as a separate case? I was perhaps reading the original ticket too literally. If so, the new `struct-field` origin type is not appropriate yet. Also, do you think the approach of detecting conflicting names in `define_name` is the proper solution here? Again, it "works" in the sense that it passes all the tests, but I'm not sure if it is the right place to address the problem.

### DIFF
--- a/spy/analyze/scope.py
+++ b/spy/analyze/scope.py
@@ -134,6 +134,9 @@ class ScopeAnalyzer:
         towards the outer.
         """
         for level, scope in enumerate(reversed(self.stack)):
+            if level > 0 and scope.kind == "class":
+                # jump over 'class' scopes
+                continue
             if sym := scope.lookup_maybe(name):
                 return level, scope, sym
         # not found
--- a/spy/tests/compiler/test_struct.py
+++ b/spy/tests/compiler/test_struct.py
@@ -481,6 +481,8 @@ class TestStructOnStack(CompilerTest):
             p = Point(1, 2)
             p.foo()
         """
-        mod = self.compile(src)
-        with pytest.raises(NameError):
-            mod.main()
+        errors = expect_errors(
+            "name `x` is not defined",
+            ("not found in this scope", "x"),
+        )
+        self.compile_raises(src, "main", errors)


antocuni:
@aisipos thank you for dealing with it.
I tend to agree with you, your current solution seems to kind-of-work but it feels more "curing the symptom" than "curing the disease".

It's probably my fault because I didn't think it too deeply when adding support for the `ClassDef` statement.

The logic in ScopeAnalyzer must match the runtime logic which will be implemented by ASTFRame&co.
The original ScopeAnalyzer works with these assumptions:
1. Each scope has a parent (apart the root, which is `builtins`)
2. Each scope can also access the variables defined in its enclosing scopes

This works very well for modules and nested functions:
```python
# enter module scope
A = 0
def foo():
    # enter "foo" scope
    b: int   # declare a variable inside "foo scope"
    b = 0
    def bar():
        # enter "foo::bar" scope
        c = 0  # declare a variable inside "foo::bar scope"
```

Here, in `bar` I can access to `c`, `b` and `A`.

However, `class` breaks this simple assumption:

```python
# enter module scope
A = 0
class Foo:
    # enter "Foo" scope

    # this does NOT declare a variable in "Foo scope".
    b: int

    def meth(self):
        print(A) # this works
        print(b) # this SHOULD BE a NameError
```

The problem is that the way the current code works, `ClassDef` defines a scope (which is right -- we need a scope where `ClassFrame` can execute its statements), but then `meth` is enclosed in "Foo scope", **which is wrong**.

Ideally, the logic should be:
1. when we analyze the ClassDef, we create a scope for it (this is currently done right)
2. when we analyze the FuncDef, it should be enclosed in the MODULE scope, not in the class scope.

Another example of the same bug is this:
```
@struct
class Point:
    x: i32
    y: i32

    def foo(self: Point) -> None:
        return x


def main() -> None:
    p = Point(1, 2)
    p.foo()
```

Here, we should expect that `p.foo()` raises `NameError`, because `x` is clearly not defined.
However, because of this bug, spy thinks that `foo` has access to the `x` defined in its enclosing scope:
```
❯ spy --symtable /tmp/x.spy 
<symbol table 'x'>
    [0] const Point       
    [0] const main        

<symbol table 'x::Point'>
    [0] const foo         
    [0] var   x           
    [0] var   y           
    [1] const Point       
    [2] const i32          => <ImportRef builtins.i32>

<symbol table 'x::Point::foo'>
    [0] var   self        
    [0] var   @return     
    [1] var   x              # THIS IS WRONG!

<symbol table 'x::main'>
    [0] const p           
    [0] var   @return     
    [1] const Point       
```

The line which I labeled as "this is wrong" says that `foo` can access a variable `x` defined "one level up".
If you try to execute the code, you get a crash in ASTFrame (which is a bug of course):
```
  [...]
  File "/home/antocuni/anaconda/spy/spy/vm/astframe.py", line 779, in eval_expr_NameOuterDirect
    assert w_val is not None
           ^^^^^^^^^^^^^^^^^
AssertionError
```


So, I THINK (or at least, I hope, unless we find other corner cases which I cannot think about now) that the correct solution is:

1. when entering a ClassDef, we push a scope as now
2. we add a new `Symtable.kind` attribute which keeps track whether it's a symtable for a module, for a class or for a funcdec. Basically, `Symtable.kind` should match the kind of Frame it will be executed on (`ModFrame`, `ClassFrame` and `ASTFrame`).
3. when we do `lookup_ref` and/or `lookup_definition`, we IGNORE Symtables whose kind is `classdef`.

aisipos:
@antocuni Hmmm, I've tried a couple of things so far with no luck in making a proper fix, other than my initial hack. 

So far, I'm not sure we can fix this issue solely at `ScopeAnalyzer` time,  I think we might also need to make changes to the name lookups in  the `astframe` `eval_expr_` functions.  Would you be open to that?

I guess in the larger sense I'm wondering how much you'd like to mimic CPython's behavior, which is rather more complicated so far than SPy's? SPy's doesn't promise to be Python compatible, but Python's special treatment of class scopes is a large and rather distinct language feature that it seems we'd like to mimic, but does add implementation complexity. I used DeepWiki to do a "Codemap" of CPython's symbol table logic and class scope special casing: https://deepwiki.com/search/describe-cpythons-symbol-table_5a5e1ceb-d494-42ec-8e4c-5faebb18e6d3 . But before I go deeper into that investigation I wanted to see your thoughts on how much you wanted to match that implementation.

 I'll still keep looking a little deeper and see if I can make a simple fix that achieves proper class scoping separation.

Issue #175, opened 2025-06-06
Implement the ** operator


antocuni:

emanuelegurini:
Hello, may I work on this issue?

antocuni:
Done!

emanuelegurini:
@antocuni Hi Antonio, apologies for the delay — it’s been a busy period. I’ve been making progress little by little. Would it be alright if I continue working on it and share an update shortly? I don’t expect it will take much longer.

tedpatrick:
I have reworked my PR to support f32 and addressed all feedback in PR from @Viriathus1

### Addressed race condition in tests: 
pytest-xdist runs test workers in parallel. Two workers could race on stdlib/__pycache__/_slice.spyc, one opens it for writing (creating an empty file), while another immediately sees it exists with a newer mtime and tries to read it, getting EOFError: Ran out of input.

Fix (spy/analyze/importing.py:184): Write to a temp file (.{basename}.tmp) and then rename() atomically. On POSIX, rename is atomic, readers either see the old complete file or the new complete file, never a partial write.

antocuni:
Closing as it was fixed by #392

PR #280, opened 2025-11-06
feat:add power operator (**) for integers and floats

emanuelegurini:
#175 

- Added support for the power operator for both i32 and f64 types
- Implemented native functions spy_operator$f64_pow and spy_unsafe$f64_unchecked_pow in C
- Updated tests to cover power cases, roots, negative exponents, and implicit conversions between integers and floats
- Created a comprehensive example in pow.spy demonstrating various use cases
- Verified correct error handling (negative exponents on zero, fractional powers of negative numbers)

### DIFF
--- a/spy/vm/modules/operator/binop.py
+++ b/spy/vm/modules/operator/binop.py
@@ -32,6 +32,7 @@ MM.register("<" , "i8", "i8", OP.w_i8_lt)
 MM.register("<=", "i8", "i8", OP.w_i8_le)
 MM.register(">" , "i8", "i8", OP.w_i8_gt)
 MM.register(">=", "i8", "i8", OP.w_i8_ge)
+MM.register("**", "i8", "i8", OP.w_i8_pow)
 
 # u8 ops
 MM.register("+",  "u8", "u8", OP.w_u8_add)
@@ -51,6 +52,7 @@ MM.register("<" , "u8", "u8", OP.w_u8_lt)
 MM.register("<=", "u8", "u8", OP.w_u8_le)
 MM.register(">" , "u8", "u8", OP.w_u8_gt)
 MM.register(">=", "u8", "u8", OP.w_u8_ge)
+MM.register("**", "u8", "u8", OP.w_u8_pow)
 
 # i32 ops
 MM.register("+",  "i32", "i32", OP.w_i32_add)


spy/tests/compiler/test_float.py
Viriathus1, line 1: - No unit tests for exception handling of ZeroDivisionError or PanicError. 
- No unit tests for int types which should be in `test_int.py`.

spy/tests/compiler/test_float.py
```diff
@@ -116,3 +129,29 @@ def to_f64(x: i32) -> f64:
             return res
         """)
         assert mod.to_f64(42) == 42.0
+
+    def test_pow_special_values(self):
+        mod = self.compile("""
+        def pow(x: f64, y: f64) -> f64: return x ** y
+        """)
+
+        result = mod.pow(-1.0, float("nan"))
```
Viriathus1, line 138: This is bit of nitpick testing but I believe the following is better for readability and is used in the `unsafe/test_div.py` 
```suggestion
        assert isnan(mod.pow(-1.0, float("nan")))
```

spy/vm/modules/operator/opimpl_int.py
```diff
@@ -41,6 +41,10 @@ def _unary_op(vm: "SPyVM", w_a: WT, fn: Any) -> Any:
     def w_add(vm: "SPyVM", w_a: WT, w_b: WT) -> WT:
         return _binop(vm, w_a, w_b, lambda a, b: a + b)
 
+    @OP.builtin_func(f"{T}_pow")
+    def w_pow(vm: "SPyVM", w_a: WT, w_b: WT) -> WT:
```
Viriathus1: Need conditional guards here to raise relevant exceptions.

examples/pow.spy
Viriathus1, line 1: The issue here is that the int/float literals will be evaluated when redshifting ie compile time. The correct way to showcase this is to replicate what's done in unit tests which is to define a function that takes in args and use `**` operator on them.

spy/libspy/include/spy/operator.h
Viriathus1, line 1: - `pow` operation hasn't been defined for `u8`, `i8`, `i32` types.
- an unchecked operation has been added here but you've missed adding anything to the unsafe module in the vm and also there aren't any unit tests.

Viriathus1 (commented):
Hi @emanuelegurini, thanks for the PR and @antocuni will take a closer look at it but there were just some things that I noticed which I've outlined in the following comments.

spy/vm/modules/operator/opimpl_f64.py
```diff
@@ -43,6 +43,11 @@ def w_f64_mul(vm: "SPyVM", w_a: W_F64, w_b: W_F64) -> W_F64:
     return _f64_op(vm, w_a, w_b, lambda a, b: a * b)
 
 
+@OP.builtin_func
+def w_f64_pow(vm: "SPyVM", w_a: W_F64, w_b: W_F64) -> W_F64:
+    return _f64_op(vm, w_a, w_b, lambda a, b: a**b)
```
Viriathus1, line 48: Need conditional guards here to raise relevant exceptions which you defined in operator.h file.

spy/vm/modules/operator/opimpl_int.py
```diff
@@ -41,6 +41,10 @@ def _unary_op(vm: "SPyVM", w_a: WT, fn: Any) -> Any:
...
+        return _binop(vm, w_a, w_b, lambda a, b: a**b)
```
Viriathus1, line 46: Need conditional guards here to raise relevant exceptions.

examples/pow.spy
```diff
@@ -0,0 +1,27 @@
...
+    # Negative exponents
+    print(2.0 ** -2) 
+    
+    # Zero exponent
+    print(5.0 ** 0) 
+
+def main() -> None:
+    print("Power operator examples:")
+    print("---------------------")
+    show_pow_examples()
```
antocuni, line 27: thanks for the example, but I don't think that it's actually needed. I think I prefer to keep in `examples` things which shows spy's unique features or peculiarities.
If we add an example for every single feature of SPy, the `examples` directory would become a duplicate of tests :).

spy/tests/compiler/test_float.py
```diff
@@ -99,13 +106,19 @@ def test_implicit_conversion(self):
...
         def add_i8(x: f64, y: i8) -> f64: return x + y
         def add_u8(x: f64, y: u8) -> f64: return x + y
         """)
         assert mod.add(1.5, 2) == 3.5
         assert mod.sub(10, 0.5) == 9.5
         assert mod.div(1.5, 2) == 0.75
+        assert mod.pow(1.5, 2) == 2.25  # float base, int exponent
+        assert (
+            mod.pow_int_base(2, 0.5) == 1.4142135623730951
+        )  # int base, float exponent (square root)
```
antocuni, line 121:
```suggestion
        assert (
            mod.pow_int_base(2, 0.5) == math.sqrt(2)
        )  # int base, float exponent (square root)
```

I think this is much more readable and "obviously correct"

spy/tests/compiler/test_float.py
antocuni: yes +1 to @Viriathus1 suggestion here, I think that `isnan` is better

spy/vm/modules/operator/opimpl_dynamic.py
```diff
@@ -28,6 +28,11 @@ def w_dynamic_add(vm: "SPyVM", w_a: W_Dynamic, w_b: W_Dynamic) -> W_Dynamic:
     return _dynamic_op(vm, OP.w_ADD, w_a, w_b)
 
 
+@OP.builtin_func
+def w_dynamic_pow(vm: "SPyVM", w_a: W_Dynamic, w_b: W_Dynamic) -> W_Dynamic:
+    return _dynamic_op(vm, OP.w_POW, w_a, w_b)
+
+
```
antocuni, line 35: if I'm not mistaken, this is missing a test in `test_dynamic.py`.
It seems that this file has been neglected a bit and it's missing lots of operations (e.g. there is no `dynamic_sub`), so it's probably a good idea to open an issue for it.

spy/vm/modules/operator/opimpl_f64.py
antocuni: +1 on this. See e.g. how we implement `w_64_div` in this file.

The idea is that we should get EXACTLY(*) the same behavior in the interpreter and in the compiler. The best way to ensure this property is to write tests, which are then automatically run in all possible configurations and they detect missing checks like that.

(with the exclusion of unsafe/unchecked operations, which are checked in interpreter and compiled-debug mode, but unchecked/undefined behavior in compiled-release mode).

spy/vm/modules/operator/opimpl_int.py
antocuni: @emanuelegurini I agree with @Viriathus1 and to reiterate on what I wrote in the past comment: the correct procedure is *first* to write the test, then to fix all the places where it fails.
Generally speaking, never add a line of code without having a failing test: this is the only way in which we can ensure to have good test coverage.

spy/libspy/include/spy/operator.h
antocuni: good catch!
I think that this missing features will become obvious as soon as we have the tests :)

spy/vm/modules/operator/opimpl_dynamic.py
antocuni: created #291

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **3 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Missing C implementation for integer power operator</strong>
   - [Location](https://github.com/spylang/spy/blob/69819870a0014175047ef5f818fbe4881ac7a6cc/spy/vm/modules/operator/binop.py#L75): `spy/vm/modules/operator/binop.py:L75`
   - Impact: Code using integer exponentiation might encounter compilation errors due to undefined symbols in the C backend.
   - Suggestion: Consider adding the corresponding `spy_operator$i32_pow` definitions to the C header files to ensure full build support.

2. <strong>Potential behavior divergence in float exponentiation</strong>
   - [Location](https://github.com/spylang/spy/blob/69819870a0014175047ef5f818fbe4881ac7a6cc/spy/vm/modules/operator/opimpl_f64.py#L48): `spy/vm/modules/operator/opimpl_f64.py:L48`
   - Impact: Negative bases with fractional exponents might behave differently between interpreted and compiled modes, potentially leading to inconsistent results.
   - Suggestion: Adding a validation check in `w_f64_pow` to match the C implementation's error handling would make the behavior more predictable.

3. <strong>Type safety check for integer exponentiation in the VM</strong>
   - [Location](https://github.com/spylang/spy/blob/69819870a0014175047ef5f818fbe4881ac7a6cc/spy/vm/modules/operator/opimpl_int.py#L46): `spy/vm/modules/operator/opimpl_int.py:L46`
   - Impact: Negative exponents currently return float values, which may cause runtime type mismatches when the VM expects an integer.
   - Suggestion: It would be safer to explicitly handle negative exponents to ensure the operation consistently adheres to integer type constraints.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #283, opened 2025-11-07
Use `@AstNode` instead of `@dataclass(eq=false)`

JeffersGlass:
As mentioned in [this code comment](https://github.com/spylang/spy/blob/6f36dbfe2f4951e4faf61b74a908ebd3eaeeec3d/spy/ast.py#L62-L74), it would be nice to decorate the ast node classes with an `@astnode` decorator, instead of manually using `@dataclass(eq=False)` for each one to make mypy understand what's going on.

This PR implements the `@AstNode` decorator as described, using [`typing.dataclass_transform`](https://typing.python.org/en/latest/spec/dataclasses.html#dataclass-transform) to let mypy know what's going on.

I also bumped the `ruff` compatibility from 3.11 to 3.12, to allow using generic syntax for functions. Hopefully this is ok, since the project itself targets >= 3.12. No worries if not, it can be re-worked using `TypeVar`.

### DIFF
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -69,14 +69,14 @@ del AST
 
 
 @dataclass_transform(field_specifiers=(dataclasses.field,), eq_default=False)
-def AstNode[T](klass: Type[T]) -> Type[T]:
+def astnode[T](klass: Type[T]) -> Type[T]:
     """Decorator to create dataclasses for AST nodes
     We want all nodes to compare by *identity* and be hashable, because e.g. we
     put them in dictionaries inside the typechecker."""
     return dataclass(eq=False)(klass)
 
 
-@AstNode
+@astnode
 class Node:
     loc: Loc = field(repr=False)
 
@@ -136,7 +136,7 @@ class Node:
                 node.visit(prefix, visitor, *args)
 
 
-@AstNode
+@astnode
 class Module(Node):
     filename: str
     docstring: Optional[str]
@@ -165,22 +165,22 @@ class Decl(Node):
     pass
 
 
-@AstNode
+@astnode
 class GlobalFuncDef(Decl):
     funcdef: "FuncDef"
 
 
-@AstNode
+@astnode
 class GlobalVarDef(Decl):
     vardef: "VarDef"
 
 
-@AstNode
+@astnode
 class GlobalClassDef(Decl):
     classdef: "ClassDef"
 
 
-@AstNode
+@astnode
 class Import(Decl):
     loc_asname: Loc
     ref: ImportRef
@@ -190,7 +190,7 @@ class Import(Decl):
 # ====== Expr hierarchy ======
 
 
-@AstNode
+@astnode
 class Expr(Node):
     """
     Operator precedence table, see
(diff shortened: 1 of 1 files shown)


spy/ast.py
```diff
@@ -59,22 +67,16 @@ def pp(self, *, hl=None) -> None:
...
-#         return dataclass (eq=False)
-#
-#     @astnode
-#     class Node:
-#         ...
-#
-# But we can't because this pattern is not understood by mypy.
+
+@dataclass_transform(field_specifiers=(dataclasses.field,))
+def AstNode[T](klass: Type[T]) -> Type[T]:
```
antocuni: I think I'd like it better to be called `@astnode` (using the same case as `@dataclass`), unless there is some convention which I'm unaware of which prescribes the use of CamelCase.

Apart from that, I swear I tried to read [the docs](https://typing.python.org/en/latest/spec/dataclasses.html) but I got more confused than before.

I have two questions here (they are real questions, not request to change anything):
  1. why do we need `field_specifiers`? The basic examples in the docs don't use it
  2. shouldn't we call `dataclass_transform(eq_default=False)` so that `mypy` knows about  the `eq=False` that we pass below?

Incidentally, this specific thing is a perfect example why spy approach in much better than current python typing: in spy, the typechecker will be able to just execute all decorators and see the final result, no need for this weird "meta language specs".

JeffersGlass: Hi Antonio! Hope you're well, I'll pop over into the Discord and say hi too :).

Happy to change it to `@astnode`, no biggie there.

----

So the `field_specifiers` seems to be necessary since without it, mypy thinks that the following line is setting a default value for `loc`, instead of just a field with a type:
https://github.com/spylang/spy/blob/0d99cc910c3b38d03b0422aab01da0004bc3e24f/spy/ast.py#L79
And therefore, without `field_specifiers`, it throws a bunch of _"Attributes without a default cannot follow attributes with one"_ errors

---

My reading of the docs is is that `eq_default` describes essentially the default value of the `eq` parameter of the _transformer_... and since `@astnode` doesn't take an `eq` parameter, type checkers don't need to know what the default is? Maybe? I'm admittedly also fuzzy here too.... regardless, mypy doesn't seem to do much in the presence of `eq` or `eq_default` except [ensure that the dataclass has a valid Self type for methods](https://github.com/python/mypy/blob/6986993532d84c3033dc1f28cda1c0c14ee1a55a/mypy/plugins/dataclasses.py#L294-L308). Happy to put it in or leave it out. Maybe I'll ping Jelle on the side and ask for some help.

antocuni: > So the field_specifiers seems to be necessary since without it, mypy thinks that the following line is setting a default value for loc, instead of just a field with a type:
[...]

ok, got it! Thanks for explaining :)

---

> My reading of the docs is is that eq_default describes essentially the default value of the eq parameter of the transformer... and since @astnode doesn't take an eq parameter, type checkers don't need to know what the default is? 

From what I understand, it's just a very elaborate/complicate way to say "this decorator will call @dataclass with eq=False".

I managed to find an example which seems to confirm my theory:
```python
from typing import dataclass_transform, Type
from dataclasses import dataclass
import dataclasses

@dataclass_transform()
def mydata[T](klass: Type[T]) -> Type[T]:
    return dataclass(frozen=True)(klass)

@mydata
class Point:
    x: int
    y: int

p = Point(1, 2)
p.x = 3
```

`p.x = 3` fails at runtime because `Point` is frozen, but mypy happily passes. However, if you do `@dataclass_transform(frozen_default=True)` then mypy correctly reports the error.

So, I think it's probably better to use use `eq_default=False` and inform mypy of what we are doing. Even if it doesn't do much now, better to be future proof in case they behavior changes later.

JeffersGlass: > So, I think it's probably better to use use eq_default=False and inform mypy of what we are doing. Even if it doesn't do much now, better to be future proof in case they behavior changes later.

That makes sense to me! I've added `eq_default=False` to that decorator.

PR #286, opened 2025-11-10
Increase ruff's target-version setting to match requires-python

Hasnep:
The project says it supports Python `>=3.12`, but Ruff is configured with the minimum target version of `3.11`, this fixes the discrepancy by updating Ruff's target version to 3.12.


antocuni:
@Hasnep thank you for your contribution. Unfortunately this PR is outdated now, since we already bumped `target-version` to `py312` when merging #283 . Sorry for that :).

Closing this without merging as it's no longer needed.

PR #289, opened 2025-11-10
Implement `list.__add__`

JeffersGlass:
Implement `list[T] + list[T]`. And some tests for the same.

My first stab at adding something to the vm-side here... please forgive if I've gotten some things wrong in implementation😅

### DIFF
--- a/spy/vm/list.py
+++ b/spy/vm/list.py
@@ -5,7 +5,7 @@ from spy.vm.b import BUILTINS, OP, B
 from spy.vm.builtin import builtin_method
 from spy.vm.object import W_Object, W_Type
 from spy.vm.opspec import W_MetaArg, W_OpSpec
-from spy.vm.primitive import W_I8, W_I32, W_U8, W_Bool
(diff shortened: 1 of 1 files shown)


antocuni (approved):
Thank you @JeffersGlass!

Sorry I didn't fully understand what you wanted to do here. The problem is that we have two different "list types":
  - the one in `vm/list.py` which works only in the interpreter
  - the one in `stdlib/_list.spy`: this is written in SPy and works in all modes

The "real" list type which SPy programs will use is `_list.spy`, although currently I miss all the syntactic sugar around it (in particular, you need to import it explicitly and there is no support for list literals).

`vm/list.py` is there mostly for bootstrapping reasons, because we need a list in the interpreter *before* we are able to compile `_list.spy`, but ideally I would like to get rid of it eventually (although I don't think it's going to happen any time soon). Because of that, I'm basically adding features to it "as needed".

Anyway, your PR is solid and the ability to `__add__` two lists might be useful, so I'm going to merge it anyway :).

PR #290, opened 2025-11-11
readme: fix path executable produced by spy

paugier:
It's silly, but it's nice when the examples in the README just work :slightly_smiling_face: 

PS: I'm really looking forward to seeing how will be Spy in few months!


antocuni:
@paugier it's not silly at all, I agree it's super important the the README "just works".
Thank you for fixing it!

PR #293, opened 2025-11-12
Remove unused encoding comments

Hasnep:
These comments are unused in modern Python, I also added a Ruff rule to disallow them in the future.

See [the Ruff documentation](https://docs.astral.sh/ruff/rules/utf8-encoding-declaration/) for more info.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ lint.select = [
     "I",  # isort
     "Q000", # double quotes
     "C",  # flake8-comprehensions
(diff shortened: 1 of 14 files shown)


antocuni:
> UTF8 has been a standard in Python for a long time, so I'm surprised to hear emacs doesn't support it by default.

oh, I think you are right. According to ChatGPT, emacs has been using utf-8 by default since 2009. I guess I just didn't know that 😂.

Thank you a lot for bringing this up, I learned something!

Merging the PR then 👍

PR #303, opened 2025-11-13
Clang format

### LINKED ISSUE #282: Format the C backend
Building on the Ruff format PR #256, we can also include a standardised formatter for the C backend.

`clang-format` looks like the best choice and there's pre-commit hook available [here](https://github.com/ssciwr/clang-format-wheel) but we need to settle on a specific style. The default style is `llvm` but others are [available](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options)

Also if we want to take it further with linters, we can discuss `clang-tidy`.

Viriathus1:
Closes #282 

- I've updated `AlwaysBreakAfterDefinitionReturnType` to `BreakAfterReturnType` because the former is deprecated. The value `AllDefinitions` functionally match the former config in my reading of it.
- `WhitespaceSensitiveMacros` didn't work as I hoped so I've added `src/jsffi/jsffi.c` to the exclude list in the pre-commit.

Let me know if you want to tweak some settings.

### DIFF
--- a/spy/libspy/include/spy/jsffi.h
+++ b/spy/libspy/include/spy/jsffi.h
@@ -1,10 +1,9 @@
 #ifndef SPY_JSFFI_H
 #define SPY_JSFFI_H
 
-#include <stddef.h>
 #include "emscripten.h"
 #include "spy.h"
-
+#include <stddef.h>
 
 typedef struct {
     int id;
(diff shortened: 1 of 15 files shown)


spy/libspy/include/spy/jsffi.h
```diff
@@ -16,55 +15,62 @@ void WASM_EXPORT(jsffi_init)(void);
 JsRef WASM_EXPORT(jsffi_string)(const char *ptr);
 JsRef WASM_EXPORT(jsffi_i32)(int32_t x);
 JsRef WASM_EXPORT(jsffi_wrap_func)(em_callback_func cfunc);
-JsRef WASM_EXPORT(jsffi_call_method_1)(JsRef c_target, const char *c_name, JsRef c_arg0);
+JsRef
+    WASM_EXPORT(jsffi_call_method_1)(JsRef c_target, const char *c_name, JsRef c_arg0);
```
antocuni, line 19: strange line. I think that in a case like that I'd prefer to put arguments on multiple lines, e.g. this:
```
JsRef WASM_EXPORT(jsffi_call_method_1)(
    JsRef c_target,
    ...
);
```

not sure if there is an easy way to achieve that

spy/libspy/src/libc.c
```diff
@@ -1,6 +1,7 @@
 #include "spy.h"
 
-int memcmp(const void *s1, const void *s2, size_t n) {
+int
+memcmp(const void *s1, const void *s2, size_t n) {
```
antocuni, line 4: oh, this file is a leftover from a distant past in which I was trying to compile for barebone wasm. We should delete it

ryanking13:
> I checked what pyodide does and they seem to just use // clang-format off around specific lines inside EM_JS macros. @hoodmane, @ryanking13 do you have suggestions?

We've been using `// clang-format off` everywhere for the emscripten macros. I never expected clang-format to work with the emscripten macros, to be honest.

> Have you tried to use WhitespaceSensitiveMacros as emscripten suggests?

No, but I think it is worth trying. No one in Pyodide (Me, hood, and even Roman) has ever touched clang-format config since it is added back in 2018.

spy/libspy/include/spy/jsffi.h
Viriathus1: So I added `WASM_EXPORT` to `WhitespaceSensitiveMacros` and it comes out as you want
```c
JsRef WASM_EXPORT(jsffi_call_method_1)(
    JsRef c_target,
    const char *c_name,
    JsRef c_arg0
);
```
🚀

Issue #297, opened 2025-11-12
[EPIC] Improve the playground


antocuni:
The current code of the [playground](https://spylang.github.io/spy/) is a bit more than a hack which I wrote in a hurry.

I think there is a lot of room for improvement, so I'm opening this issue as a starting point for anyone who wants to contribute to it. If anyone has an idea for an improvement and/or wants to work on a particular aspect, feel free to open a sub-issue.

Here is a random list of ideas to improve it in various ways:
1. UI improvements: the current look is not particularly good. It would be nice e.g. to have the editor/terminal automatically resized to fit the window and/or being able to decide on vertical vs horizontal split and/or being able to resize them. They playground is using LTK but it's not a requirement, I'd be happy to use a different framework (or no framework at all) if someone has opinions.

2. #311

3. #302 
4. #308 
5. #307 
6. #312 
7. Improve the examples selection. Would be nice to have more examples, possibly ordered from least to most complex, and show the various spy capabilities
8. anything else you can think of, happy to discuss :)

PR #306, opened 2025-11-14
Playground build fixes

### LINKED ISSUE #296: stdlib is not available in the playground
If you open [the playground](https://spylang.github.io/spy/) and try to import any module of the stdlib (e.g. `_list` or `_range`), it gives you an `ImportError`.

I *think* that this happens because `spy/stdlib/*.spy` files are NOT included in the spy wheel which we install inside pyodide.

Tagging @aisipos as he is the unofficial "playground maintainer" :)

### LINKED ISSUE #300: Playground Does Not Build without `build` pacakge
In playing with the playground a bit and trying to build it locally using `make`, I ran into a `ModuleNotFound: build` error. I think `build` needs to be added to the optional-dependencies.

I would just open a small PR, but this seems odd to me - wanted to make sure it wasn't a byproduct of using `uv` or something similar.

aisipos:
This PR fixes a few recently discovered issues with the playground.

- Fixes #300. I added `build` as a dependency. Antonio reported it didn't work for him as only a dev dependency. I don't see any harm adding it as a regular dependency, since SPy is a developer tool in itself.
- Fixes #296. I included `stdlib/` in both `MANIFEST.in` (for source distributions) and `pyproject.toml` for wheels. The solution I used was to add a `stdlib/__init.py` to make stdlib a python package, which allows setuptools to easily package it. I realize this is not ideal, since `stdlib/` is not actually a Python package, so I added a `raise` to warn anyone who tries to import it from Python. I think this solution might be ok for now, while we use CPython to bootstrap SPy we're dependent on its build tools to develop SPy, so we might have to conform to its rules for the time being. Still, happy to hear other ideas for a solution if this isn't suitable for us.
- Removed the autodetection and use of `uv` in `playground/Makefile`. Now it will use your $PYTHON env var or just `python`. I added a warning comment about `emcc` breaking if you set $PYTHON to an executable with arguments.
- I fixed a regression from commit b0f2b55a which broke the playground links to the renamed example file `myarray.spy`.
- To try to prevent playground example file name mismatches, I DRY'ed the list of example files in `playground/main.py` by reading the manifest in `pyscript.toml`

There's more to look at as part of the epic #297, so I'll look more at other improvements other PRs in the near future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,14 @@ build-backend = "setuptools.build_meta"
 include-package-data = true
 
 [tool.setuptools.packages.find]
-include = ["spy*"]
+include = ["spy*", "stdlib"]
+
+# setuptools doesn't package .spy files by default,
+# consider .spy files as "package-data" to include them. See:
+# https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data
+# https://setuptools.pypa.io/en/latest/userguide/datafiles.html#interplay-between-these-keywords
+[tool.setuptools.package-data]
+stdlib = ["*.spy"]
 
 [tool.pytest.ini_options]
 addopts = "--dist-dir=pyodide/node_modules/pyodide"


playground/main.py
```diff
@@ -62,7 +63,14 @@ def text(self, text=None):
...
 
-EXAMPLE_FILES = ["hello.spy", "bluefunc.spy", "point.spy", "array.spy"]
+# Auto-discover example files from pyscript.toml
+with open("pyscript.toml", "rb") as f:
+    config = tomllib.load(f)
+EXAMPLE_FILES = [
+    f"examples/{source}"
+    for source, dest in config.get("files", {}).items()
+    if dest == "examples/"
+]
```
antocuni, line 73: this works, but there might be an even easier way to do it.
This file runs inside pyodide in the browser and it sees the virtual emscripten filesystem: so it should be enough to just use `glob('examples/*.spy')`

antocuni (approved):
Thank you, looks very good to me.
I left a minor comment about example discovery but it's not a merge blocker.

> I included stdlib/ in both MANIFEST.in (for source distributions) and pyproject.toml for wheels. The solution I used was to add a stdlib/__init.py to make stdlib a python package, which allows setuptools to easily package it.

ah yes, the usual joy of python packaging :)
If we can't find any other way I'm fine with this solution, but I think there *should* be a better way? IIRC there are ways to include "data files" inside a python distribution (e.g. text files, images, etc.), so maybe we can reuse this mechanism for `stdlib`? After all, from the point of view of Python these are only normal text files.

antocuni:
@aisipos thanks for the PR.
I left two minor comments but don't feel required to fix them. Feel free to merge the PR as is if you want (I think you should have merge permissions, don't you?)

playground/main.py
aisipos: I initially did this with a glob, but then it didn't preserve the ordering of the files the way we have it now. The ordering became alphabetic, so the first ordered example became bluefunc.spy. I wanted hello.spy to be the first in the order, which then becomes the one that shows on first page load. So I ended up parsing the toml since they are listed there in "presentation order".

aisipos: Also as noted here: https://github.com/spylang/spy/issues/297#issuecomment-3531713498 , we'll probably be changing the presentation of the example files soon, so I might be overthinking the solution here.

antocuni:
thank you very much for the investigation! Yes, `include-package-data` is exactly what I was thinking of, although I couldn't remember the precise spelling.

No idea why setuptools includes an extra `__init__.py`, but I suppose we can live with that

Issue #246, opened 2025-10-03
Parameter with name `default` causes compilation failure


dpdani:
A function parameter with name `default` causes compilation issues when compiling to an executable.

Seen in CI: https://github.com/spylang/spy/actions/runs/18228372330/job/51905384526?pr=172#step:8:617

Small reproducer:

```python
def test(default: i32) -> None:
    print("Got: " + str(default))

def main() -> None:
    test(1)
```

```text
$ spy ./defaultparam.spy 
Got: 1
```

```text
$ spy -c ./defaultparam.spy
Traceback (most recent call last):
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/bin/spy", line 8, in <module>
    sys.exit(pyproject_entry_point())
             ~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/dp/repos/spy/spy/cli.py", line 248, in pyproject_entry_point
    return app()
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/typer/main.py", line 340, in __call__
    raise e
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/typer/main.py", line 323, in __call__
    return get_command(self)(*args, **kwargs)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/typer/core.py", line 680, in main
    return _main(
        self,
    ...<6 lines>...
        **extra,
    )
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/typer/core.py", line 198, in _main
    rv = self.invoke(ctx)
  File "/Users/dp/.pyenv/versions/3.13.2/envs/spy/lib/python3.13/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #315, opened 2025-11-15
Issue Fixing: Prevent C Reserve Naming

kanin-kearpimy:
**Issue:** [Parameter with name default causes compilation failure #246](https://github.com/spylang/spy/issues/246)
**Brief Problem:** When compiling `.spy` to c file by `--compile`, there are conflicts of C reserve naming such as correct `int default` in spy can be broken in C compilation due to its  [reserve naming](https://www.geeksforgeeks.org/c/keywords-in-c/)
**Solution:** I created a simple function `` in `serializer.py` to check variables. Then append `$` to the variable.
**Discussion:**
- I added checker into function params and local variable declaration. I'm not sure if I miss some coverage.
- I only do unit test `checker` in `serializer.py`.I tried to test end-to-end from `.spy` to `.c` to check if variable behave correctly. However, I don't yet manage to do so. (Is that necessary in this case?)

### DIFF
--- a/spy/tests/test_backend_c.py
+++ b/spy/tests/test_backend_c.py
@@ -77,18 +77,6 @@ class TestExpr:
         assert cstr(b"\x00\x01\x02") == r'"\x00\x01\x02"'
 
 
-def test_check_c_preserve():
-    for naming in C_Ident.C_KEYWORDS:
-        c_naming = C_Ident(naming)
-        assert f"{c_naming}" == f"{naming}$"
-
-
-def test_check_non_c_preserve():
-    namings = (
-        "dEfault",  # C is case-sensitive, so only `default` is reserved
-        "myName",
-        "__name_s",
-    )
-    for naming in namings:
-        c_naming = C_Ident(naming)
-        assert f"{c_naming}" == f"{naming}"
+def test_C_Ident():
+    assert str(C_Ident("hello")) == "hello"
+    assert str(C_Ident("default")) == "default$"


kanin-kearpimy:
Hi @antocuni ,

Thank for suggestions. I did move checker to `c_ast.py` and create `Ident` class accordingly.
However, I noticed there are circular dependency in `context.py` and `c_ast.py`, They refer to `C_Type`.
So, I decided to move such a cause to `share.py`. In my opinion, it's good idea to gradually create `share` venue for type and stuff before our project grow bigger. What do you think?

spy/backend/c/context.py
```diff
@@ -112,7 +98,7 @@ def c_function(self, name: str, w_func: W_ASTFunc) -> C_Function:
         for i, param in enumerate(w_functype.params):
             c_type = self.w2c(param.w_T)
             if param.kind == "simple":
-                c_param_name = check_c_preserve(funcdef.args[i].name)
+                c_param_name = str(C.Ident(funcdef.args[i].name))
```
antocuni: this calls to `str()` looks bad.
I think we can:
  1. move C.Ident to `context.py` and call it `C_Ident` (similarly to `C_Type`)
  2. modify `C_FuncParam.name` to be of type `C_Ident`

Generally speaking, you should never need an explicit call to `str(C_Ident(...))`, because `C_Ident.__str__` already does the job.

spy/backend/c/cwriter.py
```diff
@@ -85,7 +84,7 @@ def emit_local_vars(self) -> None:
                 varname not in ("@return", "@if", "@while", "@assert")
                 and varname not in param_names
             ):
-                self.tbc.wl(f"{c_type} {check_c_preserve(varname)};")
+                self.tbc.wl(f"{c_type} {str(C.Ident(varname))};")
```
antocuni:
```suggestion
                c_varname = C.Ident(varname)
                self.tbc.wl(f"{c_type} {c_varname};")
```

I think it looks much easier to read this way. Generally speaking I like to avoid having too much logic inside f-strings.
Note also the the call to str() is unnecessary.

spy/backend/c/cwriter.py
```diff
@@ -162,12 +161,12 @@ def emit_stmt_Assign(self, assign: ast.Assign) -> None:
     def emit_stmt_AssignLocal(self, assign: ast.AssignLocal) -> None:
         target = assign.target.value
         v = self.fmt_expr(assign.value)
-        self.tbc.wl(f"{check_c_preserve(target)} = {v};")
+        self.tbc.wl(f"{str(C.Ident(target))} = {v};")
```
antocuni: same as above

spy/backend/c/cwriter.py
```diff
@@ -162,12 +161,12 @@ def emit_stmt_Assign(self, assign: ast.Assign) -> None:
...
 
     def emit_stmt_AssignCell(self, assign: ast.AssignCell) -> None:
         v = self.fmt_expr(assign.value)
         target = assign.target_fqn.c_name
-        self.tbc.wl(f"{check_c_preserve(target)} = {v};")
+        self.tbc.wl(f"{str(C.Ident(target))} = {v};")
```
antocuni: same as above

spy/backend/c/cwriter.py
```diff
@@ -277,7 +276,7 @@ def fmt_expr_Name(self, name: ast.Name) -> C.Expr:
         assert False, "ast.Name nodes should not survive redshifting"
 
     def fmt_expr_NameLocal(self, name: ast.NameLocal) -> C.Expr:
-        return C.Literal(check_c_preserve(name.sym.name))
+        return C.Literal(str(C.Ident(name.sym.name)))
```
antocuni: again, no need to call str()

spy/tests/compiler/test_basic.py
```diff
@@ -1285,3 +1285,13 @@ def foo() -> i32:
         """
         mod = self.compile(src)
         assert mod.foo() == 5 * 3  # 5 outer iterations, 3 inner iterations each
+
+    def test_c_reserve_local_variable(self):
```
antocuni:
```suggestion
    def test_use_C_keywords_as_identifiers(self):
```

spy/backend/c/c_ast.py
```diff
@@ -319,3 +319,54 @@ def precedence(self) -> int:
 
     def __str__(self) -> str:
         return f"({self.type}){self.expr}"
+
+
+@dataclass
+class Ident(Expr):
+    # C_PRESERVE_NAMING is initlized once with immutable property.
+    C_PRESERVE_NAMING = (
```
antocuni:
```suggestion
    C_KEYWORDS = (
```

I don't understand why you call it `C_PRESERVE_NAMING`: it seems actually the opposite, this is the list of names which are **not** preserved :).

Anyway, I think that `C_KEYWORDS` eliminate any ambiguity

spy/tests/test_c_ident.py
antocuni, line 1: +1 for having unit tests for this specific functionality :clap: .

However, there is no need to create a new file just for that: I think you can just write a new test `test_C_Ident` in `test_backend_c.py` (which will be the union of your `test_check_c_preserve` and `test_check_non_c_preserve`)

kanin-kearpimy:
Hi @antocuni ,

I resolved all comments. Thank you so much. I enjoy learning SPy project and compiler stuff. So, please let me know if it will be beneficial for project.

Please let me know what you think.

Issue #313, opened 2025-11-15
Implement the walrus operator


antocuni:
**Warning**: this is labeled as `help wanted` but not `easy first issue`. It's not an easy task

We want to support this:
```python
def main() -> None:
    y = (x := 1)
    print(y) # 1
    print(x) # 1
```

This means:
1. add a new node `ast.AssignExpr` (subclass of `ast.Expr`)

2. add the relevant code in `parser.py` (and write a test in `test_parser.py`)

3. write the relevant logic in `analyze/scope.py`. For the purpose of ScopeAnalyzer, `ast.AssignExpr` should be treated similarly to `ast.Assign`, and introduce a new name only if the same name cannot be found elsewhere. We also need to make sure that it plays well with names defined in outer scopes, e.g.:
```
x = 1
var y = 1

def main() -> None:
    z0 = (x := 1) # error
    z1 = (y := 1) # works, and should update the global 'y'
```

Write all needed tests in `test_scope.py`. **Hint**: during development, it might be useful to play with `spy --symtable` to see which symtables are generated.

4. Write a test in `test_basic.py` and make sure it works for `[interp]` tests (hint: learn how to run a single test with pytest). The correct way to implement support in `ASTFrame` is to use a "specialization" mechanism similar to what we do for `specialized_names` and `specialized_assigns`

5. run the test in `[doppler]` mode and fix it (if needed)

6. run the test in `[C]` mode and fix it

Viriathus1:
@antocuni closing as #316 and #320 implement this operator

PR #316, opened 2025-11-15
Implement the walrus operator

Jemeljanov:
Relates to #313 

- Add `ast.AssignExpr` and parse  `:=`  into it
- Treat `AssignExpr` like `Assign`
- Runtime specialization
- Backend

### DIFF
--- a/spy/analyze/scope.py
+++ b/spy/analyze/scope.py
@@ -285,6 +285,8 @@ class ScopeAnalyzer:
             vardef.loc,
             vardef.type.loc,
         )
+        if vardef.value is not None:
+            self.declare(vardef.value)
 
     def declare_FuncDef(self, funcdef: ast.FuncDef) -> None:
         # declare the func in the "outer" scope
--- a/spy/tests/test_scope.py
+++ b/spy/tests/test_scope.py
@@ -251,6 +251,16 @@ class TestScopeAnalyzer:
             "@return": MatchSymbol("@return", "var", "auto"),
         }
 
+    def test_vardef_initializer_declares_assignexpr_target(self):
+        scopes = self.analyze("""
+        def foo() -> i32:
+            var z: i32 = (x := 1)
+            return x + z
+        """)
+        scope = scopes.by_funcdef(self.mod.get_funcdef("foo"))
+        assert scope._symbols["z"] == MatchSymbol("z", "var", "explicit")
+        assert scope._symbols["x"] == MatchSymbol("x", "const", "auto")
+
     def test_const_var_without_type(self):
         scopes = self.analyze("""
         def foo() -> None:


Jemeljanov:
Added scope fix for nested assignment expressions

spy/doppler.py
```diff
@@ -475,3 +475,22 @@ def shift_expr_CallMethod(self, op: ast.CallMethod) -> ast.Expr:
...
+                target=assignexpr.target,
+                value=new_value,
+            )
+        elif isinstance(specialized, ast.AssignCell):
+            return ast.AssignExprCell(
+                loc=assignexpr.loc,
+                target=assignexpr.target,
+                target_fqn=specialized.target_fqn,
+                value=new_value,
+            )
```
antocuni: I think you can just use `specialize.replace(value=newvalue)` instead of having to go through all the cases manually, can't you?

spy/tests/compiler/test_basic.py
```diff
@@ -125,6 +125,67 @@ def foo() -> i32:
...
+        assert mod.foo() == 4  # second call: x=2, calls=2
+        assert mod.calls == 2
+
+    def test_assignexpr_preserves_binding(self):
+        mod = self.compile("""
+        def foo() -> i32:
+            result = (x := 1)
+            return x + result
+        """)
+        assert mod.foo() == 2
```
antocuni, line 187: I'm not sure to understand what is the difference between this and `test_assignexpr_basic`?

spy/tests/compiler/test_basic.py
```diff
@@ -125,6 +125,67 @@ def foo() -> i32:
...
+        mod = self.compile("""
+        var counter: i32 = 0
+
+        def inc() -> i32:
+            return (counter := counter + 1)
+
+        def foo() -> i32:
+            return inc() + inc()
+        """)
+        assert mod.foo() == 3  # 1 + 2
```
antocuni: same question as above, this seems to be basically a copy of `test_assignexpr_updates_cell`?

spy/tests/test_scope.py
```diff
@@ -174,6 +174,40 @@ def foo(a: i32) -> None:
...
+
+        def bar() -> None:
+            while True:
+                if (y := 1):
+                    break
+        """)
+        foo_scope = scopes.by_funcdef(self.mod.get_funcdef("foo"))
+        assert foo_scope._symbols["x"] == MatchSymbol("x", "const", "auto")
+        bar_scope = scopes.by_funcdef(self.mod.get_funcdef("bar"))
+        assert bar_scope._symbols["y"] == MatchSymbol("y", "var", "auto")
```
antocuni: I think it's better to follow the same style as the other tests and test the content of the *whole* `scope._symbols` dict

spy/tests/test_scope.py
```diff
@@ -174,6 +174,40 @@ def foo(a: i32) -> None:
...
+            z0 = (x := 1) # scope captures const x; runtime should reject assigning to it
+            z1 = (y := 1)
+        """)
+        scope = scopes.by_funcdef(self.mod.get_funcdef("main"))
+        assert scope._symbols["x"] == MatchSymbol("x", "const", "global-const", level=1)
+        assert scope._symbols["y"] == MatchSymbol(
+            "y", "var", "explicit", level=1, storage="cell"
+        )
+        assert scope._symbols["z0"].varkind == "const"
+        assert scope._symbols["z1"].varkind == "const"
```
antocuni: same comment as above. Better to check the whole `scope._symbols` instead of cherry-picking just a few symbols and/or properties.

spy/tests/test_scope.py
antocuni: also, you should add a `test_var_and_const_assignexpr` and check that a single assignexpr behaves the same way as a single assign in terms of var/const

spy/vm/astframe.py
```diff
@@ -834,6 +849,28 @@ def eval_expr_NameOuterCell(self, name: ast.NameOuterCell) -> W_MetaArg:
...
+            specialized = self._specialize_AssignExpr(assignexpr)
+            self.specialized_assignexprs[assignexpr] = specialized
+        if isinstance(specialized, ast.AssignLocal):
+            return self._execute_AssignLocal(assignexpr.target, assignexpr.value)
+        elif isinstance(specialized, ast.AssignCell):
+            return self._execute_AssignCell(
+                assignexpr.target, specialized.target_fqn, assignexpr.value
+            )
+        else:
+            assert False
```
antocuni, line 892: I'm a bit confused here: in `ast.py` you introduce `ast.AssignExprLocal` and `ast.AssignExprCell`, but then you never actually instantiate them?
I expected that `_specialized_AssignExpr` would return one of those, but in reality it returns an `ast.Assign`, which is a stmt instead of an expr.

Can you explain what's going on and that is the reasoning behind?

Jemeljanov: Ok, I will refactor `_specialize_AssignExpr` to return the expr variants (AssignExprLocal/Cell) directly, I was reusing the stmt specialization and only rebuilding those in Doppler.

antocuni:
@Jemeljanov the PR seems very good and ready to be merged, but it's still marked as draft.
Is it ready or you are still planning to do more work on it?

Jemeljanov:
@antocuni I modified tests and added one, can be final change, please let me know.

antocuni:
@Jemeljanov there seem to be a problem with doppler.
Look at this example:
```python
def inc(x: int) -> int:
    return x + 1

def main() -> None:
    x = 0
    y = inc(x := 1)
    print(x)  # should be 1
    print(y)  # should be 2
```

interp works correctly:
```
❯ spy /tmp/x.spy 
1
2
```

doppler doesn't (`-rx` or `--redshift --execute` executes the redshifted AST), similarly to what happens in `[doppler]` tests:
```
❯ spy -rx /tmp/x.spy 
0
2
```

it seems that the `AssignExpr` is redshited into a `Constant`:
```
❯ spy -r /tmp/x.spy 

def inc(x: i32) -> i32:
    return x + 1

def main() -> None:
    x: i32
    x = 0
    y: i32
    y = `x::inc`(1)
    print_i32(x)
    print_i32(y)
```

You can look at the redshifted AST with `--redshift --parse`:
```
❯ spy -rp /tmp/x.spy
[...]
    body=[
        AssignLocal(
            target=StrConst(value='x'),
            value=Constant(value=0),
        ),
        AssignLocal(
            target=StrConst(value='y'),
            value=Call(
                func=FQNConst(fqn=FQN('x::inc')),
                args=[
                    Constant(value=1),     # where does this Constant come from?!?
                ],
            ),
        ),
[...]
```

spy/tests/compiler/test_basic.py
Jemeljanov: `test_assignexpr_basic` is in if statement.

spy/tests/compiler/test_basic.py
Jemeljanov: `test_assignexpr_updates_cell` checks a walrus in bump() updates a global counter that another function current() reads later. 
`test_assignexpr_updates_nonlocal_cell` checks calling the walrus function twice in one expression increments the shared counter each time.

Please let me know if you would like to remove test.

spy/tests/compiler/test_basic.py
antocuni: ah I see. Then I'd call the test "test_assignexpr_inside_if" :)

spy/tests/compiler/test_basic.py
```diff
@@ -174,6 +196,17 @@ def foo() -> i32:
...
+        mod = self.compile("""
+        @blue
+        def make_value() -> i32:
+            return (x := 1)
+
+        def foo() -> i32:
+            return make_value()
+        """)
+        assert mod.foo() == 1
+
```
antocuni: I'm just curious here, did you need to write the test to fix an actual bug, or you wrote it "just because"?

spy/doppler.py
```diff
@@ -342,7 +342,8 @@ def eval_expr(
                   -> compute shited binop (stored in .shifted_expr)
         """
         wam = super().eval_expr(expr, varname=varname)
-        if wam.color == "blue":
+        const_foldable = wam.color == "blue" and not self._expr_has_side_effect(expr)
+        if const_foldable:
```
antocuni: this works, but it seems "wrong" to detect it in the doppler.
I thin that this check should happen in the ASTFrame, and the expression should be marked as `red` if it has side effects.

Also note that not *all* assignexpr must be red: only the ones to variables.
For example:
```
def foo(x: int) -> None:
    pass

def main() -> None:
    x = 0
    foo(x := 1)  # x is var, don't constant fold
    foo(y := 2)  # y is const, it's safe to constant fold
    print(x)
    print(y)
```

this should print `1 2`, but also the `y := 2` should be redshifted away and be left with only `2`.

spy/tests/test_doppler.py
```diff
@@ -133,6 +133,131 @@ def foo() -> i32:
...
+            return x + 1
+
+        def main() -> None:
+            x: i32
+            x = 0
+            y: i32
+            y = `test::inc`(x := 1)
+            print_i32(x)
+            print_i32(y)
+        """)
```
antocuni: why do you need this test? What does it test in addition to the one above?

spy/tests/test_doppler.py
```diff
@@ -133,6 +133,131 @@ def foo() -> i32:
...
+            return (x := 1)
+
+        def foo() -> i32:
+            return make_value()
+        """)
+        self.assert_dump("""
+        def foo() -> i32:
+            return 1
+        """)
+
```
antocuni, line 182: I don't think you need tests for *all* these cases.
I don't see any code in `doppler.py` which does different things for all these cases, so they are probably not needed.
The code which *does* depend on "if/while/etc" seems to be in ScopeAnalyzer, but I think we already check all these variants in test_scope.py

spy/vm/astframe.py
```diff
@@ -452,41 +453,65 @@ def _specialize_Assign(self, assign: ast.Assign) -> ast.Stmt:
...
+    def _execute_AssignLocal(
+        self, target: ast.StrConst, value: ast.Expr, expr: bool = False
+    ) -> W_MetaArg:
         varname = target.value
         lv = self.locals.get(varname)
         if lv is None:
             # first assignment, implicit declaration
             wam = self.eval_expr(value)
-            self.declare_local(varname, wam.color, wam.w_static_T, target.loc)
+            desired_color = "red" if expr else wam.color
```
antocuni: this seem to be related to what I wrote above. Not all assignexpr are red, only the ones to `var`.

spy/doppler.py
Jemeljanov: ok, seems fixed.

spy/tests/compiler/test_basic.py
Jemeljanov: I wrote it just in case but probably doesn't add much value as it doesn't check the assigned value. I will check also other test that could be redundant.

Jemeljanov:
@antocuni thanks again for the reviews, I pushed the latest changes. 

Please let me know if you would like to make changes.

spy/tests/compiler/test_basic.py
antocuni: yes, generally speaking in SPy we follow Test Driven Development: first you write a failing test, THEN you write the code to fix it.
This way, we can make sure that each piece of code is linked to a test which actually excercises it

spy/tests/test_scope.py
```diff
@@ -251,6 +251,16 @@ def foo() -> None:
...
+    def test_vardef_initializer_declares_assignexpr_target(self):
+        scopes = self.analyze("""
+        def foo() -> i32:
+            var z: i32 = (x := 1)
+            return x + z
+        """)
+        scope = scopes.by_funcdef(self.mod.get_funcdef("foo"))
+        assert scope._symbols["z"] == MatchSymbol("z", "var", "explicit")
+        assert scope._symbols["x"] == MatchSymbol("x", "const", "auto")
+
```
antocuni, line 263: please use the same "check all the symbols" approach that we use everywhere else

spy/analyze/scope.py
```diff
@@ -285,6 +285,8 @@ def declare_VarDef(self, vardef: ast.VarDef) -> None:
             vardef.loc,
             vardef.type.loc,
         )
+        if vardef.value is not None:
+            self.declare(vardef.value)
```
antocuni, line 289: when I read this line I though "this should have a test in test_scope"... and then I saw it's there! Good job :muscle:

spy/vm/astframe.py
```diff
@@ -881,34 +878,38 @@ def eval_expr_AssignExpr(self, assignexpr: ast.AssignExpr) -> W_MetaArg:
...
         )
 
-    def _mark_assignexpr_red(self, wam: W_MetaArg) -> W_MetaArg:
-        """Assignment expressions mutate state; force them red to block folding."""
-        return wam.as_red(self.vm)
+    def _mark_assignexpr_red(self, target: ast.StrConst, wam: W_MetaArg) -> W_MetaArg:
+        sym = self.symtable.lookup(target.value)
+        if sym.varkind == "var":
+            return wam.as_red(self.vm)
+        return wam
```
antocuni, line 912: this code is correct and I'm going to merge the branch anyway, but I'd like to add a couple of stylistic comments:
1. `_mark_assignexpr_red` implies that you are always make it red. I think more appropriate names are e.g. `_mark_assignexpr_red_maybe` or even better `set_assignexpr_color`.
2. For a specific case of very small function like this, it's debatable whether it's better to factor out the common logic, or just replicate  the logic in the caller.

Again, it's not wrong and I might have made the same choice, but it's worth keeping it in mind that "avoid repetition at all costs" is not always the right thing to do. I think a lot depends on how complex is the logic.

antocuni (approved):
VERY GOOD 🎉.
I'm merging it, congratulations for your first SPy contribution :clap: 

Feel free to open a follow up PR to fix the very minor comments which I left on this one.

PR #318, opened 2025-11-16
Add a loading indicator before pyodide initialization.

aisipos:
This is a quick PR for some easy playground wins. See: #297 

# Add a loading indicator before pyodide runs

It can take a couple of seconds before you see the initial terminal that tells you that SPy is loading. To signal to the user that the page is working, I added a small animated loading screen:
image

# Add some console.log statements to indicate when pyscript events happen

To give more insight into what the page is doing, I added some `console.log` statements (in both .js and .py) to signal when particular lifetime events are happening. It might be a little useful in debugging if things break. 
image


antocuni (approved):
love it! :disguised_face:

PR #322, opened 2025-11-30
Implement `and` and `or`

### LINKED ISSUE #248: Implement `and` and `or`

Jemeljanov:
Closes #248

### DIFF
--- a/spy/tests/test_parser.py
+++ b/spy/tests/test_parser.py
@@ -644,29 +644,13 @@ class TestParser:
     def test_BoolOp(self, op, node):
         mod = self.parse(f"""
         def foo() -> bool:
-            return a {op} b
+            return a {op} b {op} c
         """)
         stmt = mod.get_funcdef("foo").body[0]
         expected = f"""
         Return(
             value={node}(
-                left=Name(id='a'),
-                right=Name(id='b'),
-            ),
-        )
-        """
-        self.assert_dump(stmt, expected)
-
-    def test_BoolOp_chained(self):
-        mod = self.parse("""
-        def foo() -> bool:
-            return a or b or c
-        """)
-        stmt = mod.get_funcdef("foo").body[0]
-        expected = """
-        Return(
-            value=Or(
-                left=Or(
+                left={node}(
                     left=Name(id='a'),
                     right=Name(id='b'),
                 ),


spy/tests/compiler/test_basic.py
```diff
@@ -354,6 +354,67 @@ def bar() -> None:
...
+        mod = self.compile("""
+        var x: i32 = 0
+
+        def make_true() -> bool:
+            x = x + 1
+            return True
+
+        def make_false() -> bool:
+            x = x + 1
+            return False
```
antocuni: I think it's a good idea to be consistent with the names `counter` and `bump_{true,false}` which are used in the test below

spy/tests/compiler/test_basic.py
```diff
@@ -354,6 +354,67 @@ def bar() -> None:
         mod.bar()
         assert mod.x == 3
 
+    def test_boolops_short_circuit(self):
```
antocuni: these test should be placed AFTER `test_bool_logical_ops`.
The idea is that first we check the basic behavior in `test_bool_logical_ops`, then we check "advanced" behavior.
Also, consistent naming. I propose:
   - test_bool_ops_simple
   - test_bool_ops_short_circuit

The reason for not using just `test_bool_ops` is that it doesn't play well with `pytest -k`, because `pytest -k test_bool_ops` would select ALSO `test_bool_ops_short_circuit`

spy/tests/compiler/test_basic.py
```diff
@@ -354,6 +354,67 @@ def bar() -> None:
...
+            return make_true() or make_false()
+        """)
+
+        assert mod.check_and() is False
+        assert mod.x == 1
+
+        assert mod.check_or() is True
+        assert mod.x == 2
+
+    def test_boolops_chained_short_circuit(self):
```
antocuni: I don't understand what are we testing here that it's not already tested in `test_boolops_short_circuit`.

I know it's not written anywhere (my fault) but in SPy we follow TDD: you should write a failing test, THEN the code to fix it.

spy/vm/astframe.py
```diff
@@ -947,6 +947,64 @@ def eval_expr_UnaryOp(self, unop: ast.UnaryOp) -> W_MetaArg:
         w_opimpl = self.vm.call_OP(unop.loc, w_OP, [wam_v])
         return self.eval_opimpl(unop, w_opimpl, [wam_v])
 
+    def _eval_bool(self, wam: W_MetaArg) -> W_MetaArg:
```
antocuni: this function seems unnecessary.
If you call `self.eval_expr(op.left, varname="@if")`, you ALREADY get a value which has been converted to bool because of `varname="@if"`.
So, wam.w_static_T is already `w_bool`, and `w_typeconv` is always None.
The code after `w_typeconv is None` is NEVER executed (try to put an `assert False` and run the tests)

spy/vm/astframe.py
```diff
@@ -947,6 +947,64 @@ def eval_expr_UnaryOp(self, unop: ast.UnaryOp) -> W_MetaArg:
...
+            self.vm,
+            wam.color,
+            w_typeconv.w_functype.w_restype,
+            w_val,
+            wam.loc,
+            sym=wam.sym,
+        )
+
+    def eval_expr_And(self, op: ast.And) -> W_MetaArg:
+        wam_l = self._eval_bool(self.eval_expr(op.left, varname="@if"))
```
antocuni: I think I'd prefer to have `varname="@and"` and `varname="@or"`.

spy/vm/astframe.py
```diff
@@ -947,6 +947,64 @@ def eval_expr_UnaryOp(self, unop: ast.UnaryOp) -> W_MetaArg:
...
+        w_bool_l = wam_l._w_val if isinstance(wam_l._w_val, W_Bool) else None
+
+        if w_bool_l is not None and self.vm.is_False(w_bool_l):
+            return wam_l
```
antocuni, line 993: You should never access `wam._w_val` directly. If you find yourself in this situation, it's a red flag that something is wrong.

In this precise case, what we want is:
1. if we are in normal execution, we want to check the truth value of w_bool_l and evaluate w_bool_r only if necessary
2. if we are redshifting, we want to evaluate BOTH sides always.

So probably we want logic like this:
```python
    def eval_expr_And(self, op: ast.And) -> W_MetaArg:
        if self.redshifting:
            wam_l = self.eval_expr(op.left, varname="@if")
            wam_r = self.eval_expr(op.right, varname="@if")
            color = maybe_blue(wam_l.color, wam_r.color)
            if color == "blue":
                w_val = wam_l.w_val and wam_r.w_val
            else:
                w_val = None
            return W_MetaArg(self.vm, color, B.w_bool, w_val, op.loc)

        else:
            # in this case we know for sure that we can access wam.w_val
            wam_l = self.eval_expr(op.left, varname="@if")
            wam_r = self.eval_expr(op.right, varname="@if")
            if self.vm.is_False(wam_l.w_val):
                ...
            ...
```

We need this split logic because `And` and `Or` are special: they are expressions BUT they also influence the control flow: until now, all the control flow nodes where statements (like `if` and `while`) and statements are handled specially by doppler.

spy/tests/compiler/test_basic.py
Jemeljanov: Actually this test was failing earlier because I had only implemented support for a single comparison first, so expressions like `1 and 1 and 1` were not handled at that stage. I can remove the test above.

spy/tests/compiler/test_basic.py
```diff
@@ -659,6 +659,96 @@ def foo() -> None:
...
+        assert mod.or_bool(True, False) is True
+        assert mod.or_bool(False, True) is True
+        assert mod.or_bool(False, False) is False
+
+        assert mod.and_ints(1, 0) is False
+        assert mod.and_ints(0, 0) is False
+        assert mod.or_ints(0, 1) is True
+        assert mod.or_ints(0, 0) is False
+
+    def test_bool_ops_func_const(self):
```
Jemeljanov: This test was failing during modification in astframe. Is it possible to keep it or should be removed?

spy/vm/astframe.py
Jemeljanov: Ok, thank you. In the final implementation if not self.redshifting, the check `if self.vm.is_False(wam_l.w_val):` will be before `wam_r = self.eval_expr(op.right, varname="@if")` so that short circuit behaviour is preserved, right?

spy/vm/astframe.py
Jemeljanov: It was entering the `if self.redshifting` block but this is because during reshifting the conversion is deferred, right?

spy/tests/compiler/test_struct.py
```diff
@@ -194,6 +194,36 @@ def foo() -> Point:
         mod = self.compile(src)
         assert mod.foo() == (0, 0)
 
+    def test_class_body_bool_ops_declaration(self):
```
Jemeljanov, line 197: This test was recommended by AI, not sure if should be supported but previously only `self.declare_local("@if", "red", B.w_bool, Loc.fake())` was declared inside run in `ClassFrame`

antocuni: good catch! We can keep it 👍

spy/vm/astframe.py
```diff
@@ -953,6 +957,60 @@ def eval_expr_UnaryOp(self, unop: ast.UnaryOp) -> W_MetaArg:
...
+            if color == "blue":
+                w_left = wam_l.w_val
+                w_right = wam_r.w_val
+                assert isinstance(w_left, W_Bool)
+                assert isinstance(w_right, W_Bool)
+                w_val = (
+                    B.w_True
+                    if self.vm.is_True(w_left) and self.vm.is_True(w_right)
+                    else B.w_False
+                )
```
antocuni:
```suggestion
                w_val = vm.wrap(
                    vm.unwrap_bool(w_left) and vm.unwrap_bool(w_right),
                )
```

this should have the very same semantics as your code, but it's much shorter and easier to read.

spy/vm/astframe.py
antocuni: yes exactly!

spy/vm/astframe.py
```diff
@@ -953,6 +957,60 @@ def eval_expr_UnaryOp(self, unop: ast.UnaryOp) -> W_MetaArg:
...
+    def eval_expr_And(self, op: ast.And) -> W_MetaArg:
+        if self.redshifting:
+            wam_l = self.eval_expr(op.left, varname="@and")
+            wam_r = self.eval_expr(op.right, varname="@and")
+            color = maybe_blue(wam_l.color, wam_r.color)
+            if color == "blue":
+                w_left = wam_l.w_val
+                w_right = wam_r.w_val
+                assert isinstance(w_left, W_Bool)
+                assert isinstance(w_right, W_Bool)
```
antocuni: ouch, we have another problem which I didn't think about.
In your PR, the following code works in interp, but crashes in doppler:
```
x = 1 and 2
```

I *think* that the problem is here: when we are redshifting, we don't apply the type conversion here, so w_left/w_right are not `W_Bool`:
https://github.com/Jemeljanov/spy/blob/b4b409f5db3b43c9c3e797b6fea9d81a8ac70efd/spy/vm/astframe.py#L215-L221

antocuni (changes requested):
very good, but I found another problem, see comments

spy/tests/compiler/test_basic.py
antocuni: yes but chained comparisons are handled directly in the parser, aren't they?
From the point of view of ASTFrame, there is 0 difference between `a and b and c` and `(a and b) and c`, isn't it?

spy/vm/astframe.py
antocuni: oh, maybe this is what solved the problem `1 and 2` which I explain below. If so, probably it was correct to have it. Sorry for making you doing duplicated work

spy/tests/compiler/test_basic.py
antocuni: if it's a test which check some actual behavior yes, we should keep it. What was the failure?

spy/tests/compiler/test_basic.py
Jemeljanov: yes, ok, then the test was not needed

spy/vm/astframe.py
Jemeljanov: seems fixed now

spy/tests/compiler/test_basic.py
Jemeljanov: I triggered this once while making changes, but did not reproduce it in other situations. I’m wondering whether it’s still worth keeping as a guard, or if it can be safely removed.

antocuni: the question is: is there any line or part of code which would make THIS test to fail, when all the others are still passing?
If the answer is yes, then the test should be kept. If the answer is not, the test is redundant.

spy/tests/compiler/test_basic.py
```diff
@@ -659,6 +669,119 @@ def foo() -> None:
...
+        assert mod.or_false_true() is True
+        assert mod.or_true_false() is True
+
+    def test_bool_ops_assignment(self):
+        mod = self.compile("""
+        def foo() -> bool:
+            x = 1 and 2
+            return x
+        """)
+        assert mod.foo() is True
```
antocuni, line 710: why do you need this test?

spy/tests/compiler/test_basic.py
Jemeljanov: Ok, thank you for the clarification.

At this point, making only this specific test fail would require a fairly intentional change to the code. I will remove it.

spy/tests/compiler/test_struct.py
antocuni: on a second thought, I think we can make this test simpler, shorter and merge it with `test_bool_ops_module_level`.

Something like this:
```
def test_bool_ops_in_modframe_and_classframe:
    src = """
    x = True and False
    y = True or False
    
    @struct
    class Foo:
        if True and False:
            aaa: i32
        if True or False:
            bbb: i32
```

and the in the test you check the values of `x`, `y` and the fields of `Foo`.
There is no need to test all possible combinations (they are tested elsewhere), here we want to check that `and` and `or` works in a modframe and in a classframe

spy/tests/compiler/test_basic.py
Jemeljanov: This is the reason of adding `_ensure_bool`, without it, this test would crash during doppler and c test.

PR #324, opened 2025-12-02
Added u32

Viriathus1:


spy/libspy/include/spy/operator.h
```diff
@@ -15,11 +15,13 @@
 #define i8 int8_t
 #define u8 uint8_t
 #define i32 int32_t
+#define u32 uint32_t
 #define f64 double
 
 DEFINE_CONV(i32, bool)
 DEFINE_CONV(i32, i8)
 DEFINE_CONV(i32, u8)
+DEFINE_CONV(i32, u32)
```
antocuni, line 24: I'm not sure this is correct, because there are `i32` values which cannot be represented as `u32` and vice-versa.

I'll add my thoughts about it the review comment

antocuni (commented):
The PR looks good.

I'm not sure about the implicit i32<->u32 conversion because you can easily get overflows.

I'm going to merge the PR as is, but I created issue #333 to remember about it.

PR #326, opened 2025-12-02
Support --format=json output with colorization

aisipos:
Add a JSON format option to `--colorize`, like so:

```
spy --colorize example.spy
```

Which will output JSON like so:
```
[
  {
    "line": 21,
    "col": 4,
    "length": 22,
    "type": "red"
  },
  {
    "line": 21,
    "col": 8,
    "length": 18,
    "type": "red"
  },
...
```

The intent of this PR is to make colorize output readable by other tools so they can also add colorization support. For an early example of this, see: https://github.com/aisipos/spy-mode.el
image

Eventually I think this sort of functionality belongs in an LSP server. But since we don't have one yet, adding JSON support to SPy cli outputs seemed like a quick win.

### DIFF
--- a/spy/cli.py
+++ b/spy/cli.py
@@ -21,7 +21,12 @@ from spy.doppler import ErrorMode
 from spy.errors import SPyError
 from spy.magic_py_parse import magic_py_parse
 from spy.textbuilder import Color
-from spy.util import cleanup_spyc_files, colors_coordinates, highlight_src_maybe
+from spy.util import (
+    cleanup_spyc_files,
+    colors_coordinates,
+    format_colors_as_json,
+    highlight_src_maybe,
+)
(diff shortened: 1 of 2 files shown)


antocuni (approved):
LGTM, I'm going to merge it.

However, cli.py is getting VERY complicated and hairy, we should probably rethink how the various options interact with each other.

E.g., now you can use `--colorize --parse` to see the colorized AST, but at this point it's probably better to use `--colorize --format=ast`?

JeffersGlass:
> However, cli.py is getting VERY complicated and hairy, we should probably rethink how the various options interact with each other.

@antocuni Since a bunch of the flags are mutually exclusive (or *almost* mutually exclusive), how would you feel if they were subcommands instead of flags? I'd be happy to take a swing at that if it's of interest.

PR #331, opened 2025-12-04
Handle conditional struct fields in class bodies

### LINKED ISSUE #248: Implement `and` and `or`

Jemeljanov:
Follow-up changes for #322 
Fixes #248
- Merge `test_bool_ops_module_level` and `test_class_body_bool_ops_declaration`
- Refactor class handling: `ClassDef` now does not carry `fields`
- Add support for
```
@struct
class Point:
    if True and False:
        a: i32
    if True or False:
        b: i32
```


spy/tests/compiler/test_struct.py
```diff
@@ -194,35 +194,37 @@ def foo() -> Point:
...
-
-        def read_or() -> i32:
-            return Flags(0).or_result()
-        """
-        mod = self.compile(src)
-        assert mod.read_and() == 2
-        assert mod.read_or() == 3
+    def test_bool_ops_in_modframe_and_classframe(self):
+        mod = self.compile(
+            """
```
antocuni: this has a weird indentation, especially compared to all the other tests. I suspect it depends on how your autoformatter decides it should look.

I suggest to use a different style, which should be autoformatter-friendly:
```
src = """
...
"""
mod = self.compile(srC)
```

spy/tests/compiler/test_struct.py
```diff
@@ -271,6 +273,20 @@ def dir_inst() -> list[str]:
...
+        class Foo:
+            pass
+        """
+        mod = self.compile(src)
+        w_Foo = mod.w_mod.getattr("Foo")
+        reserved = {"@if", "@and", "@or", "@while", "@assert"}
+
+        field_names = {w_field.name for w_field in w_Foo.iterfields_w()}
+        assert reserved.isdisjoint(field_names)
+        assert reserved.isdisjoint(w_Foo.dict_w.keys())
```
antocuni: this seems a rather complicated way to say `field_names == set()`.

spy/vm/classframe.py
```diff
@@ -29,24 +29,26 @@ def __init__(
...
-            # ignore variables which were just declared but never assigned; this
-            # includes .e.g field declarations
-            if lv.w_val is not None:
+            # ignore reserved bool locals
+            if name.startswith("@"):
+                continue
+            if lv.w_val is None:
+                # locals declared but not assigned
+                body.fields_w[name] = W_Field(name, lv.w_T)
+            else:
```
antocuni: this code seems to do the job, but if we support having `VarDef` inside `classdef.body`, then `classdef.fields` becomes redundant/unnecessary.

We could modify `ast.ClassDef` into this:
```
@astnode
class ClassDef(Stmt):
    body_loc: Loc
    name: str
    kind: ClassKind
    docstring: Optional[str]
    body: list["Stmt"]
    symtable: Any = field(repr=False, default=None)
```

but then it requires serious thinking to handle all possible cases.
See my comment in the PR review box.

PR #345, opened 2025-12-30
Implementing `slice()` and slices

JeffersGlass:
Implements slice objects by (1) writing a Slice class in spy and (2) turning Slice ast nodes into calls to the Slice constructor at frame eval time.

### DIFF
--- a/spy/doppler.py
+++ b/spy/doppler.py
@@ -494,6 +494,7 @@ class DopplerFrame(ASTFrame):
         v_stop = self.shifted_expr[op.stop]
         v_step = self.shifted_expr[op.step]
         return self.shift_opimpl(op, w_opimpl, [v_start, v_stop, v_step])
+
     def shift_expr_Dict(self, dict: ast.Dict, wam: W_MetaArg) -> ast.Expr:
         if len(dict.items) == 0:
             assert wam.w_static_T is SPY.w_EmptyDictType


docs/src/contributing.md
antocuni, line 1: this seems to be an artifact of a spurious/wrong merge? Probably it should disappear if you rebase and/or merge the current `main`

spy/analyze/symtable.py
```diff
@@ -197,6 +197,8 @@ def add_sym(attr: str, loc: Optional[Loc], impref: ImportRef) -> None:
 
         add_sym("range", None, ImportRef("_range", "range"))
         add_sym("list", None, ImportRef("_list", "list"))
+        add_sym("Slice", None, ImportRef("_slice", "Slice"))
+        add_sym("slice", None, ImportRef("_slice", "Slice"))
```
antocuni, line 200: why the capital `Slice`? I think it should just be lowercase `slice` as on CPython

spy/backend/c/context.py
```diff
@@ -100,7 +100,9 @@ def decl(self) -> str:
         if self.params == []:
             s_params = "void"
         else:
-            paramlist = [f"{p.c_type} {p.name}" for p in self.params]
+            paramlist = [
+                f"{p.c_type} {p.name}" for p in self.params if p.c_type.name != "void"
+            ]
```
antocuni, line 106: interesting. I suppose this is needed because "somewhere" you get a function with a `void` parameter?
Which function is that?

spy/backend/spy.py
```diff
@@ -514,6 +514,12 @@ def fmt_expr_Tuple(self, node: ast.Tuple) -> str:
         items = ", ".join(itemlist)
         return f"({items})"
 
+    def fmt_expr_Slice(self, node: ast.Slice) -> str:
+        (start, stop, step) = [
+            self.fmt_expr(exp) for exp in (node.start, node.stop, node.step)
+        ]
+        return f"Slice({start}:{stop}{f':{step}' if step else ''})"
+
```
antocuni, line 530: this should have a corresponding test in `test_backend_spy`.

spy/tests/stdlib/test__list.py
```diff
@@ -91,6 +91,57 @@ def out_of_bounds() -> int:
...
             mod.out_of_bounds()
 
+    def test_getitem_slice_type(self):
+        mod = self.compile("""
+            def foo() -> list[i32]:
+                l = [1,2,3]
+                return l[0:2:1]
+            """)
+        x = mod.foo()
+        assert type(x) == list
```
antocuni: this test can probably be merged with the next one (you can check both the type and the content at the same time)

antocuni: also, any specific reason why you use the full syntax `0:2:1` instead of just `0:2`?

spy/tests/stdlib/test__list.py
```diff
@@ -91,6 +91,57 @@ def out_of_bounds() -> int:
...
+        assert mod.long_stride() == [1, 3, 5]
+
+    def test_getitem_slice_zero_step(self):
+        mod = self.compile("""
+            def zero_slice() -> list[i32]:
+                l = [1,2,3]
+                return l[1:2:0]
+            """)
+        with pytest.raises(SPyError):
+            mod.zero_slice()
```
antocuni, line 127: you can use `with SPyError.raises("W_xxx")` for more precise type.
See e.g.:

https://github.com/spylang/spy/blob/9599ec7c3f6cbeea5c7bec421243b55b599ce333/spy/tests/stdlib/test__list.py#L550-L551

spy/tests/stdlib/test__slice.py
```diff
@@ -0,0 +1,73 @@
...
+
+        def make_slice_none() -> Slice:
+            return slice(None, None, None)
+        """
+        mod = self.compile(src)
+        s = mod.make_slice()
+        assert type(s) == UnwrappedStruct
+        assert "_slice::Slice" in str(s.fqn)
+        assert (s.start, s.stop, s.step) == (0, 1, 2)
+        assert not any([s.start_is_none, s.stop_is_none, s.step_is_none])
```
antocuni:
```suggestion
        assert (s.start_is_none, s.stop_is_none, s.step_is_none) == (0, 0, 0)
```

looks much easier to read to me

spy/tests/stdlib/test__slice.py
```diff
@@ -0,0 +1,73 @@
...
+
+        sn = mod.make_slice_none()
+        assert type(s) == UnwrappedStruct
+        assert "_slice::Slice" in str(s.fqn)
+        assert all([sn.start_is_none, sn.stop_is_none, sn.step_is_none])
```
antocuni: same as above

spy/tests/stdlib/test__slice.py
```diff
@@ -0,0 +1,73 @@
...
+    def test__slice_indices(self):
+        def get_slice_indices(
+            start=None, stop=None, step=None, *, length
+        ) -> tuple[int, int, int]:
+            src = f"""
+                def _get_indices() -> str:
+                    s: Slice = Slice({start}, {stop}, {step})
+                    indices = s.indices({length})
+                    return str(indices.start) + ", " + str(indices.stop) + ", " + str(indices.step)
+            """
```
antocuni: this seems unnecessarily complicated. Our test machinery supports returning structs, so you could just return `_slice.tuple3` and compare them to app-level tuples; something like this (untested):
```python
src = """
    def get_indices(...) -> tuple3:
        ...
        return s.indices(length)
"""
mod = self.compile(src)
assert mod.get_indices(...) == (1, 2, 3)
```

antocuni: also, would be nice to compile the module a single time, it would be easier and faster.
I think that the problem is that you don't have a way to declare an argument as `int | None`, isn't it?

You *might* be able to work around the issue by marking the test as `@only_interp` and type the arguments as `dynamic`, but I don't really know if it works (it's worth trying)

spy/tests/conftest.py
```diff
@@ -1,14 +1,26 @@
...
+# know why, but if we create the ENGINE so early, then calls to spy_panic crashes with
+# "Illegal instruction" instead of correctly raising WasmTrap.
+#
+# I have no idea why it happens.
+if "spy.llwasm" in sys.modules:
+    raise Exception("""
+    You shouldn't import spy.llwasm so early in the conftest, else spy_panic crashes
+    during tests. See PR #378.
+    """)
+
```
antocuni: also spurious merge

stdlib/_list.spy
```diff
@@ -71,16 +73,56 @@ def list(T):
...
+                        i = i + ll.length
+                    if i < 0:
+                        raise IndexError
+                    if i >= ll.length:
+                        raise IndexError
+                    return ll.items[i]
+                return OpSpec(getitem_int)
+            
+            elif m_x.static_type == Slice:
+                def _py_list_slice_wrap(_self: _ListImpl, s: Slice) -> _ListImpl:
```
antocuni: why this name instead of e.g. `getitem_slice`?

stdlib/_list.spy
```diff
@@ -71,16 +73,56 @@ def list(T):
...
+                    ll = _self.__ll__
+                    unpacked = _py_slice_unpack(ll.length, s)
+                    new_length = _py_adjust_indexes(ll.length, unpacked.start, unpacked.stop, unpacked.step)
+                    indices = s.indices(new_length)
+
+                    if new_length <= 0: return []
```
antocuni, line 104: nitpick, but this formatting (and others) is a bit strange.
Probably it would be good to enable ruff autoformatting also for .spy files (not necessarily in this PR)

stdlib/_list.spy
```diff
@@ -252,4 +294,72 @@ def list(T):
             self.append(item)
             return self
 
+
     return _ListImpl
+
+def _py_adjust_indexes(length: int, start: int, stop: int, step: int) -> int:
```
antocuni: I seem to understand that this function translates "abstract" indices (e.g. negative or greater than the length) into "concrete" ones, is this right?
I think it's worth writing a docstring, because it took a while to me to decipher it

stdlib/_list.spy
```diff
@@ -252,4 +294,72 @@ def list(T):
...
+    else:
+        if start < stop:
+            val = int((stop - start - 1) / step) + 1
+        
+    return val
+        
+def _py_slice_unpack(length:int, s:Slice) -> tuple3:
+    _start: int = 666
+    _stop: int = 666
+    _step: int = 666
```
antocuni: why 666? I guess it's just a placeholder but then:
1. is it really necessary? Can't we just declare them as `_start: int` without an initializer?
2. if it is necessary, just add a comment saying "placeholder"

stdlib/_list.spy
```diff
@@ -252,4 +294,72 @@ def list(T):
...
+        _step = 1
+    else:
+        _step = s.step
+
+    if _step == 0:
+        raise ValueError("slice step cannot be zero")
+    
+    if s.start_is_none:
+        if _step < 0:
+            _start = 10000000
```
antocuni: I think we should use the equivalent of `sys.maxint` here.
We don't have `sys` yet, but you can probably put a constant `maxi32 = 2147483647`at module level.

stdlib/_list.spy
antocuni: also, the rest of the list/range functions use `i32` everywhere. Better to use it also for slices, for consistency.
(Eventually we need to decide what is the right type to use for indices/lengths, but I think that for now `i32` is fine)

stdlib/_slice.spy
```diff
@@ -0,0 +1,116 @@
...
+from operator import OpSpec, MetaArg
+from types import NoneType
+
+from __spy__ import interp_list
+
+@struct
+class tuple3:
+    start: int
+    stop: int
+    step: int
```
antocuni: same comment as above about `int` vs `i32`

stdlib/_slice.spy
```diff
@@ -0,0 +1,116 @@
...
+class Slice:
+    start: int
+    start_is_none: int # bool
+    stop: int
+    stop_is_none: int # bool
+    step: int
+    step_is_none: int # bool
+
+    @blue.metafunc
+    def __new__(m_cls, n_start, n_stop, n_step) -> OpSpec:
```
antocuni: the naming convention for metaargs is `m_*` (where `m` stands for "meta")

stdlib/_slice.spy
```diff
@@ -0,0 +1,116 @@
...
+    step: int
+    step_is_none: int # bool
+
+    @blue.metafunc
+    def __new__(m_cls, n_start, n_stop, n_step) -> OpSpec:    
+        
+        if (n_start.static_type, n_stop.static_type, n_step.static_type) == (int, int, int):
+            def impl1(_start:int, _stop:int, _step:int) -> Slice:
+                return Slice.__make__(_start, 0, _stop, 0, _step, 0)
+            return OpSpec(impl1, interp_list[MetaArg](n_start, n_stop, n_step))
```
antocuni: this totally needs a comment to explain what's going on, it took 10 minutes to me to understand it 😅.

The point is that we cannot use the normal `[]` list literal syntax because `_list.spy` import `_slice.spy`, and thus we would get a circular dependency which causes a very strange error.

I also suggest a more compact way of writing this function; something along these lines:
```python
    def __new__(m_cls, m_start, m_stop, m_step) -> OpSpec:
        TYPES = (m_start.static_type, m_stop.static_type, m_step.static_type)
        args_m = interp_list[MetaArg](m_start, m_stop, m_step)  # XXX explain

        if TYPES == (int, int, int):

            def impl1(_start: int, _stop: int, _step: int) -> Slice:
                return Slice.__make__(_start, 0, _stop, 0, _step, 0)

            return OpSpec(impl1, args_m)

        elif TYPES == (NoneType, int, int):

            def impl2(_start: NoneType, _stop: int, _step: int) -> Slice:
                return Slice.__make__(0, 1, _stop, 0, _step, 0)

            return OpSpec(impl2, args_m)

        ...
```

antocuni (commented):
very good job @JeffersGlass 👏 
I left some comments here and there; they might look a lot but it's just because this PR is pretty big, but overall it looks good.

Another note: this PR implements slice support only for `__getitem__`, correcT? It might be worth raising a `WIP` in the `__setitem__` case (and leave the actual implementation for another PR)

spy/analyze/symtable.py
JeffersGlass: I was under the impressoin that in CPython the function was `slice` and class was `Slice`... but it turns out I was wrong =)

spy/backend/c/context.py
JeffersGlass: This code suppresses any `None` parameters of functions - in particular opimpls returned by `slice.__new__`.

Consider the following spy code:
```py
def main() -> None:
    l = [1,2,3][:1]
```
The metafunc that is `slice.__new__` ends up selecting `impl6` to us. For typechecking reasons, impl6 takes parameters of the form `(None, int, None)`. But in the C code, we don't need (or want) to pass a void parameteter - we know we only need a single `int` to specify, in this case, the end of the slice. Otherwise, we end up with code like:

```c
spy__slice$Slice spy__slice$Slice$__new__$impl6(void _start, int32_t _stop, void _step) {  //<<< errror
    return (spy__slice$Slice){ 0, 1, _stop, 0, 1, 1 };
}
```

spy/tests/stdlib/test__list.py
JeffersGlass: My thinking on this test was to have a test where none of the arguments are `None`, since that is, in a sense, the simplest case and future changes around optional args could potentially change what needs to be tested. But either way, it can be combined with the following test.

spy/tests/stdlib/test__slice.py
JeffersGlass: Thank you! Agreed.

spy/tests/stdlib/test__list.py
JeffersGlass: Thank you!

spy/tests/stdlib/test__slice.py
JeffersGlass: Thank you! Agreed.

spy/tests/stdlib/test__slice.py
JeffersGlass: > this seems unnecessarily complicated.

Agreed, I like your change and it avoids the unnecessary conversion to a string

> You might be able to work around the issue by marking the test as @only_interp and type the arguments as dynamic, but I don't really know if it works (it's worth trying)

So doing a quick rewrite to something like this, I get an TypeError, stemming from a null opspec:

```py
from _slice import tuple3
def _get_indices(start: dynamic, stop: dynamic, step: dynamic, length: int) -> tuple3:
    s: slice = slice(start, stop, step)
    return s.indices(length)

def main() -> None:
    s = _get_indices(1,2,3,4)
```
```
TypeError: cannot call objects of type `types::StructType`
  | /workspaces/spy/foo.spy:3
  |     s: slice = slice(start, stop, step)
  |                |___| this is `types::StructType`

  | <builtins>:0
  | 
  | ^ `slice` defined here

  | /workspaces/spy/foo.spy:3
  |     s: slice = slice(start, stop, step)
  |                |______________________| operator::CALL called here
```

Another option would be to rewrite it so it generates all the source of the module with all the necessary functions, compiles it once, then calls each function.  [I implemented this](https://github.com/JeffersGlass/spy/blob/b95d08577c867ba0f00d3a07dcc6600765981b44/spy/tests/stdlib/test__slice.py#L28-L85) and pushed in the latest set of commits. It's a little hairy, and I won't mind if we want to go back to the previous version.

What do you think? It does run faster.

stdlib/_list.spy
JeffersGlass: It's actually mirroring the name in CPython that the implementation is inspired by, same as `__py_slice_unpack` and `_py_adjust_indexes`. But I'll change it so it makes more sense for this context =)

stdlib/_list.spy
JeffersGlass: Opened #387 on this =)

stdlib/_list.spy
JeffersGlass: It's not strictly necessary - apologies, a byproduct of tracking down an error and having a value that didn't appear anywhere else in code or tests was helpful. Removed.

stdlib/_slice.spy
JeffersGlass: Agreed, and a very good idea to simplify thing. I've added a couple clarifying comments and adjusted the notation as you suggested.

.python-version
JeffersGlass, line 1: This appears to have been committed in error in an earlier PR

stdlib/_list.spy
JeffersGlass: Also ran `ruff format` over just `_slice.spy` since it's new.

stdlib/_list.spy
JeffersGlass: I've added a docstring to explain that function, and switched to i32 everywhere.

spy/backend/c/context.py
antocuni: ah I see, basically you want to support this scenario:
```python
def foo(x: None) -> None:
    pass
```

I never thought about it but I suppose it makes sense to allow it. It should have a test in `test_basic.py` though (like `test_None_arguments`)

antocuni (approved):
thank you @JeffersGlass!
There is a conflict in doppler.py, but feel free to merge it once the conflict is resolved and you add `test_None_arguments` in test_basic.py

Issue #305, opened 2025-11-14
Add complex numbers


Viriathus1:
I noticed that in the [Python docs](https://docs.python.org/3/reference/expressions.html#the-power-operator) for the `**` operator that it should return a complex number when raising a negative number to a fractional power. In this PR #280 - if it gets fixed - we raise a `ValueError` which is correct according to earlier versions of Python but I believe we should have a proper complex number type.

Viriathus1:
@antocuni the python complex number maps to a c doubles so I believe that should be the default, no? Then we can have another narrower type in `complex64` which would map to c floats.

Issue #333, opened 2025-12-05
Better conversion rules between numeric types


antocuni:
PR #324 adds `u32`, with an implicit conversion to and from `i32`, but it's unsafe because the values might be out of range.

This is another instance of a problem which already exists:

https://github.com/spylang/spy/blob/62276d998766a76ed8665d2512c23106140861ca/spy/vm/modules/operator/convop.py#L174-L186

I think that *ideally*, we would like this situation:
  - implicit conversions are allowed only when they are safe
  - we should have explicit conversions for potentially unsafe casts: these would always be checked and panic if the value is not in range
  - we should also have explicit _unsafe_ conversions, which are checked only in DEBUG mode
  - numeric *literals* should be treated specially; e.g. if we have `x + 1`, `1` should be treated to be the same type as `x`. This probably involves to create a `literal_number` type which can be converted as needed.

Which precise syntax to use to do these conversion is up to discussion. A possible proposal is this:

```python
from operator import convert
from unsafe import unsafe_convert

def foo(x: i32) -> u32:
    return convert[u32](x)

def bar(x: i32) -> i8:
    return unsafe_convert[i8](x)
```

Internally, we should continue to maintain a table of allowed conversions, but we should distinguish between implicit and explicit.

Custom types should be able to participate in the conversion by implementing `__convert_from__` and `__convert_to__` as now, but these special methods should probably also get an `implicit: bool` argument to be able to distinguish between the two.

PR #347, opened 2026-01-02
Added single precision floating point `f32`

Viriathus1:
Added single precision float as type `f32` which was mentioned in #305.

Slightly different implementation from the other numerical primitives in that all operations call the C backend directly through WASM exported functions as discussed.

### DIFF
--- a/spy/vm/modules/unsafe/div.py
+++ b/spy/vm/modules/unsafe/div.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING, Annotated, Any, Protocol
 
 from spy.errors import SPyError
 from spy.vm.b import B
+from spy.vm.modules.operator.multimethod import MultiMethodTable
 from spy.vm.opspec import W_MetaArg
 from spy.vm.primitive import W_F32, W_F64, W_I8, W_I32, W_U8, W_U32
 from spy.vm.w import W_Object, W_OpSpec, W_Type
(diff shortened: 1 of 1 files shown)


spy/libspy/src/operator.c
```diff
@@ -0,0 +1,126 @@
+#include "spy.h"
+#include <math.h>
+#include <stdbool.h>
+
+float
+spy_operator$f32_add(float a, float b) {
+    return a + b;
+}
+
```
antocuni:
```suggestion
spy_operator$f32_add(float x, float y) {
    return x + y;
}

```

very nitpick, but for consistency with the rest

spy/vm/modules/unsafe/div.py
```diff
@@ -151,23 +157,66 @@ def w_f64_unchecked_mod(vm: "SPyVM", w_a: W_F64, w_b: W_F64) -> W_F64:
...
+    match w_T:
+        case B.w_f64:
+            return W_OpSpec(UNSAFE.w_f64_ieee754_div)
+        case B.w_f32:
+            return W_OpSpec(UNSAFE.w_f32_ieee754_div)
+        case _:
+            raise SPyError(
+                "W_TypeError",
+                f"Unsupported type `{w_T.fqn.human_name}` for unchecked division operation",
+            )
```
antocuni: this seems suboptimal, because e.g. you cannot call it `ieee754_div` on ints.
E.g. the following works on main but fails on this branch:
```python
from unsafe import ieee754_div

def main() -> None:
    x = ieee754_div(10, 3)
    print(x)
```

I suggest to use the same approach as normal div, using multimethods. You can even add the `ieee754_div` op to the "main" multimethod table in operator/binop.py, and then use it from here.

Coercion rules should be the same as normal div:
- `int / int -> f64`
- `int / f64 -> f64`
- `f64 / int -> f64`
- `f32 / f64 -> f64`
- `f64 / f32 -> f64` 
- `f32 / f32 -> f32`
- `f32 / int -> f32`
- `int / f32 -> f32`

antocuni (changes requested):
@Viriathus1 thank you a lot for this.
It seems mostly complete, but it lacks two things IMHO:
1. automatic coercion for `ieee754_div`, see my comment below
2. conversion rules for `f32`.

For example, the following two calls should work, but they fail:
```
def foo(x: f32) -> f32:
    return x

foo(1)
foo(1.2)
```

```
TypeError: mismatched types
  | /tmp/x.spy:7
  |     foo(1)
  |         ^ expected `f32`, got `i32`
```

```

TypeError: mismatched types
  | /tmp/x.spy:7
  |     foo(1.2)
  |         |_| expected `f32`, got `f64`
```


similarly, I should be able to convert from `f32` to `int` if I do it explicitly:
```
def foo(x: f32) -> int:
    return int(x)
```

```
TypeError: cannot call objects of type `type`
  | /tmp/x.spy:4
  |     return int(x)
  |            |_| this is `type`
```

Viriathus1:
@antocuni I've added the changes for this one. I was a bit lazy with the coercion rules mostly because it would be resolved in this issue #333. Let me know if I've missed something

spy/libspy/include/spy/operator.h
```diff
@@ -60,6 +63,17 @@ spy_operator$f64_to_i32(double x) {
...
+spy_operator$f32_to_i32(float x) {
+    if (isnan(x))
+        return 0;
+    if (x > INT32_MAX)
+        return INT32_MAX;
+    if (x < INT32_MIN)
+        return INT32_MIN;
+    return (int32_t)x;
+}
+
```
antocuni, line 76: good choice here! I think that saturating conversion is what happens in rust and C#, so +1 to do the same in SPY

PR #350, opened 2026-01-05
Don't Highlight when `colorize`-ing

### LINKED ISSUE #344: New Highlighter Can Overwrite Colorizer
Image

The new highlighter seems to be emitting color code in a way that can 'shadow' the colors from the colorizer. See the sample code below.

```py
from tempfile import NamedTemporaryFile

from spy.cli import colorize_sourcecode

data = [
    ("1:2:3", {1: [("0:5", "blue")]}),
    ("1,2,3", {1: [("0:5", "blue")]}),
    ("def foo()", {1: [("0:8", "blue")]}),
]

for src, coords in data:
    fp = NamedTemporaryFile("w+", delete=False)
    fp.write(src)
    fp.close()
    print(colorize_sourcecode(fp.name, coords))
    del fp
```

(To see that the code is being correctly colorized by the colorizer, add `print("".join(highlighted_lines))` before the return in `cli::colorize_sourcecode()`)

Not a huge issue of course, but might be fun for someone to tackle; stumbled across it while working on Slices and thought I was doing something wrong with the Loc.

JeffersGlass:
Because the built-in highlighter and colorizer conflict, don't highlight lines that have been colorized.

Fixes #344

### DIFF
--- a/spy/tests/test_cli.py
+++ b/spy/tests/test_cli.py
@@ -178,32 +178,36 @@ class TestMain:
     def test_colorize_source(self):
         # source formatting is the default - run all the examples below
         # with both 'colorize --format spy' and bare 'colorize'
-        argsets = [["colorize", "--format", "spy"], ["colorize"]]
-        for argset in argsets:
-            _, stdout = self.run(*argset, self.factorial_spy, decolorize_stdout=False)
-            # B stands for Blue, R for Red, [/COLOR] means that the ANSI has been reset
-            expected_outout = """\
-            def factorial(n: i32) -> i32:
-                [R]res = [/COLOR][B]1[/COLOR]
-                for i in [B]range[/COLOR][R](n)[/COLOR]:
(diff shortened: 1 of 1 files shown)


antocuni:
> I think we'll want to preserve `decolorize_stdout`, since otherwise a bunch of other tests which assume the output is plaintext without ANSI start breaking...

oh indeed you are right, I read the code too fast.
Yes, we need to keep `decolorize_stdout`, sorry for the noise.

> Agreed about removing the loop and `*argset` stuff, I'll get rid of it.

thank you!
Feel free to merge the PR autonomously once you do that, if you want (assuming you have the permissions)

JeffersGlass:
All set! Looks like I don't have permissions to actually merge, but should be good to go.

antocuni:
> All set! Looks like I don't have permissions to actually merge, but should be good to go.

ah, it seems that I sent you an invite some time ago but you never accepted and it's no longer valid. I sent a new one!

PR #358, opened 2026-01-08
Show Nice Error when Variable Shadowing Occurs in Inner Functions

### LINKED ISSUE #340: Missing Clean Error when Untyped Local Shadows Outer Variable
The following code produces an assertion error ( `assert sym.is_local`). It would be nice if it could show the friendly ` variable 'x' shadows a name declared in an outer scope` error.

```py
@struct
class Point:
    x: i32

    def __new__(i: i32) -> Point:
        x = i
        return Point.__make__(x)

def main() -> None:
    p = Point(1)
```

Adding a type to x inside `__new__` (i.e. `x: i32 = i`) produces the friendly error - it would be nice if this happened even if x is untyped.

JeffersGlass:
Turns an `assert` statement that symbol name should be local into a nice error message, when a variable in an inner func shadows an outer func. Currently interp_only, since funcdefs cannot be redshifted (yet).

#231 may be the root of this as well, or another cause - this PR is just about adding the nice error message for now.

Closes #340


aisipos:
@antocuni Sorry for the wait on this, I have updated #279, hopefully it's mergeable now.

PR #364, opened 2026-01-12
Add `redshift --format` option

### LINKED ISSUE #362: `spy redshift` should have a `--format` option
@JeffersGlass I found another regression with the CLI, sorry for not catching it during the review.

Currently `spy redshift` shows the AST by default; I found that you need to use `--human-readable` to get the spy code as it was before, but I admit I had to look at the tests to understand what was going on :).

I think that:
  1. we should support `spy redshift --format` (as we do for e.g. `spy colorize`)
  2. the default format should be `spy` (this is by far the most useful output if you want to understand what's left after redshifting)
  3. we should remove `--human-readable`

One reason for having a `--format` is that it then becomes easy to add _new_ ways of visualizing; for example, I have an experimental branch which can render AST graphically using graphviz, so I can imagine having `spy redshift --format=dot`.

Please let me know if you want to take this (else, I'll be happy to do it by myself)

JeffersGlass:
Changes the redshift flag "`--human-readable" into the more flexible `--format` option, to make it clearer what the option does and to make it easier to add further output formats.

Closes #362

### DIFF
--- a/spy/cli/commands/redshift.py
+++ b/spy/cli/commands/redshift.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 from typing import Annotated
 
+import click
 from typer import Option
 
 from spy.analyze.importing import ImportAnalyzer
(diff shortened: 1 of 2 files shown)


antocuni (approved):
awesome!

PR #366, opened 2026-01-17
Feature/dict/parser

kanin-kearpimy:


spy/parser.py
```diff
@@ -611,6 +611,16 @@ def from_py_expr_List(self, py_node: py_ast.List) -> spy.ast.List:
         items = [self.from_py_expr(py_item) for py_item in py_node.elts]
         return spy.ast.List(py_node.loc, items)
 
+    def from_py_expr_Dict(self, py_node: py_ast.Dict) -> spy.ast.Dict:
+        keyValuePairItems = []
+        for key, value in zip(py_node.keys, py_node.values):
+            keyValuePairItems.append(
+                spy.ast.KeyValuePair(
+                    py_node.loc, self.from_py_expr(key), self.from_py_expr(value)
```
antocuni: `py_node.loc` seems wrong here, because it points to the whole dict. It's probably a good idea to use `key.loc`.

To see the difference, imagine to have this:
```
d = {
    a: 1,
    b: 2,
}
```

`py_node.loc` points to the WHOLE expression from `{` to `}`, while `key.loc` points just to `a`.

antocuni (approved):
LGTM so far, but make sure to write proper tests

spy/tests/test_parser.py
```diff
@@ -39,10 +39,12 @@ def assert_dump(self, node: ast.Node, expected: str):
...
 
     def test_Module(self):
-        mod = self.parse("""
+        mod = self.parse(
+            """
         def foo() -> None:
             pass
-        """)
+        """
+        )
```
antocuni, line 47: something went wrong with your autoformatter, it created a lot of unnecessary changes which clutter the PR, like this parenthesis which was moved to the next line.
Can you please fix your autoformatter?

The only official formatter supported by spy is `ruff format`. Depending on your setup, you could either configure your editor to run it, or disable editor autoformat and run `ruff format` manually.

spy/tests/compiler/test_dict.py
```diff
@@ -0,0 +1,19 @@
+from spy.fqn import FQN
+from spy.tests.support import CompilerTest, expect_errors, no_C, only_interp
+from spy.vm.b import B
+from spy.vm.object import W_Type
+
+
+class TestDict(CompilerTest):
+    @no_C
+    def test_empty_dict_singleton(self):
```
antocuni, line 9: this test is failing because redshifting is not implemented yet.
I am fine to accept this PR even without support for redshifting, but then we should mark this test as `@skip_backends('doppler', 'C')`, to clearly mark that they are just temporary skips and to make the build green

spy/vm/astframe.py
```diff
@@ -1122,6 +1122,66 @@ def eval_expr_List(self, lst: ast.List) -> W_MetaArg:
 
         return wam_list
 
+    def eval_expr_Dict(self, dict: ast.Dict) -> W_MetaArg:
```
antocuni, line 1125: I haven't read this code yet, but it seems that all the tests are missing.

We can proceed two ways:
- make a smaller PR with just the parser (which is ready to be merged), and continue the work in astframe in a different PR
- write all the proper tests for this code :)

Issue #342, opened 2025-12-22
Add support for dict literals


antocuni:
PRs #337 and #339 added support for list literals, in the forms `[a, b, c, ...]` and `[]`, which are translated into calls to `_list.list` (implemented by `stdlib/_list.spy`).

We need to do the same for dicts:
  - add parser/ast support for dict literal expressions `{}` and `{a: b, ...}`

  - modify the builtins symtable to add `dict` as an alias to `_dict.dict`: https://github.com/spylang/spy/blob/86888c9b2c03807ab61deab99907a7c998f6d88b/spy/analyze/symtable.py#L198-L200

  - modify astframe.py and doppler.py to evaluate dict literal expressions as a series of calls to `_dict.dict` helpers

kanin-kearpimy:
Hi, if no one take this, I want to handle it. :)

PR #367, opened 2026-01-26
ISSUE-342: Add support for dict literals - parser support

kanin-kearpimy:
### Issue
- https://github.com/spylang/spy/issues/342

### Description
This PR adds support for parser of `dict` structure with test. It added new SPy AST nodes and parser to convert Python Dict to SPy Dict.

### DIFF
--- a/spy/tests/test_parser.py
+++ b/spy/tests/test_parser.py
@@ -650,6 +650,18 @@ class TestParser:
 
         self.assert_dump(stmt, expected)
 
+    def test_Dict_unpacking_unsupport(self):
+        src = """
+        def foo() -> None:
+            dict_unpacking = {"key3": 30}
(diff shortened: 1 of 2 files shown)


spy/tests/test_parser.py
```diff
@@ -39,10 +39,12 @@ def assert_dump(self, node: ast.Node, expected: str):
...
 
     def test_Module(self):
-        mod = self.parse("""
+        mod = self.parse(
+            """
         def foo() -> None:
             pass
-        """)
+        """
+        )
```
antocuni: still the same problem with autoformatting :(

kanin-kearpimy: I'm resolving in. I disabled everything but formatter is still there.

PR #370, opened 2026-01-27
ISSUE #342 - Add support for dict literals - astframe & scope

kanin-kearpimy:
This added support for ASTFrame, Scope, and dict literal unwrapped. Test are already added. 

However, I'm still working on astframe testing. If you have any recommendation, please let me know.

### DIFF
--- a/spy/tests/stdlib/test_dict.py
+++ b/spy/tests/stdlib/test_dict.py
@@ -1,5 +1,5 @@
 from spy.errors import SPyError
-from spy.tests.support import CompilerTest, only_interp
+from spy.tests.support import CompilerTest, only_interp, skip_backends
 from spy.vm.b import B
 
 
@@ -243,7 +243,7 @@ class TestDict(CompilerTest):
         assert not mod.test_neq_missing_key()
         assert not mod.test_neq_key()
 
-    @only_interp
+    @skip_backends("doppler", "C", reason="FIXME")
     def test_literal_stdlib(self):
         mod = self.compile("""
         def foo() -> dict[i32, i32]:
@@ -254,7 +254,7 @@ class TestDict(CompilerTest):
         print(x)
         assert x == {0: 1, 50: 2, 30: 3}
 
-    @only_interp
+    @skip_backends("doppler", "C", reason="FIXME")
     def test_literal_preserves_order(self):
         mod = self.compile("""
         def foo() -> dict[i32, i32]:
@@ -263,7 +263,7 @@ class TestDict(CompilerTest):
         x = mod.foo()
         assert list(x.keys()) == [1, 2, 3]
 
-    @only_interp
+    @skip_backends("doppler", "C", reason="FIXME")
     def test_empty_dict_unsupport(self):
         # this test must be removed when empty dict support is implemented.
         # it will come along with compiler part.
@@ -276,7 +276,7 @@ class TestDict(CompilerTest):
         ):
             mod.foo()
 
-    @only_interp
+    @skip_backends("doppler", "C", reason="FIXME")
     def test_literal_single_element(self):
         # useful for single element type
         mod = self.compile("""
@@ -286,17 +286,19 @@ class TestDict(CompilerTest):
         x = mod.foo()
         assert x[42] == 100
 
-    @only_interp
+    @skip_backends("doppler", "C", reason="FIXME")
     def test_literal_mixed_value_types_key_value(self):
         # useful for mixed type support
-        # type of x must be f64
-        # because union(i32, f64) = f64
+        # type of x must be i32
+        # because union(i32, i32) = i32
+        # However, mixed of different type like union(i32, f64) still not available yet
+        # we need to implement interp_dict for type fallback like interp_list
         mod = self.compile("""
-        def foo() -> dict[i32, f64]:
-            x: f64 = 1
-            y: f64 = 2.2
+        def foo() -> dict[i32, i32]:
+            x: i32 = 1
+            y: i32 = 1000
             return {0: x, 1: y}
         """)
         x = mod.foo()
         assert x[0] == 1
-        assert x[1] == 2.2
+        assert x[1] == 1000
--- a/spy/vm/astframe.py
+++ b/spy/vm/astframe.py
@@ -1188,6 +1188,7 @@ class AbstractFrame:
             else:
                 # second iteration and so on.
                 # compute union type covering both current key, value and new key, value type
+                assert w_valuetype is not None
                 w_keytype = self.vm.union_type(w_keytype, key.w_static_T)
                 w_valuetype = self.vm.union_type(w_valuetype, value.w_static_T)
 


spy/vm/astframe.py
```diff
@@ -1145,6 +1145,79 @@ def eval_expr_Tuple(self, op: ast.Tuple) -> W_MetaArg:
...
 
+    def eval_expr_Dict(self, dict: ast.Dict) -> W_MetaArg:
+        # 0. empty dict[k, v] are hanlded as immutable dictionary.
+        # We need to know type of [k, v] (e.g. [str, i32])
+        if len(dict.items) == 0:
+            # 0.1: Register empty dict type
+            w_T = SPY.w_EmptyDictType
+            # 0.2: Register empty dict value
+            w_val = SPY.w_empty_dict
+            return W_MetaArg(self.vm, "red", w_T, w_val, dict.loc)
```
antocuni: this seems wrong, because `SPY.w_EmptyDictType` doesn't exist.
It's fine to add support for empty dict later, but then here we should raise a clean `WIP` exception saying "empty dict literals are not supported yet".

For an example of using `WIP`, see here:
https://github.com/spylang/spy/blob/6e62219020216dcda1e78b77a5794d4aa84d2595/spy/vm/astframe.py#L554-L560

spy/vm/astframe.py
```diff
@@ -1145,6 +1145,79 @@ def eval_expr_Tuple(self, op: ast.Tuple) -> W_MetaArg:
...
+        key_value_pair = []
+        w_keytype = None
+        w_valuetype = None
+        key_color: Color = "red"  # copy from list eval
+        value_color: Color = "red"
+        for pair in dict.items:
+            key = self.eval_expr(pair.key)
+            value = self.eval_expr(pair.value)
+
+            # not sure...
```
antocuni:
```suggestion
            # If we have two blue items which happen to be equal, we reuse the same
            # w_opimpl for push() below, with the result of pushing the first item
            # twice, and the second item never. By making it red, we force to create a
            # more generic opimpl. See also the corresponding code in eval_expr_List
            
```

spy/vm/astframe.py
```diff
@@ -1145,6 +1145,79 @@ def eval_expr_Tuple(self, op: ast.Tuple) -> W_MetaArg:
...
+            w_T = SPY.w_EmptyDictType
+            # 0.2: Register empty dict value
+            w_val = SPY.w_empty_dict
+            return W_MetaArg(self.vm, "red", w_T, w_val, dict.loc)
+
+        # 1. evaluate type of key, value then infer the whole items type.
+        key_value_pair = []
+        w_keytype = None
+        w_valuetype = None
+        key_color: Color = "red"  # copy from list eval
```
antocuni:
```suggestion
        key_color: Color = "red"
```

spy/vm/astframe.py
```diff
@@ -1145,6 +1145,79 @@ def eval_expr_Tuple(self, op: ast.Tuple) -> W_MetaArg:
...
+            key_value_pair.append((key, value))
+            key_color = maybe_blue(key_color, key.color)
+            value_color = maybe_blue(value_color, value.color)
+            if w_keytype is None:
+                w_keytype = key.w_static_T
+            w_keytype = self.vm.union_type(w_keytype, key.w_static_T)
+
+            if w_valuetype is None:
+                w_valuetype = value.w_static_T
+            w_valuetype = self.vm.union_type(w_valuetype, value.w_static_T)
```
antocuni: this double `if` seems redundant: `w_keytype` and `w_valuetype` are `None` ONLY during the first iteration, then they become both not-None.
I think it's clearer to write it this way:
```
if w_keytype is None:
    assert w_valuetype is None
    w_keytype = key.w_static_T
    w_valuetype = value.w_static_T
...
```

spy/vm/astframe.py
```diff
@@ -1145,6 +1145,66 @@ def eval_expr_Tuple(self, op: ast.Tuple) -> W_MetaArg:
...
+        assert isinstance(w_T, W_Type)
+        fqn_setitem = w_T.fqn.join("__setitem__")
+        w_setitem = self.vm.lookup_global(fqn_setitem)
+        wam_setitem = W_MetaArg.from_w_obj(self.vm, w_setitem)
+
+        for pair, (wam_key, wam_val) in zip(dict.items, key_value_pair):
+            w_opimpl = self.vm.call_OP(
+                dict.loc, OP.w_CALL, [wam_setitem, wam_dict, wam_key, wam_val]
+            )
+            self.eval_opimpl(pair, w_opimpl, [wam_setitem, wam_dict, wam_key, wam_val])
```
antocuni, line 1216: this works fine for the `interp` mode, but it will fail for doppler.
For doppler, you will probably need to the same as `list` and implment a `_push` function which calls setitem *and* return the dict: this way we end up with a "chained" sequense of `_push` calls and the redshifter has a much easier way to deal with it.

spy/vm/struct.py
```diff
@@ -299,3 +303,34 @@ def unwrap_list(vm: "SPyVM", w_list: W_Object) -> list[Any]:
         w_item = vm.getitem_w(w_list, vm.wrap(i), color="red")
         items.append(vm.unwrap(w_item))
     return items
+
+
+def unwrap_dict(vm: "SPyVM", w_dict: W_Object) -> dict[Any, Any]:
+    """
+    Testing benefit
```
antocuni:
```suggestion
    Only useful in tests
```

let's use the same wording that we use for `list` ;)

spy/vm/astframe.py
kanin-kearpimy: added

spy/vm/astframe.py
kanin-kearpimy: Nice. Thank for your suggestion. Is `doppler` necessary for current PR? Otherwise, I can handle it with next PRs which focus on compiler part. What do you think? @antocuni

antocuni: it's fine to leaver `doppler` for the next PR 👍

spy/tests/stdlib/test_dict.py
```diff
@@ -241,3 +242,61 @@ def test_neq_key() -> bool:
         assert not mod.test_neq_value()
         assert not mod.test_neq_missing_key()
         assert not mod.test_neq_key()
+
+    @only_interp
```
antocuni:
```suggestion
    @skip_backends("doppler", "C", reason="FIXME")
```

antocuni: it's better to use `@skip_backends` here:
  - `@only_interp` means "this test is SUPPOSED to be run only in interp mode
  - `@skip_backends` means "this test SHOULD workd on all backends but some are broken for now, will fix later".

In other words: `@only_interp` is "forever", `@skip_backends` is temporary.

This applies also to all the other tests below

spy/tests/stdlib/test_dict.py
```diff
@@ -241,3 +242,61 @@ def test_neq_key() -> bool:
...
+        # because union(i32, f64) = f64
+        mod = self.compile("""
+        def foo() -> dict[i32, f64]:
+            x: f64 = 1
+            y: f64 = 2.2
+            return {0: x, 1: y}
+        """)
+        x = mod.foo()
+        assert x[0] == 1
+        assert x[1] == 2.2
```
antocuni: this test is interesting but the code contradicts the comment :)
`x` and `y` have type `f64` since the beginning (because they are explicitly typed that way), so we never do `union(i32, f64)`.

The code that corresponds to your comment is this:
```
x: i32 = 1
y: f64 = 2.2
return {0: x, 1: y}
```

kanin-kearpimy: Hi, I resolved this to [i32, i32] as in new code.
I noticed that dict doesn't have `interp_dict` as list `interp_list` yet. 
We need to implement it in compiler backend for mixed type (e.g. i32, f64).
As of now, I will leave it like new code.

Issue #252, opened 2025-10-06
Should C compiler warnings be considered bugs?


dpdani:
I had a test fail for a _call to undeclared function_. In the compiler error output we can also see a warning and that's what I want to consider.
The warning was there before I modified the code to perform the call to the undeclared function, but I could never see it.
The reason why the C compiler emits the warning is sensible and it should be fixed.

I'm thinking whether we should consider compilation warnings as errors.
It would have been pretty bad if the warning had never been seen.

```text
E           /Users/dp/repos/spy/stdlib/_dict.spy:56:34: warning: operator '<<' has lower precedence than '-'; '-' will be evaluated first [-Wshift-op-parentheses]
E              56 |     return 1 << data.p->log_size - 1;
E                 |              ~~ ~~~~~~~~~~~~~~~~~^~~
E           /Users/dp/repos/spy/stdlib/_dict.spy:56:34: note: place parentheses around the '-' expression to silence this warning
E              56 |     return 1 << data.p->log_size - 1;
E                 |                                  ^  
E                 |                 (                   )
E           /Users/dp/repos/spy/stdlib/_dict.spy:60:12: fatal error: call to undeclared function 'spy_builtins$hash_i32'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
E              60 |     return spy_builtins$hash_i32(key) & spy__dict$dict__builtins$i32_builtins$i32$mask(data);
E                 |            ^
```

PR #376, opened 2026-01-30
C compiler warnings

### LINKED ISSUE #369: Fix C warnings when compiling libspy
Compiling libspy generates a lot of C warnings, which they all seem related to the same line of code:

```
❯ make -C spy/libspy/
make: Entering directory '/home/antocuni/anaconda/spy/spy/libspy'
make TARGET=wasi BUILD_TYPE=release
make[1]: Entering directory '/home/antocuni/anaconda/spy/spy/libspy'
mkdir -p build/wasi/release/src/
/home/antocuni/anaconda/spy/venv/lib/python3.12/site-packages/ziglang/zig cc -fvisibility=hidden -Iinclude -fPIC -DNDEBUG -DSPY_RELEASE -O3 -DSPY_TARGET_WASI --target=wasm32-wasi-musl -mmultivalue -Xclang -target-abi -Xclang experimental-mv -mbulk-memory -c src/str.c -o build/wasi/release/src/str.o
In file included from src/str.c:1:
In file included from include/spy.h:46:
include/spy/operator.h:70:13: warning: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Wimplicit-const-int-float-conversion]
   70 |     if (x > INT32_MAX)
      |           ~ ^~~~~~~~~
[...]
```

I think this related to #347.

@Viriathus1 do you feel like taking a look?

Viriathus1:
Resolves #369 
I also added a warning free build flag that uses `-Werror` flag instead of the default as per this issue #252. I suppose for it to be enabled within the CI build then it will have to be added to pytest in some manner, let me know if you want this as well.

### DIFF
--- a/spy/build/config.py
+++ b/spy/build/config.py
@@ -32,7 +32,7 @@ LDFLAGS = [
 ]
 
 WARNING_CFLAGS = ["-Werror=implicit-function-declaration"]
-WARNING_AS_ERROR_CFLAGS = ["-Werror"]
+WARNING_AS_ERROR_CFLAGS = ["-Werror", "-Wno-unreachable-code"]
 
 RELEASE_CFLAGS  = ["-DSPY_RELEASE", "-O3", "-flto"]
 RELEASE_LDFLAGS = ["-flto"]
--- a/spy/libspy/Makefile
+++ b/spy/libspy/Makefile
@@ -45,6 +45,7 @@ else
     $(error Invalid BUILD_TYPE: $(BUILD_TYPE). Must be 'release' or 'debug')
 endif
 
+# All compiler warnings are treated as errors
 ifeq ($(SPY_WERROR), true)
 	CFLAGS += "-Werror"
 else ifeq ($(SPY_WERROR), 1)
--- a/spy/tests/compiler/test_py_cffi.py
+++ b/spy/tests/compiler/test_py_cffi.py
@@ -1,14 +1,7 @@
-import sys
-
-import pytest
-
 from spy.tests.support import CompilerTest, only_py_cffi
 
 
 @only_py_cffi
-@pytest.mark.skipif(
-    sys.platform == "darwin", reason="clang issues with -Wunreachable-code"
-)
 class TestPyCFFI(CompilerTest):
     def test_simple(self):
         mod = self.compile("""


spy/cli/commands/build.py
```diff
@@ -54,6 +54,11 @@ class _build_mixin:
         ),
     ] = 0
 
+    warning_free: Annotated[
+        bool,
+        Option("--warning-free", help="All compiler warnings are treated as errors"),
```
JeffersGlass: This might be bikeshedding, but I wonder if this wants to be called something like "--error-on-warn" or "[--werror](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)". To me `--warning-free` could mean that warnings are suppressed, rather than elevated.

antocuni: I agree with Jeff. I think we should call it `--Werror`, for consistency with gcc

spy/build/config.py
```diff
@@ -14,22 +14,24 @@ class BuildConfig:
...
+    warning_free: bool = False
 
 
 # ======= CFLAGS and LDFLAGS logic =======
 
 # fmt: off
 CFLAGS = [
     "--std=c99",
-    "-Werror=implicit-function-declaration",
-    "-Wfatal-errors",
```
antocuni, line 25: I think that we always want `-Wfatal-error` which, despite the name, has nothing to do with warnings (it means, "stop compiling at the first error")

antocuni (commented):
thank you, it looks like a good idea.

> I suppose for it to be enabled within the CI build then it will have to be added to pytest in some manner, let me know if you want this as well.

yes, that would be nice.
Probably the easiest thing to do is to use an env variable `SPY_WERROR`, which when it's set it automatically enables `--Werror`.

And we should probably do the same also in libspy's Makefile.

spy/cli/commands/build.py
Viriathus1: Sure I see how that can be confusing. I'll rename it to have consistency with gcc.

Viriathus1:
@antocuni yes it looks like it's only an issue on macOS. I've now disabled the comp flag.

PR #377, opened 2026-01-30
Add 'Contributing to the Docs' Section to Docs

JeffersGlass:
I was thinking it might be useful to add some information to the docs, since a couple of recent PRs have been really good learning opportunities for me, and @antocuni has shared lots of wisdom in various threads and conversations... and as a place to start, here's a PR on how to build and contribute to the docs. 😃

Also a small change to the deployment workflow to install the docs dependencies specified in `pyproject.toml` and pin their versions.

### DIFF
--- a/docs/src/contributing.md
+++ b/docs/src/contributing.md
@@ -8,7 +8,7 @@ For guidelines on setting up your machine to build and contribute to SPy, see [C
 The first two steps below (forking and cloning the repo) are the same as for the steps for contributing to SPy's source code. You do not need a separate fork and clone for documentation.
 ///
 
-To set up your machine to contribute to the SPy documentation, first Fork the repo <https://github.com/spylang/spy>, go to the project repo page and use the Fork button. This will create a copy of the repo, under your username (e.,g. <https://github.com/yourghname/spy>). (For more details on how to fork a repository see this guide <https://help.github.com/articles/fork-a-repo/>.)
+- Fork [the  SPy repo](https://github.com/spylang/spy), go to the project repo page and use the Fork button. This will create a copy of the repo, under your username (e.,g. `https://github.com/yourghname/spy`). (For more details on how to fork a repository see [this guide](https://help.github.com/articles/fork-a-repo/).
 
 Next, clone the repo to your machine:
 


docs/src/contributing.md
```diff
@@ -0,0 +1,73 @@
+## SPy Source
+
+For guidelines on setting up your machine to build and contribute to SPy, see [CONTRIBUTING.md](https://github.com/spylang/spy/blob/main/CONTRIBUTING.md)
+
+## Documentation
+
+/// note | Cloning the Repo
+The first two steps below (forking and cloning the repo) are the same as for the steps for contributing to SPy's source code. You do not need a separate fork and clone for documentation.
+///
```
antocuni, line 9: uh, interesting syntax, never saw it before. Is it the same as "admonitions"?

In my personal blog I use this kind of syntax:
https://github.com/antocuni/antocuni.github.io/blob/91930612f3b15df9ffb7f24f7c557af3326b8587/blog/posts/2025/10-spy-motivations-and-goals/index.md#L69-L78

```
!!! note "The myth of 100% compatibility"
    The vast majority of "compilers for Python" which exists or existed in the past are
    not "100% compatible". They don't support the full language even when they claim
    so. SPy prefers to be more explicit and honest about it.
```

I'm wondering whether it's just the same, or a different construct.

antocuni (approved):
wonderful! Merging

antocuni:
> @antocuni has shared lots of wisdom in various threads and conversations... 

it might actually be a good idea to have some kind of "unsorted info" section where to collect these: it's not proper documentation, but it's still probably better than not having it.

docs/src/contributing.md
JeffersGlass: Very similar! It seems that not too long ago, Pymdownx introduced a new "[blocks](https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/admonition/)" syntax ; this notation is from [`blocks.admonition`](https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/admonition/),  which is the 'new style' version of `admonitions`.

Issue #381, opened 2026-02-03
Fix the playground


antocuni:
The [SPy playground](https://spylang.github.io/spy/) is broken and fails with this errror:
```
Installing spy... Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 597, in eval_code_async
    await CodeRunner(
  File "/lib/python312.zip/_pyodide/_base.py", line 413, in run_async
    await coroutine
  File "<exec>", line 18, in <module>
  File "/lib/python3.12/site-packages/micropip/_commands/install.py", line 142, in install
    await transaction.gather_requirements(requirements)
  [...]
ValueError: Requested 'packaging>=24.0', but packaging==23.2 is already installed
```

I don't 100% know what's happening, but I **guess** it's something along these lines:
1. pyproject.toml specifies pinned versions
2. *indirect* dependencies are not pinned
3. pyodide comes with its own version of `packaging` (I'm not sure about this one, we should double check)
4. some indirect dependencies has been updated and now they conflict with pyodide's builtin `packaging`

JeffersGlass:
For what it's worth, upgrading to PyScript 2026.1.1 cures the symptom. Probably worth doing a `micropip.freeze()` and generating a lockfile so this doesn't happen again, but I'm not sure pyscript supports loading custom lockfiles directly. There's references to a `lockFileURL` param to pyconfig, but it's not documented.

PR #382, opened 2026-02-04
playground: Upgrade pyscript to 2026.1.1

aisipos:
Partial fix for #381 by updating pyscript to the latest version. It does fix the break and allows the playground to load correctly, but puts no protections in place to help detect or prevent this in the future.

I'll work on getting a micropip lockfile next. I'd like to upgrade pyodide to the latest as part of this, but there are many breaking changes since our current version (see https://pyodide.org/en/stable/project/changelog.html), so in the spirit of fixing the playground quickly I put up this quick PR first and will work on the micropip / pyodide part in a separate PR next.

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -6,9 +6,9 @@
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <link rel="icon" href="favicon.ico" type="image/x-icon">
 
-    <link rel="stylesheet" href="https://pyscript.net/releases/2025.2.1/core.css">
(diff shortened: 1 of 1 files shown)


antocuni:
> I'll work on getting a micropip lockfile next. I'd like to upgrade pyodide to the latest as part of this, but there are many breaking changes since our current version (see https://pyodide.org/en/stable/project/changelog.html), so in the spirit of fixing the playground quickly I put up this quick PR first and will work on the micropip / pyodide part in a separate PR next.

yes, makes a lot sense to deploy a quick fix and calmly do a "proper" fix. Merged, thank you!

PR #385, opened 2026-02-05
Add support for dict literals - C Backend

kanin-kearpimy:
This PR addresses C Backend of [dict literal](https://github.com/spylang/spy/issues/342).

What added:
- EmptyDict singleton.
- change from __setitem__ to _push function for dictionary item.

*Please note that `interp_dict` is left for next (and last, I hope!) PR.

### DIFF
--- a/spy/tests/stdlib/test_dict.py
+++ b/spy/tests/stdlib/test_dict.py
@@ -264,7 +264,6 @@ class TestDict(CompilerTest):
             return x
         """)
         x = mod.foo()
-        print(x)
         assert x == {0: 1, 50: 2, 30: 3}
 
     def test_literal_preserves_order(self):
--- a/spy/tests/wasm_wrapper.py
+++ b/spy/tests/wasm_wrapper.py
@@ -206,6 +206,19 @@ class WasmFuncWrapper:
         ll_ptr = pyres._content["__ll__"]
         assert isinstance(ll_ptr, WasmPtr)
 
+        # Read DictData struct from memory
+        # struct DictData {
+        #     ptr[i32] index;  // ptr is 8 bytes (4 for ptr + 4 for length)
+        #     i32 log_size;
+        #     i32 length;
+        #     ptr[Entry] entries;
+        # };
+        #
+        # struct Entry {
+        #     i32 empty;
+        #     i32 key;
+        #     i32 value;
+        # };
         addr = ll_ptr.addr
         index = self.ll.mem.read_i32(addr)
         log_size = self.ll.mem.read_i32(addr + 8)
@@ -219,7 +232,8 @@ class WasmFuncWrapper:
             empty = self.ll.mem.read_i32(entry_addr + 0)
             key = self.ll.mem.read_i32(entry_addr + 4)
             value = self.ll.mem.read_i32(entry_addr + 8)
-            result[key] = value
+            if not empty:
+                result[key] = value
 
         return result
 


stdlib/_dict.spy
```diff
@@ -292,4 +298,8 @@ def dict(Key, Value):
         def keys(self) -> _dict:
             return self
 
+        def _push(self, key: Key, val: Value) -> _dict:
+            self.__setitem__(key, val)
+            return self
+
```
dpdani, line 304: Is the new method necessary? It seems a bit redundant

spy/vm/modules/__spy__/__init__.py
```diff
@@ -79,4 +79,22 @@ def w_CALL_METHOD(
...
+    """
+    An object representing '{}'
+    """
+
+    def __init__(self) -> None:
+        # just a sanity check, W_EmptyDict is a singleton
+        raise Exception("You cannot instantiate W_EmptyDictType")
+
+    def __repr__(self) -> str:
+        return "<spy empty_dict {}>"
```
dpdani, line 93:
```suggestion
        return "{}"
```

Is there a reason for the extra verbosity?

dpdani (commented):
Hi there 👋

I just have a couple of questions for clarification, I think I'm missing some context here.

spy/tests/stdlib/test_dict.py
```diff
@@ -243,7 +243,21 @@ def test_neq_key() -> bool:
...
+            d = dict[i32, i32]()
+            d = d._push(1, 10)
+            d = d._push(2, 20)
+            d = d._push(3, 30)
+            return d[1] + d[2] + d[3]
+        """
+        mod = self.compile(src)
+        assert mod.test() == 60
+
+    @skip_backends("C", reason="FIXME")
```
antocuni: since this PR says that it adds support for dict literals in the C backend... we shouldn't skip the C backend tests 😅.

I see that the reason most of these test are failing is because when we compile with the C backend, we don't know how to unwrap the struct `DictData` into an app-level dict.

Doing it in the general case it's hard and probably not worth; you should use the same approach that we use for lists: add a special case to unwrap `dict[i32, i32]` which is used only by tests, and be happy with that:
https://github.com/spylang/spy/blob/c618310055c5d7056ef0c1d7625c333934b52223/spy/tests/wasm_wrapper.py#L163-L165

spy/vm/modules/__spy__/__init__.py
antocuni: yes, it's following the same convention that we use for empty lists:
https://github.com/spylang/spy/blob/c618310055c5d7056ef0c1d7625c333934b52223/spy/vm/modules/__spy__/__init__.py#L44-L55

The reason why they exist is the following (I'm using lists because they are slightly easier, but the same applies to dicts):
```
# here we know that STATIC_TYPE(a) is list[i32], we can infer from the type of elements
a = [1, 2, 3]

# what is STATIC_TYPE(b) here?
b = []
```

we cannot infer the item type for `b`; we *could* infer it from subsequent operations (e.g. if we call `append()`) but it would require a type inference engine which we don't have (and I'm not sure that we want).

So the solution which I came up with "temporarily" is to declare that "empty list" is a special type which is **immutable**, but can be converted to any specialized type list.

So, the following is an error:
```
b = []
b.append(1) # TypeError: cannot mutate an untyped empty list
```

The only supported way to mutate empty lists is to immediately convert them into a typed one:
```
b: list[i32] = []
b.append(1) # WORKS
```

the code above is equivalent to say:
```
b: list[i32] = list[i32]()
b.append(1)
```

So, the "empty list literal" is a very strange beast; that's why I chose to give it a special repr, to avoid confusion with a "real" typed empty list.

stdlib/_dict.spy
antocuni: yes, it's needed to implement dict literals; it's the same hack that we use for lists.
The "problem" is that a list/dict literal is an expression, but in order to initialize them we need a series of method calls, and we want the result of the expression to be the list/dict itself.

The solution which I found for lists is to desugar this:
```
[1, 2, 3]
```
into
```
_push(_push(_push(newlist(), 1), 2), 3)
```

This is the same logic, but for dicts.
See `astframe.eval_expr_Dict`

spy/vm/modules/__spy__/__init__.py
dpdani: Ahh I see now, thanks 👍

spy/tests/stdlib/test_dict.py
kanin-kearpimy: Okay, I added it up.

PR #392, opened 2026-02-07
Implementation of the POW (**) Operator

tedpatrick:
This pull request adds support for the power operator (`**`) across the SPy virtual machine and its standard numeric types, including both integer and floating-point types. It introduces new operator implementations, updates the operator dispatch tables, and adds comprehensive tests to ensure correct behavior for power operations. Additionally, it improves the atomicity of cache file writes to prevent issues with concurrent access.

### Power operator (`**`) support

* Added the `**` operator to the operator dispatch tables for all relevant numeric types (`i8`, `u8`, `i32`, `u32`, `f32`, `f64`) in `spy/vm/modules/operator/binop.py` and registered the corresponding implementations. [[1]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R30) [[2]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R50) [[3]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R70) [[4]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R90) [[5]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R105) [[6]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R125-R126) [[7]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R147) [[8]](diffhunk://#diff-2a5fe649f192f146bcff6fbae73263bcde0fb09e8f159dab7819b9c87ed99345R164-R165) [[9]](diffhunk://#diff-d262d0746c5c71f756cc0d019dabfe0e81a0693d3029602c96167f97a54c0c45R80)
* Implemented the `w_POW` operator dispatch function, which handles the power operation and falls back to `__pow__` if available.
* Added per-type implementations for the power operation in the operator implementation files: `w_f32_pow` in `opimpl_f32.py`, `w_f64_pow` in `opimpl_f64.py`, and generic integer `w_pow` in `opimpl_int.py`. [[1]](diffhunk://#diff-89ee3eefbd514617168e39691154eb506013e2dd369265989652fd480a6636a4R57-R61) [[2]](diffhunk://#diff-6188b8788064e6f57faf8657a853e0ed884c0e481db24bc0a8dc4901c88a9fcbR67-R71) [[3]](diffhunk://#diff-2c4e36d84749e19a0e8a7de20e454007fd450fd3bb61ec7f0e200a633ef91ed6R89-R92)
* Added C-level implementations for power operations for all relevant numeric types in `spy/libspy/include/spy/operator.h` and `spy/libspy/src/operator.c`. [[1]](diffhunk://#diff-8bd632e0285a668b81aecc65cf75cfca94cb450af8302837d6eb6b1a4e2ba47bR451) [[2]](diffhunk://#diff-8bd632e0285a668b81aecc65cf75cfca94cb450af8302837d6eb6b1a4e2ba47bR504-R576) [[3]](diffhunk://#diff-bb5a8100b58e850ee5ecac858b9c6be3887c4e05b56035e950edf74e7a09c35cR123-R127)

### Testing

* Added comprehensive tests for the `**` operator for both integer and floating-point types, including tests for negative bases, zero exponents, and fractional exponents, in `spy/tests/compiler/test_int.py` and `spy/tests/compiler/test_float.py`. [[1]](diffhunk://#diff-55bb5a1e26902fe1b24e6d4812aeedcc72dbf06523aca25f769572b0103781f4R148-R173) [[2]](diffhunk://#diff-3d798c7ce9777ca6d5371f16b70593f39314165cab9cbcbd4cf9ee83ca5ccb33R133-R145)

### Reliability improvements

* Improved atomicity of cache file writes in `spy/analyze/importing.py` by writing to a temporary file and renaming it, preventing partial writes in concurrent environments.


tedpatrick:
### C layer:
- spy/libspy/src/operator.c — added spy_operator$f32_pow using powf(x, y)
- spy/libspy/include/spy/operator.h — added WASM_EXPORT declaration for spy_operator$f32_pow
- Rebuilt libspy.wasm (debug) so the new symbol is available at runtime

### Python layer:
- spy/vm/modules/operator/opimpl_f32.py — added w_f32_pow delegating to _f32_op_f32(..., "pow")
- spy/vm/modules/operator/binop.py — registered "**" for f32/f32, and both directions of f32/i32 mixed
### Tests:
- spy/tests/compiler/test_float.py — added test_pow using the float_type fixture (runs for both f32 and f64 across all three backends)
- spy/tests/compiler/test_int.py — added test_pow and test_pow_negative_base using the int_type fixture (covers i32, u32, i8, u8)
- spy/tests/compiler/test_pow.py — deleted

tedpatrick:
Closes #175

spy/libspy/include/spy/operator.h
```diff
@@ -500,4 +501,77 @@ spy_operator$bool_not(bool x) {
...
+        b *= b;
+        e >>= 1;
+    }
+    return result;
+}
+
+static inline int32_t
+spy_operator$i32_pow(int32_t base, int32_t exp) {
+    if (exp == 0) return 1;
+    if (exp < 0) return 0;
```
Viriathus1: Python allows negative exponents for ints but the returned result must be a float - this also affects `i8`.

spy/tests/compiler/test_int.py
Viriathus1, line 1: Needs to have unit tests that cover negative exponents also ZeroDivisionError check for zero raised to a negative exponent.

spy/tests/compiler/test_float.py
Viriathus1, line 1: Should also have test cases for negative bases, ZeroDivisionError and bases between 0 and 1 because they exhibit slightly different behaviour.

Viriathus1 (changes requested):
Thanks for the changes on this one. There's a few minor things that still need to be rectified. It needs to conform to Python specs more broadly as described [here](https://docs.python.org/3/reference/expressions.html#the-power-operator). The following is currently unaccounted for:

- For int operands, the result must be a float if the exponent arg is negative.
- Zero with a negative exponent raises a ZeroDivisionError exception.
- Negative base with fractional exponent returns a complex number - eg: `(-5) ** 0.5 == (1.3691967456605067e-16+2.23606797749979j)`

spy/tests/compiler/test_float.py
tedpatrick: added

spy/tests/compiler/test_int.py
tedpatrick: added

spy/libspy/include/spy/operator.h
tedpatrick: added

Viriathus1:
Hi @tedpatrick, thanks for the changes. I see that you widen the result to `f64` for all pow operations involving signed ints. It doesn't exactly match python semantics but I don't see how you could do anything apart from inspecting the rhs `W_MetaArg` for a concrete value which can only be done when they're blue. This extends to complex number situation as well.

What do you think @antocuni? Union types are probably required here which have yet to be implemented.

Viriathus1:
@antocuni when you have time, can you take a look at this pr?

antocuni:
(full disclosure for those who aren't aware: I know Ted personally as he's my colleague at Anaconda, and he was my manager _at the time SPy was born_ :) ).

@tedpatrick thanks for this and sorry for taking so long.
@Viriathus1 thanks a lot for your review. Your comments helped me to realize that the semantics of `**` in a language like SPy is not obvious at all. I didn't foresee it!

Long story short: we cannot follow CPython here because we we discovered on CPython the result _type_ depends on the input _values_, and since we are statically typed we must decide the result type solely based on the input types.  This is a real language design issue.

I had some conversation with claude which helped to clarify my thoughts about it. I will copy&paste the summary below. Please note that even if the text is AI generated, the actual design and thinking is mine :).

I propose the following procedure to unblock this PR and finally be able to merge it:
1. @tedpatrick and @Viriathus1: as a first step, I'd like your opinion on the design decision which I took and which are explained below, I hope you will agree with them but in case you don't, please let me know
2. assuming that we have agreement on how to proceed, @tedpatrick should rebase on top of the current `main` (there are conflicts) and modify the PR according to the plan
3. I delegate @Viriathus1 to review and merge it when it's ready (I think you have merging right, don't you?)

What follows is the plan with the rationale

antocuni:
(the following has been written by claude after an extensive brainstorming session with me. All the ideas are mine and/or approved by me).

### Main issue: integer `**` should return the base type, not f64

Currently signed ints return `f64` while unsigned ints return the int type. Two problems:

1. **Signed/unsigned asymmetry is hard to justify.** There's no principled reason `i32 ** i32 -> f64` but `u32 ** u32 -> u32`.

2. **The interp and C backends compute *different results*.** The interp does `float(int(a) ** int(b))` (arbitrary precision, then cast), while the C side does integer exponentiation-by-squaring in the fixed-width type (which *wraps*). So e.g. `2**10` on `i8` is `1024.0` on interp but a wrapped value on C. The current tests never use a value big enough to overflow, so this divergence is invisible right now.

Every statically-typed language we looked at keeps `int ** int` as an integer (result = base type) and makes float power a separate explicit operation:
- **Rust:** `i32::pow(self, exp: u32) -> i32` — base type, unsigned exponent (negatives impossible), opt into float via `powi`/`powf`.
- **Zig:** `powi(T, x: T, y: T) !T` — base/exp/result all `T`; negative exp → `error.Underflow`, overflow → `error.Overflow`.
- **Go:** no integer pow at all, only `math.Pow(float64, float64)`.

It's tempting to argue "match SPy's `int / int -> f64`", but `/` is *inherently fractional* (that's why CPython promoted only `/`), whereas `int ** non-negative-int` is mathematically an integer. The right analogy is `*`: `**` is repeated multiplication and should share multiplication's type and overflow behavior.

### Proposed semantics (model A)

| operation | result | example (i32) |
|---|---|---|
| `iN ** iN` / `uN ** uN` | base type, **wraps** | `2**31 -> i32 min` |
| `fN ** fN` | `fN` | `2.0**0.5 -> 1.414…` |
| `int ** negative int` | raises `ValueError` | `2 ** -1` |
| `0 ** negative int` | raises `ZeroDivisionError` | `0 ** -1` |

- Overflow wraps, consistent with `i32 * i32` under `-fwrapv`. (We may revisit and add `OverflowError` detection across all ops later, but for now we stay consistent with multiplication.)
- Want a float? Opt in explicitly: `f64(x) ** y`. The f64/f32 paths you already added cover this.
- Nice consequence: `2**31 - 1 == i32 max` works by two's-complement (the `2**31` overflow and the `-1` cancel), so it's a good regression test.

### Concrete changes requested

- Unify `make_signed_pow`/`make_unsigned_pow` into one helper returning the **base type**; do the exponentiation in the fixedint type so it wraps.
- Add a negative-exponent check raising `ValueError`; keep `0 ** -1 -> ZeroDivisionError`.
- Register `**` for `i64`/`u64` too (currently missing).
- C side: return the integer type (not `double`) so it matches the interp.
- Tests: add `2**31 - 1 == 2147483647`, a plain overflow case to lock wrapping, and a negative-exponent-raises case. The float tests are fine as-is.

Minor: trailing whitespace on the blank lines inside `u8_pow`/`u32_pow` in `operator.h`. Also the PR description mentions an `importing.py` atomic-write change that isn't in the diff — probably stale, worth removing from the description.

Viriathus1:
@antocuni The proposed semantics makes sense to me and happy to merge it when ready.

On another note, it may be useful to start compiling deviations from CPython semantics for future users of SPy as I expect they'll mostly be users of Python. Could start with the power operator and then add as necessary?

tedpatrick:
Makes sense. Rebasing and starting changes.

tedpatrick:
My changes are in, and I am reviewing the entire set to ensure it aligns with the desired approach.

JeffersGlass:
@Viriathus1 
> On another note, it may be useful to start compiling deviations from CPython semantics for future users of SPy as I expect they'll mostly be users of Python. Could start with the power operator and then add as necessary?

I think this is a great idea, in particular a section of documentation for _operators_ that may behave differently than expected. 

I wonder if some of the differences that came up in #520 could fit there as well - e.g. the `in` operator will just throw a type error if the type of the container doesn't match the operand, instead of just returning False (which I think is good!). 

(Technically those are behaviors of the container types themselves and not the operator, but I think documenting surprising behavior in multiple places isn't bad 😄 )

antocuni:
> @antocuni The proposed semantics makes sense to me and happy to merge it when ready.

thank you a lot for this, it helps a lot :pray: 

> On another note, it may be useful to start compiling deviations from CPython semantics for future users of SPy as I expect they'll mostly be users of Python. Could start with the power operator and then add as necessary?

yes, I think it's a great idea. I see that @JeffersGlass is already on it 🎉

spy/vm/modules/operator/opimpl_int.py
Viriathus1, line 1: This all works but it should be a sub function in `make_ops` as per the rest of the operators.

spy/tests/compiler/test_int.py
```diff
@@ -279,6 +284,60 @@ def b_xor(x: T, y: T) -> T: return x ^ y
...
+            pytest.skip("Skipping negative base test for unsigned types")
+        mod = self.compile(f"""
+        T = {int_type}
+        def pow(x: T, y: T) -> T:
+            return x ** y
+        """)
+        assert mod.pow(-2, 3) == -8
+        assert mod.pow(-2, 2) == 4
+        assert mod.pow(-1, 5) == -1
+        assert mod.pow(-1, 4) == 1
```
Viriathus1: Should use the `signed_int_type` parameter here and remove the conditional check.

Viriathus1 (changes requested):
thanks @tedpatrick, just some quick non-functional changes and then it can be merged 🙂

Viriathus1:
Nice @JeffersGlass I'll create an issue for it and everyone can start collecting deviations going forward.

spy/vm/modules/operator/opimpl_int.py
tedpatrick: fixed in https://github.com/spylang/spy/pull/392/changes/11601579c8f582719bead3851384814087fd5eb1

spy/tests/compiler/test_int.py
tedpatrick: Fixed in https://github.com/spylang/spy/pull/392/changes/579f0376753de7efd5365f8601664059f5d18d4d

tedpatrick:
@Viriathus1 Requested changes are in. Need 1 ✅

Viriathus1 (approved):

Viriathus1:
Thanks @tedpatrick, going to merge it now 🚀

PR #395, opened 2026-02-11
Allow method calls on refs

### LINKED ISSUE #380: Make it possible to call methods via raw_ref
consider this example:

```python
from unsafe import raw_alloc, raw_ref

@struct
class Point:
    x: i32
    y: i32

    def foo(self) -> i32:
        return self.x + self.y

def main() -> None:
    p: raw_ptr[Point] = raw_alloc[Point](1)
    r: raw_ref[Point] = p[0]
    print(r.foo())
```

It currently fails but it should work:

```
TypeError: method `unsafe::raw_ref[x::Point]::foo` does not exist
  | /tmp/x.spy:15
  |     print(r.foo())
  |           ^ this is `unsafe::raw_ref[x::Point]`
```

Similarly, it should be possible to compare `raw_ref[T] == T`.
I **think** that this should work out of the box as soon as we implement method lookup on raw_ref (because it would find and call the `__eq__`). If it doesn't, we need to open a new issue to track that.

JeffersGlass:
Implements `__call_method__` on references, so that it's possible to treat refs like the pointers they refer to for the purposes of method calls.

Closes #380.

As predicted in that issue, does fix the standing issue with creating lists-of-structs:
```py
@struct
class C:
    x: int

def main() -> None:
    l = [C(0), C(1)]
```


spy/vm/modules/unsafe/ptr.py
```diff
@@ -449,6 +449,23 @@ def w_ptr_deref(vm: "SPyVM", w_ref: REF) -> T:
...
 
+    @builtin_method("__call_method__", color="blue", kind="metafunc")
+    @staticmethod
+    def w_CALL_METHOD(
+        vm: "SPyVM", wam_T: "W_MetaArg", wam_name: "W_MetaArg", *args_wam: "W_MetaArg"
+    ) -> "W_OpSpec":
+        ref_T = wam_T.w_static_T
+        assert isinstance(ref_T, W_RefType)
+        w_T = ref_T.w_itemT
+        assert isinstance(w_T, W_Type)
```
antocuni: nitpick, but is this `assert` actually needed? `W_RefType.w_T` should already be typed as `W_Type`, shouldn't it?

JeffersGlass: Ah yeah that's fair, it's automatically typed correctly. I'll remove the assert.

PR #398, opened 2026-02-23
Complex numbers

### LINKED ISSUE #305: Add complex numbers

Viriathus1:
Resolves #305 

This is missing kwargs for explicit complex number init as in `complex(real=1, imag=2)` also missing `abs` function for this type. Furthermore I didn't use the inbuilt complex type as defined in the c99 standard because I couldn't get it to work with wasm output specifically with function args. If there's a way to use them then I'll be happy to know.

### DIFF
--- a/spy/tests/compiler/test_complex.py
+++ b/spy/tests/compiler/test_complex.py
@@ -105,22 +105,20 @@ class TestComplex(CompilerTest):
         assert mod.str_to_complex("-4.5j") == -4.5j
         assert mod.str_to_complex("-1.23+4.5j") == -1.23 + 4.5j
         assert mod.str_to_complex("\t( -1.23+4.5J )\n") == -1.23 + 4.5j
-        with SPyError.raises(
-            "W_ValueError", match=r"complex\(\) arg is a malformed string"
+        for str_arg in (
+            " ",
+            "(",
+            "  )",
+            " (   ) ",
+            " ( -1.23+4.5j  } ",
+            "-4.5j+1.23",
+            "1 + 2j",
+            "-1.23*4.5j",
         ):
-            mod.str_to_complex(" ")
-        with SPyError.raises(
-            "W_ValueError", match=r"complex\(\) arg is a malformed string"
-        ):
-            mod.str_to_complex(" (   ) ")
-        with SPyError.raises(
-            "W_ValueError", match=r"complex\(\) arg is a malformed string"
-        ):
-            mod.str_to_complex("-4.5j+1.23")
-        with SPyError.raises(
-            "W_ValueError", match=r"complex\(\) arg is a malformed string"
-        ):
-            mod.str_to_complex("1 + 2j")
+            with SPyError.raises(
+                "W_ValueError", match=r"complex\(\) arg is a malformed string"
+            ):
+                mod.str_to_complex(str_arg)
         c = mod.str_to_complex("-Infinity+NaNj")
         assert isinf(c.real)
         assert isnan(c.imag)


spy/libspy/include/spy/complex.h
```diff
@@ -0,0 +1,99 @@
...
+static inline double
+spy_complex128_get_imag(spy_Complex128 x) {
+    return x.imag;
+}
+
+static inline spy_Complex128
+spy_complex128_conjugate(spy_Complex128 x) {
+    x.imag = -x.imag;
+    return x;
+}
```
antocuni, line 34: nitpick, but this looks a bit weird because it looks like it's modifying `x`, while in practice it's making a copy (which is right).
What about something like:
```
return (spy_Complex128){ x.real, -x.imag };
```

?

spy/libspy/src/str.c
```diff
@@ -251,3 +251,118 @@ spy_operator$str_to_u8(spy_Str *s) {
     spy_check_range(val, 0LL, 255LL, "u8");
     return (uint8_t)val;
 }
+
+spy_Complex128
+spy_operator$str_to_complex128(spy_Str *s) {
```
antocuni: I admit I didn't read this in details, but I trust it's correct as long as there are enough tests to cover all the cases :).
Did you follow TDD to develop it?

Also, just a random thought: ideally in the long term we would like to write this kind of logic directly in SPy, but I agree that for now it makes more sense to write it in C, so that's good 👍

spy/tests/compiler/test_complex.py
```diff
@@ -0,0 +1,135 @@
...
+        T = {complex_type}
+        def add_int(x: T, y: i32) -> T: return x + y
+        def add_float(x: T, y: f64) -> T: return x + y
+        def sub_int(x: i32, y: T) -> T: return x - y
+        def sub_float(x: f64, y: T) -> T: return x - y
+        def mul_int(x: T, y: i32) -> T: return x * y
+        def mul_float(x: T, y: f64) -> T: return x * y
+        def div_int(x: i32, y: T) -> T: return x / y
+        def div_float(x: f64, y: T) -> T: return x / y
+        """)
```
antocuni, line 78: I'm not sure whether this is nitpick or whether I'm missing something important.
The test is called `test_implicit_conversion`, so I expect it tests conversion from i32/f64 to complex, right?
Then, I don't understand why you need the whole `add/sub/mul/div`.
To test the conversion itself this should be enough:

```
def from_complex(x: complex128) -> T:
    return x

def to_complex(x: T) -> complex128:
    return x
```

where T in `(i32, f64)`.

By using `+` etc. you still test it by very indirectly, because `x + y` invokes `complex128_add`, which in turns requires a type conversion to be called.

After finishing writing this comment I realized that you probably took inspiration from `test_float.test_implicit_conversion`... so ultimately it's my fault :).
But the explanation still stand, so probably we should change the name of the test to e.g. `test_mixed_types` (both here and in test_float0. Suggestions for a better name are welcome :)

spy/tests/compiler/test_complex.py
```diff
@@ -0,0 +1,135 @@
...
+            "W_ValueError", match=r"complex\(\) arg is a malformed string"
+        ):
+            mod.str_to_complex("-4.5j+1.23")
+        with SPyError.raises(
+            "W_ValueError", match=r"complex\(\) arg is a malformed string"
+        ):
+            mod.str_to_complex("1 + 2j")
+        c = mod.str_to_complex("-Infinity+NaNj")
+        assert isinf(c.real)
+        assert isnan(c.imag)
```
antocuni, line 124: 💪

spy/tests/test_parser.py
```diff
@@ -407,19 +407,6 @@ def foo() -> None:
...
-        src = """
-        def foo() -> i32:
-            return 42j
-        """
-        self.expect_errors(
-            src,
-            "unsupported literal: 42j",
-            ("this is not supported yet", "42j"),
-        )
-
```
antocuni, line 422: woooooo 🎉

spy/vm/modules/operator/opimpl_str.py
```diff
@@ -83,3 +83,13 @@ def w_str_to_u8(vm: "SPyVM", w_s: W_Str) -> W_U8:
     val = _parse_int(vm, w_s)
     _check_range(val, 0, 255, "u8")
     return W_U8(val)
+
+
+@OP.builtin_func
+def w_str_to_complex128(vm: "SPyVM", w_x: W_Str) -> W_Complex128:
+    try:
+        val = complex(vm.unwrap_str(w_x))
+    except ValueError:
```
antocuni: this relies on cpython's complex parsing, but since we have our own non-trivial parsing logic in C, maybe it would be better to use it also from here?
Something like `vm.ll.call("spy_operator$str_to_complex128", ...)`?

spy/vm/primitive.py
```diff
@@ -275,6 +276,67 @@ def w_str(vm: "SPyVM", w_self: "W_F32") -> "W_Str":
         return vm.wrap(f"{f:.7g}")
 
 
+@B.builtin_type("complex128", lazy_definition=True)
+class W_Complex128(W_Object):
+    __spy_storage_category__ = "value"
+    value: complex
+    w_real: Annotated[W_F64, Member("real")]
+    w_imag: Annotated[W_F64, Member("imag")]
+
```
antocuni, line 285: you might want to consider using `B.struct_type` here, instead of `B.builtin_type`.

By using `struct_type`, the compiler knows it's an actual struct and knows how to get/set fields, without having to do it manually. E.g. it would allow you to get rid of e.g. `spy_complex128_get_imag` in the C src, etc.

See PR #396 which introduced this possibility and applied the same transformation to `posix.TerminalSize`.

spy/libspy/src/str.c
Viriathus1: Yeah it was good to use TDD in this case, I started with the parsing examples from the python docs and then went from there. Yes when we can move it to SPy proper it will be a good day.

spy/tests/compiler/test_complex.py
Viriathus1: Yes as you say at the end it was transferred across from the float unit test but all good the names have been updated in a more recent commit.

PR #400, opened 2026-02-24
Add A "CPython Builtins" Page to Docs

JeffersGlass:
Adds a page to the docs that illustrates which built-ins from CPython are currently implemented and which are not. Inspired by `setattr` and `getattr` getting added.

There's surely more info that could end up on this page - like which builtins are (likely) to be introduced down the road and which are not - but it seemed like a useful start.

Also working on pages for SPy builtins and modules, but this seemed like the simplest place to start and to think about formatting as well, and will leave those for future PRs.

### DIFF
--- a/docs/src/api_reference/python_builtins.md
+++ b/docs/src/api_reference/python_builtins.md
@@ -19,16 +19,18 @@ The following built-in functions work similarly to their equivalents in CPython;
 
 :   Drops the user into an interactive SPy debugging session via `spdb`, a `pdb-like` debugging interface for SPy.
 
-### __dict__\[type\]() { data-toc-label='dict()' }
-:   In SPy, `dict` must always be fully typed and used as `dict[keytype, valuetype]`. The syntax `dict[keytype, valuetype]()` can be used to create a new empty dict of the given types; unlike CPython, this does not (currently) accept an Iterable to create a new dict from. The simpler syntax `d: dict[keytype, valuetype] = {}` can also be used.
+### __dict__\[keytype, valuetype\]() { data-toc-label='dict()' }
+:   In SPy, `dict` must always be fully typed and used as `dict[keytype, valuetype]`., The syntax `dict[keytype, valuetype]()` can be used to create a new empty dict of the given types. The simpler syntax `d: dict[keytype, valuetype] = {}` can also be used.
 
-:   The implemention (in SPy) of `dict` can be [viewed here](https://github.com/spylang/spy/blob/main/stdlib/_dict.spy).
+:   Unlike CPython, this does not (currently) accept an Iterable to create a new dict from.
+
+:   The implementation (in SPy) of `dict` can be [viewed here](https://github.com/spylang/spy/blob/main/stdlib/_dict.spy).
 
 ### __dir__(object) {data-toc-label='dir()'}
 
 :   Returns a list of object’s attributes’ names, the names of its class’s attributes, and recursively of the attributes of its class’s base classes. `dir(type)` is not currently implemented.
 
-:   The no-argument form of `dir()` a-la CPython (i.e. print local variables) is not implemented. Custom `__dir__` methods on objects are not currently supported.
+:   The no-argument form of `dir()` (i.e. print local variables) is not currently implemented. Custom `__dir__` methods on objects are not currently supported.
 
 ### __float__(object) { data-toc-label='float()' }
 
@@ -36,16 +38,18 @@ The following built-in functions work similarly to their equivalents in CPython;
 
 ### __getattr__(obj, name: str) { data-toc-label='getattr()' }
 
-:   `attr` must be blue
+:   Return the value of the named attribute of object. `attr` must be blue
 
 ### __hash__(object) { data-toc-label='hash()' }
 :   Currently implemented for types: `i8`,`i32`, `u8`, `bool`, `str`.
 
-:   By default, instances of SPy structs are not hashable. A future feature will be that structs will have auto-generated `__hash__` by default, btu that is awaiting implementation. Currently, users can implement the `__hash__` function to permit hashing. 
+:   By default, instances of SPy structs are not hashable. As a planned future feature, structs will have auto-generated `__hash__` by default, but this is awaiting implementation. Currently, users can implement the `__hash__` function to permit hashing. 
 
 ### __int__(object) { data-toc-label='int()' }
 
-:    Converts `object` to an int if able. Works for number types, as well as strings. The `int` type is currently an alias to `i32`. In the future, `int` will alias preferred individual types for specific platforms, but currently it is always `i32`.
+:   Converts `object` to an int if able. Works for number types, as well as strings.
+
+:   The `int` type is currently an alias to `i32`. In the future, `int` will alias preferred individual types for specific platforms, but currently it is always `i32`.
 
 ### __len__(object) { data-toc-label='len()' }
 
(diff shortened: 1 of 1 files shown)


docs/src/api_reference/python_builtins.md
```diff
@@ -85,4 +85,40 @@ Creates an iterable set of indices between `start` and `stop`, jumping over `ste
...
+
+### Async
+
+SPy does not currently have an async story.
+
+:   aiter(), anext()
+
+### Iterables and Iterators
+
+Iterables and collections are very much an area of active developmen; their API solidifies, these types of builtins should become more straightfoward to implement.
```
antocuni: we should also note that *generators* are not supported ATM

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+The following built-in functions work similarly to their equivalents in CPython; see the specific functions below for notes
+
+### __abs__(object) { data-toc-label='abs()' }
+
+:   Currently only implemented for int32's or objects convertible to int32's. The `__abs__` attribute is not currently supported.
+
+### __dict__\[type\]() { data-toc-label='dict()' }
+:   The syntax `dict[keytype, valuetype]()` can be used to create a new empty dict of the given types; unlike CPython, this does not (currently) accept an Iterable to create a new dict from.
+
```
antocuni, line 34: this is technically correct, although the more straightforward way to create an empty dict is this:
```
d: dict[K, V] = {}
```

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+## Implemented CPython-Like Built-ins 
+
+The following built-in functions work similarly to their equivalents in CPython; see the specific functions below for notes
+
+### __abs__(object) { data-toc-label='abs()' }
+
+:   Currently only implemented for int32's or objects convertible to int32's. The `__abs__` attribute is not currently supported.
+
+### __dict__\[type\]() { data-toc-label='dict()' }
+:   The syntax `dict[keytype, valuetype]()` can be used to create a new empty dict of the given types; unlike CPython, this does not (currently) accept an Iterable to create a new dict from.
```
antocuni:
```suggestion
:   in SPy, `dict` must always be fully typed and used as `dict[keytype, valuetype]`. The syntax `dict[keytype, valuetype]()` can be used to create a new empty dict of the given types; unlike CPython, this does not (currently) accept an Iterable to create a new dict from.
```

docs/src/api_reference/python_builtins.md
antocuni: also, we should probably explain that `dict` is implemented in SPy and put a link to `stdlib/_dict.spy`

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### __float__(object) { data-toc-label='float()' }
+
+:   Converts `object` to a float if able
```
antocuni: I'd also say that this is just an alias to `f64`

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### __getattr__(obj, name: str) { data-toc-label='getattr()' }
+
+:   `attr` must be blue
+
+### __hash__(object) { data-toc-label='hash()' }
+:   Currently implemented for types: `i8`,`i32`, `u8`, `bool`, `str`.
+
+:   By default, instances of SPy classes are not hashable are not hashable. Users can implement the `__hash__` function to permit hashing.
```
antocuni: this is a bit misleading since we don't have SPy **classes** yet. We have structs though.
Currently, that you say it's technically correct, although the plan is that struct will have an autogenerated `__hash__` by default

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### __int__(object) { data-toc-label='int()' }
+
+:    Converts `object` to an int if able. Works for number types, as well as strings.
+
+:    For conversion to specific integer types, see also [str_to_i32()](), [str_to_u32()](), [str_to_i8()](), [str_to_u8()]().
```
antocuni: the various `str_to_*` are just implementation details in libspy and I don't think they are visible from user code, are they? I'd probably just remove this sentence.

As for `float`, we should say that this is an alias to `i32`.
Eventually, we will support multiple target platforms and thus `int` will be an alias to the most appropriate integral type for that platform, but for now it's hardcoded to `i32`

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### __len__(object) { data-toc-label='len()' }
+
+:   Return the length (the number of items) in a container
+
+### __list__\[type\]() { data-toc-label='list()' }
+
+:   The syntax `list[type]()` can be used to create a new empty list of the given type; unlike CPython, this does not (currently) accept an Iterable to create a new list from.
```
antocuni: same comments as `dict`

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### __max__(x: i32, y: i32) { data-toc-label='max()' }
+
+:   Currently only implemented for int32's or objects convertible to int32's.
+
+### __max__(x: i32, y: i32) { data-toc-label='min()' }
+
+:   Currently only implemented for int32's or objects convertible to int32's.
```
antocuni, line 70: this seems a duplicate?

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### __print__(obj) { data-toc-label='print()' }
+
+:   The print function is currently not variadic, in the sense that it only accepts a single argument. The built-in types are special-cased, and SPy can always print blue objects by pre-computing their string representation
+
+### __range__(stop) { data-toc-label='range()' }
+<h3> <b>range</b>(start, stop, step) { data-toc-label='' }</h3> <!-- An HTML label to hide this in the TOC -->
+
+Creates an iterable set of indices between `start` and `stop`, jumping over `step` indices between each.
```
antocuni: add a link to `stdlib/_range.spy`

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+Number types beyond int and float are in active development; some of the math functions below are also in active development.
+
+:   divmod(), bin(), hex(), oct(), pow(), round()
+
+### Function Types, Introspection and Metaprogramming
+
+The internals of SPy are significantly different from CPython; as such, the road to (and need for) some of these built-ins is less straightfoward. Some are also waiting on internal details to solidify prior to implementation.
+
+:   breakpoint(), callable(), classmethod(), compile(), delattr(), dir(), eval(), exec(), globals(), hasattr(), help(), id(), isinstance(), issubclass(), locals(), property(), staticmethod(), super(), type(), vars(). \_\_import\_\_()
```
antocuni: `breakpoint` is implemented and starts `spdb` :)

antocuni: `type()` is implemented

antocuni: - `staticmethod` is implemented
- `classmethod` and `property` are technically implemented, but so far they are used only by interp-level code, and they are not really usable from app-level code

antocuni: `dir()` is implemented

docs/src/api_reference/python_builtins.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Type Conversion
+
+Many of these types are not implemented yet; others are in active development.
+
+:   ascii(), bool(), bytearray(), bytes(), chr(), complex(), format(), frozenset(), memoryview(), object(), ord() set(), tuple()
```
antocuni: `object` is "implemented" in the sense that exists as a builtin and can be used as a type (you cannot do much with variables typed `object`, but you can downcast them):
```python
def foo(a: object):
    x = a + 1 # TypeError
    b: i32 = a  # implicit downcast, might raise
    y = b + 1 # works
```

antocuni: `tuple` is implemented now, see `stdlib/_tuple.spy`

docs/src/api_reference/python_builtins.md
JeffersGlass: Ah, should be `min`, was a copy-paste error.

PR #405, opened 2026-02-26
C: sequence expressions for deterministic eval order

Jemeljanov:
C does not guarantee operand evaluation order, but SPy requires left-to-right semantics.
This PR adds a backend-local expression sequencer that rewrites typed C-backend AST statements so evaluation order is explicit.

The pass introduces temporaries only when required for correctness:
- materialize effectful parts when later parts are ordering-sensitive;
- snapshot unstable reads only when later effects could change the observed value;
- hoist local assignment expressions when that preserves order without creating a tmp.

It also performs control-flow rewrites needed to preserve lazy semantics:
- short-circuit expressions with RHS preludes are lowered into explicit `if` flow;
- `while`/`assert` cases with preludes are rewritten so preludes execute at the correct time.

Added module-level and function docs in `expr_sequencer.py`.

### DIFF
--- a/spy/backend/c/cwriter.py
+++ b/spy/backend/c/cwriter.py
@@ -98,13 +98,8 @@ class CFuncWriter:
         """
         Emit a #line directive, but only if it's needed.
         """
-        desired_spy = loc.line_start
         # line numbers corresponding to the last emitted #line
         last_spy, last_c = self.last_emitted_linenos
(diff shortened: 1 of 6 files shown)


antocuni:
Closing this as it has the problem has been fixed by #478 .
@Jemeljanov sorry for closing without a proper review, but as you surely saw this was not trivial and required serious thinking. The approach adopted by #478 is more general and more easily testable.

Issue #353, opened 2026-01-07
Add support for argv and exit code


antocuni:
I would like to be able to receive `argv` from command line.
CPython relies on `sys.argv`, but for SPy I think that we should rely on the signature of the `main` function. Something like this:

```python
def main() -> None:
    "this is what we have now: you cannot access argv, and the exit code is always 0"

def main() -> int:
    "you can return a value which will be used as exit code"

def main(argv: list[str]) -> None:
    "you can use argv"

def main(argv: list[str]) -> int:
    "you can use argv AND return an exit code"
```

This requires support in both the interpreter and the C backend:
  - the interpreter needs to handle the various cases here:
https://github.com/spylang/spy/blob/da4fba86762d722f3fb9c708db8696d80e1c40cf/spy/cli/_runners.py#L78-L80

  - the C backend needs to handle the various cases here:
https://github.com/spylang/spy/blob/da4fba86762d722f3fb9c708db8696d80e1c40cf/spy/backend/c/cmodwriter.py#L153-L161

kanin-kearpimy:
Sub-tasks

Interpreter:
- [x] return type
- [x] args: list[str]

Compiler:
- [x] return type
- [x] args: list[str]

PR #419, opened 2026-03-14
353 add support for argv and exit code

kanin-kearpimy:
## Issue
- https://github.com/spylang/spy/issues/353

## Description
This PR enable **exit code** in main() for both interp-level and compile-level. Also, it enable `argv` parameters in interp-level `main(argv: list[str])` function. The `argv` at compile-level (i.g. C) is following up. Exit code is constraint to `None` and `i32` type only.


Viriathus1:
@kanin-kearpimy I was getting that cli error as well it's a Mac build error which I suspect is related to the Mac's llvm version but I didn't look too deeply into it at the time.

PR #421, opened 2026-03-16
Compound assignment

### LINKED ISSUE #177: Implement += &co. for complex expressions
```
def main() -> void:
    a.b += 1
```

```
❯ spy /tmp/x.spy 
ParseError: not implemented yet: assign to complex expressions
   --> /tmp/x.spy:2:5
  2 |     a.b += 1
    |     |_| this is not supported
```

Viriathus1:
Closes #177 

This adds compound assignment support for attributes and subscripts.


spy/tests/compiler/test_basic.py
```diff
@@ -887,6 +887,76 @@ def foo(x: i32) -> i32:
         """)
         assert mod.foo(10) == ((10 + 1) * 2) - 3
 
+    def test_aug_assign_subscript(self):
+        mod = self.compile("""
+        from _list import list
```
antocuni: I think it's better to use `__spy__.interp_list` here.
I'd like to avoid `test_basic` to depend on `_list.spy`, because that needs "the whole spy" to work, and thus it defeats a bit the utility of having a "test_basic"

antocuni (approved):
thank you!
The overall approach works, but I'm not 100% sure that doing the work directly in the parser is the right thing to do.

See e.g. this comment:
https://github.com/spylang/spy/pull/435#issuecomment-4149860474

One possible "more clean" solution is to have proper "AugSetItem" and "AugSetAttr" nodes, and the we do desugaring in astframe as we do for `for` loops and exising `AugAssign`. What do you think?

Moreover, we need to make sure that the index is evaluated only once.
E.g.:

```
def foo() -> int:
    print('foo')
    return 0

x[foo()] += 1
```

needs to print `foo` only once, but I think that with your change it would print it twice

Viriathus1:
Thanks for the review, the approach just seemed like the easiest solution but I do see that it breaks away from desugaring in the AST frame and the other issues you mentioned. I'll work on the cleaner solution when I can.

Viriathus1:
Hi @antocuni this requires another review. I've pushed the desugaring into the astframe, let me know if anything caught your eye.

spy/vm/astframe.py
```diff
@@ -740,6 +740,78 @@ def _desugar_AugAssign(self, node: ast.AugAssign) -> ast.Assign:
             ),
         )
 
+    def _desugar_AugSetAttr(self, node: ast.AugSetAttr) -> list[ast.Stmt]:
```
antocuni: I think it would be nice to have a comment which explain what we have before and after the desugaring, similar to what we do for `desugar_for`

spy/vm/astframe.py
```diff
@@ -740,6 +740,78 @@ def _desugar_AugAssign(self, node: ast.AugAssign) -> ast.Assign:
...
+                        loc=node.loc,
+                        value=ast.Name(loc=node.target.loc, id=temp_name),
+                        attr=node.attr,
+                    ),
+                    right=node.value,
+                ),
+            ),
+        ]
+
+    def _desugar_AugSetItem(self, node: ast.AugSetItem) -> list[ast.Stmt]:
```
antocuni: same as above

spy/ast.py
```diff
@@ -751,13 +751,41 @@ class SetAttr(Stmt):
...
+@astnode
+class AugSetAttr(Stmt):
+    target: Expr
+    attr: StrConst
+    op: str
+    value: Expr
+
+    def target_name(self) -> str:
+        suffix = f"{self.loc.line_start}_{self.loc.col_start}_{self.loc.line_end}_{self.loc.col_end}"
+        return f"_$augattr_obj_{suffix}"
```
antocuni: probably we should kill this, see my longer comment in the review

spy/ast.py
```diff
@@ -751,13 +751,41 @@ class SetAttr(Stmt):
...
+    op: str
+    value: Expr
+
+    def target_name(self) -> str:
+        suffix = f"{self.loc.line_start}_{self.loc.col_start}_{self.loc.line_end}_{self.loc.col_end}"
+        return f"_$augitem_obj_{suffix}"
+
+    def arg_name(self, index: int) -> str:
+        suffix = f"{self.loc.line_start}_{self.loc.col_start}_{self.loc.line_end}_{self.loc.col_end}"
+        return f"_$augitem_arg{index}_{suffix}"
```
antocuni: same

antocuni (changes requested):
Hi @Viriathus1 and sorry for the late review (the last month has been crazy).
I like this approach much better than the earlier one 🎉.

The part which I'm not sure about is using `AugSetAttr.target_name()` to get a temporary variable.
We already do something similar for `for` loops: in that case, the parser assigns an unique `seq` ID, which is then used to generate the temp variable for the iterator.
I think we could use the same approach for Aug assign, what do you think?

Another small note about desugaring: for other cases we do desugaring on the first execution and then we cache the result in e.g. `ASTFrame._desugared_fors`, and then the doppler can just reuse it without having to re-desugar.
Is there any particular reason why you didn't do that?

antocuni:
@Viriathus1 have a look at PR #649.
Now that we have the `astcompile` pass, this PR should become easier: you can do the desugaring inside `astcompile` (by using a tmp variable, probably), and then all the rest should work out of the box.

antocuni (approved):
Thank you @Viriathus1 .
I like this version much better :).

I noticed that you added a test in `test_cli.py` to check the output of astcompile: that's good but test_cli was not really the right place.

I realized that we should have proper astcompile tests, so I added them in #651 , rebased your PR on top of it and added "proper" tests for AugSetItem and AugSetAttr, I hope you don't mind.

Will merge it as soon as it's green! :)

Issue #426, opened 2026-03-18
Make it possible to embed code in the playground URL


antocuni:
[Compiler Explorer](https://godbolt.org/) makes it possible to generate "share" URLs which embeds the snippet of code directly in the URL. E.g. the following:

[...]

PR #427, opened 2026-03-19
Playground snippet

Viriathus1:
A very basic attempt at resolving #426.


antocuni:
> A very basic attempt at resolving #426.

wow, that was super quick, thank you! 😍

The functionality seems to work perfectly.
I think that the `Share` button is not super visible though, it took a while for me to find.

Maybe it would be easier to find if we put it e.g. on the top toolbar on the right?

Issue #357, opened 2026-01-07
Create a `file` object in the stdlib


antocuni:
Create `stdlib/_io.spy`; it should have a `file` objects which uses `os.open/read/write` to do I/O

Issue #354, opened 2026-01-07
Add preliminary support for file I/O


antocuni:
It would be nice to be able to read/write files.

It's not clear what is the best long term strategy for that, because we need to think about buffering, encodings, unicode vs bytes, etc.

However, I think that for now _any_ way to read files would be good, because it would unlock more usage of SPy.

I think that the best way to do is the following:

1. implement the `open`, `read` and `write` syscalls in the `posix` builtin module:
https://github.com/spylang/spy/blob/da4fba86762d722f3fb9c708db8696d80e1c40cf/spy/vm/modules/posix.py#L1-L43

2. create a new file `stdlib/_io.spy`, which exposes a `file` object which uses the syscalls above to implement functionalities (with buffering). I think that for now we could support only "utf-8" encoding, because that's what our `str` objects supports natively.

PR #434, opened 2026-03-26
io: implement posix.open/read/write.close + _io.FileIO + open

### LINKED ISSUE #355: Implement posix.open and posix.read
We should be able to use low-level file descriptors to open a file and read bytes out of it.

The big problem is that we currently don't have a `bytes` object, so our `posix.read` must return a `str` object. I think it's fine for now, but we probably need to think better about it.

### LINKED ISSUE #356: Implement posix.write
Self-explanatory, depends on #355

paugier:
- Fixes #355
- Fixes #356
- ~~Fixes~~ related to #357 
- ~~Fixes~~ related to #354

I was helped by Claude, which in particular wrote the C code! It seems reasonable but I'm quite unable to judge if it is really correct.

The most important tests (`test_open_read` and `test_open_write`) raises an error for the C backend.

```
>               raise SPyError.simple(etype, message, "", loc)
E               spy.errors.SPyError: Traceback (most recent call last):
E               
E               ValueError: No such file or directory
E                 | /home/pierre/dev/spy/spy/libspy/include/spy/posix.h:51
E                 |         spy_panic("ValueError", strerror(errno), __FILE__, __LINE__);
E                 |  |__________________________________________________________________|
```

Claude told me that it is due to the WASI sandbox and that the test can be skipped with

```
@skip_backends("C", reason="WASI sandbox cannot access host filesystem paths")
```

Nevertheless a good news is that this works fine:

```
cd examples
spy build io_low_level.spy
./build/io_low_level
```

### DIFF
--- a/docs/src/api_reference/python_builtins.md
+++ b/docs/src/api_reference/python_builtins.md
@@ -73,6 +73,10 @@ The following built-in functions work similarly to their equivalents in CPython;
 
 :   `object` is implemented as a type, and can be used as a parameter or return type. "Plain" objects (i.e. `x = object()`) are not supported.
 
+### __open__(file: str, mode: str) { data-toc-label='open()' }
+
+:    Open file and return a `_io.FileIO`.  Raise OSError upon failure.
+
 ### __print__(obj) { data-toc-label='print()' }
 
 :   The print function is currently not variadic, in the sense that it only accepts a single argument. The built-in types are special-cased, and SPy can always print blue objects by pre-computing their string representation
@@ -151,4 +155,4 @@ Many of these types are not implemented yet; others are in active development.
 
 The I/O story is currently a high priority and is in active development.
 
-:   input(), open()
+:   input()
--- a/docs/src/api_reference/python_builtins.md
+++ b/docs/src/api_reference/python_builtins.md
@@ -7,7 +7,7 @@ h3 {
 }
 </style>
 
-## Implemented CPython-Like Built-ins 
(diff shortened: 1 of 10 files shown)


antocuni:
> I see that it would be a very nice demo that SPy can be useful for low level to implement in SPy something like 2. (implement a python-like I/O tower in C). And I see that we need (i) memcpy / memmove, (ii) bytes and str/bytes interop and (iii) more pointers.
> 
> To start to use SPy for things like benchmarks, it seems to me that the IO can be quite limited. Just FileIO might even be sufficient?

I think that would be very bad, especially for benchmarks: unbuffered IO is *extremely* slow and would make all the benchmark looking bad.
I just merged #447 which adds a `file` object based on libc, which is good enough for now.

Sorry for having made you unnecessary work. I even considered merging this PR anyway and keep it for the future, but I don't think it's a good idea because we need a much deeper thinking about it: in particular, on CPython `os.read` *allocates* and return a string at every call. We surely don't want this in our performance oriented code: for `FileIO`, we probably want a different API which lets us to read into a preallocated buffer.

I'm going to leave this PR open, I might decide to "extract" the implementations of `os.read` and `os.write` (which seems good), but without rush.

antocuni:
@paugier now that we have `stdlib/_file.spy`, do you think we can close this PR?
I see it includes implementation of `posix.read` and `posix.write`, but they are probably outdated since they were written when we didn't have `bytes`

Issue #417, opened 2026-03-14
Syntaxic sugar for @blue.generic


paugier:
I feel that it would be nice to be able to write

```
def add[T, x](y: T) -> T:
    return x + y

add_4 = add[i32, 4]
```

for 

```
@blue
def make_adder(T, x):
    def adder(y: T) -> T:
        return x + y
    return adder

add_4 = make_adder(i32, 4)
```

and of course

```
@struct
class ArrayData[DTYPE]:
    length: i32
    capacity: i32
    items: gc_ptr[DTYPE]
```

for

```
@blue.generic
def ArrayData(DTYPE):
    @struct
    class _ArrayData:
        length: i32
        capacity: i32
        items: gc_ptr[DTYPE]

    return _ArrayData
```

I see that such syntaxic sugar would simplify (at least shorten + improve readability) a lot of SPy examples.

And I think it would be possible to make https://github.com/spylang/spy/blob/main/playground/point.spy generic just with this simple diff:

```
 @struct
-class Point:
-    x: f64
-    y: f64
+class Point[DTYPE]:
+    x: DTYPE
+    y: DTYPE
 
 # eventually structs will get an automatic ctor, like dataclasses
-def new_point(x: f64, y: f64) -> gc_ptr[Point]:
+def new_point[DTYPE](x: DTYPE, y: DTYPE) -> gc_ptr[Point]:
     p = gc_alloc[Point](1) # allocate 1 Point
     p.x = x
     p.y = y
     return p
 
-def squared_distance(p0: gc_ptr[Point], p1: gc_ptr[Point]) -> f64:
+def squared_distance[DTYPE](p0: gc_ptr[Point], p1: gc_ptr[Point]) -> DTYPE:
     dx = p0.x - p1.x
     dy = p0.y - p1.y
     return dx*dx + dy*dy
 
 def main() -> None:
-    p0 = new_point(1, 2)
-    p1 = new_point(3, 4)
+    p0 = new_point[f64](1, 2)
+    p1 = new_point[f64](3, 4)
```

But is it clever to introduce this? And if yes when?

PR #435, opened 2026-03-28
parser: simple implementation of PEP 695 for functions

paugier:
Related to #417

With it, this works

```
def my_func[T](a0: T, a1: T) -> T:
    return a0 + a1

def main() -> None:
    print(my_func[i32](2, 2))
```

This is nice, but I don't think it is the proper long term solution. For example, it's bad that `spy parse` gives something quite different from the code.


antocuni:
closing this is favor of #437

PR #437, opened 2026-03-29
Type parameter syntax (PEP 695) with generic_args in FuncDef

paugier:
Related to #417

At least, this seems to work but I had to ask Claude, in particular for the ScopeAnalyzer/SymTable part.

Advice would be welcome!

### DIFF
--- a/spy/analyze/scope.py
+++ b/spy/analyze/scope.py
@@ -83,8 +83,11 @@ class ScopeAnalyzer:
     def by_module(self) -> SymTable:
         return self.mod_scope
 
-    def by_funcdef(self, funcdef: ast.FuncDef | ast.GenericFuncDef) -> SymTable:
+    def by_funcdef(self, funcdef: ast.FuncDef) -> SymTable:
         return self.inner_scopes[funcdef]
+
+    def by_generic_funcdef(self, gfuncdef: ast.GenericFuncDef) -> SymTable:
+        return self.inner_scopes[gfuncdef]
 
     def by_classdef(self, classdef: ast.ClassDef) -> SymTable:
         return self.inner_scopes[classdef]
@@ -502,7 +505,7 @@ class ScopeAnalyzer:
     def flatten_GenericFuncDef(self, gfuncdef: ast.GenericFuncDef) -> None:
         for arg in gfuncdef.args:
             self.flatten(arg)
-        inner_scope = self.by_funcdef(gfuncdef)
+        inner_scope = self.by_generic_funcdef(gfuncdef)
         self.push_scope(inner_scope)
         self.flatten_FuncDef(gfuncdef.inner)
         self.pop_scope()
--- a/spy/tests/test_scope.py
+++ b/spy/tests/test_scope.py
@@ -536,8 +536,8 @@ class TestScopeAnalyzer:
             v: T = 1
         """)
 
-        generic_funcdef = self.mod.get_generic_funcdef("foo")
-        scope = scopes.by_funcdef(generic_funcdef)
+        gfuncdef = self.mod.get_generic_funcdef("foo")
+        scope = scopes.by_generic_funcdef(gfuncdef)
         assert scope.name == "test::foo"
         assert scope.color == "blue"
         assert scope._symbols == {
@@ -546,7 +546,7 @@ class TestScopeAnalyzer:
             "@return": MatchSymbol("@return", "var", "auto"),
         }
 
-        inner_scope = scopes.by_funcdef(generic_funcdef.inner)
+        inner_scope = scopes.by_funcdef(gfuncdef.inner)
         assert inner_scope.name == f"test::foo::__impl"
         assert inner_scope.color == "red"
         assert inner_scope._symbols == {


spy/parser.py
```diff
@@ -183,6 +183,42 @@ def from_py_stmt_FunctionDef(
...
+        # PEP695 type params, def foo[T, U](...)
+        generic_args: Optional[list[spy.ast.FuncArg]] = None
+        if py_funcdef.type_params:
+            if decorators:
+                # mixing [T] syntax with user-defined decorators is confusing; reject it
+                self.error(
+                    "cannot use type parameters with decorators",
+                    "remove the decorators or the type parameters",
+                    py_funcdef.loc,
+                )
```
antocuni, line 198: This is an interesting problem. Should the decorator be applied to the inner or outer function?

My first gut feeling is that it should be applied to the INNER function. Something like this:
```
@deco
def foo[T](x: T) -> T:
    ...

===>

@blue.generic
def foo(T):
    @deco
    def impl(x: T) -> T:
        ...
```

but then indeed, it's a bit surprising because `deco` is not called at "import time" but at "redshift time", and might also be called multiple times.

spy/parser.py
```diff
@@ -183,6 +183,42 @@ def from_py_stmt_FunctionDef(
...
+                        "only plain TypeVar type parameters are supported",
+                        "ParamSpec and TypeVarTuple are not supported yet",
+                        tp.loc,
+                    )
+                if tp.bound is not None:
+                    self.error(
+                        "bounded TypeVars are not supported yet",
+                        "remove the bound",
+                        tp.loc,
+                    )
```
antocuni: another interesting problem. I think that the "bound parameter" syntax of Python is basically wrong because it reuses the syntax for "type annotation" but gives it a different meaning.

I'm tempted to declare a difference on CPython and say that the `T: expr` syntax in square brackets is a standard type annotation, something like this:

```
def foo[T: type, n: int](x: T) -> T:
    ...

==>

@blue.generic
def foo(T: type, n: int):
    ...
```

But at the same time, it's bad to introduce an unnecessary incompatibility w.r.t Python.

spy/vm/function.py
```diff
@@ -355,6 +355,10 @@ def __init__(
         *,
         locals_types_w: Optional[dict[str, W_Type]] = None,
     ) -> None:
+        assert not funcdef.has_generic_args, (
+            "W_ASTFunc must wrap a concrete funcdef; "
+            "PEP695 generic_args should have been desugared by exec_stmt_FuncDef"
```
antocuni:
```suggestion
            "generic_args should have been desugared by exec_stmt_FuncDef"
```

spy/tests/test_parser.py
```diff
@@ -344,6 +353,48 @@ def foo() -> i32:
         """
         self.assert_dump(funcdef, expected)
 
+    def test_pep695_FuncDef(self):
```
antocuni:
```suggestion
    def test_FuncDef_generic_args(self):
```

spy/tests/compiler/test_basic.py
```diff
@@ -942,6 +942,20 @@ def bar() -> str:
         assert mod.foo() == 3
         assert mod.bar() == "hello world"
 
+    def test_pep695_generic(self):
```
antocuni:
```suggestion
    def test_generic_args(self):
```

spy/analyze/scope.py
```diff
@@ -280,16 +280,21 @@ def declare_FuncDef(self, funcdef: ast.FuncDef) -> None:
         self.define_name(funcdef.name, "const", "funcdef", protoloc, protoloc)
         # add function arguments to the "inner" scope
         scope_color = funcdef.color
+
+        inner_scope = self.new_SymTable(funcdef.name, scope_color, "function")
+        self.push_scope(inner_scope)
+        self.inner_scopes[funcdef] = inner_scope
+        # PEP695 generic args are always blue consts in the function scope
```
antocuni:
```suggestion
        # generic args are always blue consts in the function scope
```

antocuni: I prefer not to mention PEP695 in the code because we are not really implementing PEP 695, just (ab)using its syntax

spy/analyze/scope.py
```diff
@@ -280,16 +280,21 @@ def declare_FuncDef(self, funcdef: ast.FuncDef) -> None:
         self.define_name(funcdef.name, "const", "funcdef", protoloc, protoloc)
         # add function arguments to the "inner" scope
         scope_color = funcdef.color
+
```
antocuni: the code in `scope.py` seems correct, but I'd like a test in `test_scope.py`

spy/vm/astframe.py
```diff
@@ -352,6 +355,107 @@ def exec_stmt_FuncDef(self, funcdef: ast.FuncDef) -> None:
...
+
+            @blue.generic
+            def my_func(T):
+                def _my_func(a0: T, a1: T) -> T:
+                    return a0 + a1
+                return _my_func
+        """
+        assert funcdef.generic_args is not None
+        loc = funcdef.loc
+        inner_name = f"_{funcdef.name}"
```
antocuni: I suggest to use something short like `@impl` (hardcoded) here, else the FQN might become very long if the funcname is long.

The other advantage is that `@impl` is not a valid identifier so we are also sure not to get any clash

spy/vm/astframe.py
```diff
@@ -352,6 +355,107 @@ def exec_stmt_FuncDef(self, funcdef: ast.FuncDef) -> None:
...
+
+        # Build the inner (concrete red) funcdef — same body, no generic_args
+        inner_funcdef = funcdef.replace(
+            generic_args=None,
+            kind="plain",
+            color="red",
+            name=inner_name,
+            decorators=[],
+        )
+        inner_funcdef.symtable = funcdef.symtable
```
antocuni: uh, I'm not 100% convinced that this is right, although it might work in practice.
Take this:
```
def foo[T](x: T):
    print(x)
    print(T)
```

from the point of view of the inner function, `x` is a local var and needs to exist in the inner symtable, while `T` does not exist in the inner symtable and should be found as an `NameOuterDirect` in the enclosing scope.

I probably the right solution is to modify scope.py to create two nested symtables for this case (and this reinforces the need for tests in test_scope.py)

spy/vm/astframe.py
```diff
@@ -352,6 +355,107 @@ def exec_stmt_FuncDef(self, funcdef: ast.FuncDef) -> None:
...
+            return_type=ast.Auto(loc),
+            docstring=funcdef.docstring,
+            body=[inner_funcdef, return_inner],
+            decorators=[],
+            symtable=outer_st,
+        )
+
+        self.exec_stmt_FuncDef(outer_funcdef)
+
+    def _make_outer_generic_symtable(self, funcdef: ast.FuncDef) -> SymTable:
```
antocuni: same as above. I think it's much better if `scope.py` generates the correct symtable for both instead of "faking" it here

spy/vm/astframe.py
paugier: > to create two nested symtables for this case

I think I understand but where should be stored the inner symtable?

Just to be sure that I understood, would it make sense to have a test like this (just the principle, I see some evident changes so that both cases work):

```
    def test_generic_args(self):
        scopes = self.analyze("""
        @blue
        def outer(T):
            def inner(x: T) -> None:
                pass
            return inner

        def outer_generic[T](x: T):
            pass
        """)

        for fname in ("outer",): #, "outer_generic"):
            if fname.endswith("generic"):
                name_inner = "@impl"
            else:
                name_inner = "inner"

            funcdef = self.mod.get_funcdef(fname)
            scope = scopes.by_funcdef(funcdef)
            assert scope.name == "test::" + fname
            assert scope.color == "blue"
            assert scope._symbols == {
                "T": MatchSymbol("T", "const", "blue-param"),
                name_inner: MatchSymbol(name_inner, "const", "funcdef"),
                "@return": MatchSymbol("@return", "var", "auto"),
            }

            inner_funcdef = funcdef.body[0]
            assert isinstance(inner_funcdef, ast.FuncDef)
            inner_scope = scopes.by_funcdef(inner_funcdef)
            assert inner_scope.name == "test::outer::inner"
            assert inner_scope.color == "red"
            assert inner_scope._symbols == {
                "x": MatchSymbol("x", "var", "red-param"),
                "@return": MatchSymbol("@return", "var", "auto"),
            }
```

In the `outer_generic` case, we cannot do `inner_scope = scopes.by_funcdef(inner_funcdef)` since there is no inner_funcdef...

antocuni: > I think I understand but where should be stored the inner symtable?

currently `ScopeAnalyzer.inner_scopes` is typed this way:
```
    inner_scopes: dict[ast.FuncDef | ast.ClassDef, SymTable]
```

What we want is to have a way to store a symtable for the "generic funcdef" and a symtable for the "specialized funcdef". Maybe something like this?

```
@dataclass(frozen=True)
class GenericFuncdef:
    "just a placeholder to use inside inner_scopes"
    funcdef: ast.FuncDef

    def __post_init__(self):
        assert self.funcdef.has_generic_args


class ScopeAnalyzer:
    inner_scopes: dict[ast.FuncDef | GenericFuncDef | ast.ClassDef, SymTable]

    def by_funcdef(self, funcdef: ast.FuncDef) -> SymTable:
        return self.inner_scopes[funcdef]

    def by_generic_funcdef(self, funcdef: ast.FuncDef):
        return self.inner_scopes[GenericFuncDef(funcdef)]

    # etc.
```

So basically, `inner_scopes[funcdef]` will hold the symtable for the "inner" function, and `inner_scopes[Generic(funcdef)]` will hold the symtable for the "outer" function.

But then you also need to find a way to store both symtables in the `FuncDef`, so probably you need this:

```
@astnode
class FuncDef(Stmt):
    ...
    symtable: Any = field(repr=False, default=None)
    generic_symtable: Any = field(repr=False, default=None)
```

This SHOULD work, although it seems a bit hacky.
We could try to revisit again the idea of `ast.GenericFuncDef`, but in a lighter way. Something like this:

```
# ast.py

@astnode
class GenericFuncDef(Stmt):
    args: list[FuncArg]
    inner: FuncDef
    symtable: ...
```

In this way:
1. the parser needs to do part of the work which is currently done by astframe.py
2. we keep the code duplication between GenericFuncDef and FuncDef minimal, because we "embed" the inner funcdef
3. we have a "proper" place where to store the symtable, and a proper key for `inner_scopes`
4. `astframe.py` STILL needs to do some desugaring to turn a `GenericFuncDef` into a `@blue.generic` funcdef

Sorry for the continuous back and forth between different ideas, but as you can see the problem is a bit tricky and while there are multiple possible solutions, there seem to be a tradeoff between "theoretically clean" and "easy". Finding the best sweet spot is not straighforward but I think we are slowly converging :).

paugier: > I think we are slowly converging

I agree and I like the idea of `GenericFuncDef` with an inner FuncDef. I'm going to adapt the PR soon.

spy/analyze/scope.py
paugier: I added a test in test_scope.py.

spy/parser.py
paugier: Yes, it's an issue. But we will need that for struct so I think it should indeed be like that.

spy/vm/astframe.py
paugier: `@impl` is not a valid identifier in C so I used `__impl` to get the CI green. Is there a better solution?

spy/vm/astframe.py
paugier: I think I followed your last idea. It seems to me that it is indeed cleaner.

spy/vm/astframe.py
paugier: Yes, no need for this function now.

spy/parser.py
antocuni: `@struct` (like `@blue`) is special and it's not an actual decorator, it's special-cased by the parser. So that would not be an issue, but yes, probably applying it to the inner function makes sense.
Anyway, happy to defer this decision to the future

spy/vm/astframe.py
antocuni: indeed, using `@impl` would require to special case it in the C backend. It's doable but probably not worth it, `__impl` is good enough👍

PR #448, opened 2026-04-02
Generic types (syntactic sugar for blue.generic)

paugier:
Related to #417 and direct continuation of #437.

At this point, this PR is mostly to show that it's possible and to be able to discuss about Generic types.

Now that we have `def func[T](x: T) -> T`, it really makes sense to have something equivalent for classes.

```python
@struct
class ArrayData[DTYPE]:
    length: i32
    capacity: i32
    items: gc_ptr[DTYPE]
```

would just be syntactic sugar for

```python
@blue.generic
def ArrayData(DTYPE):
    @struct
    class Self:
        length: i32
        capacity: i32
        items: gc_ptr[DTYPE]

    return Self
```

The current method to write generic types is by using a blue.generic function (it is widely used in the standard lib).

I understand (as discussed in #417) that there are unresolved issues with this with respect to wanted behavior of generic types, in particular regarding `isinstance` and `issubclass`. However,

- these issues are only weakly related to this syntactic sugar
- SPy has currently no `isinstance` and `issubclass`

Therefore, it seems to me that these issues will be treated later and that we can introduce the generic type syntactic sugar before.

We see the potential of this syntactic sugar with the simplification of playground/myarray.spy, which becomes much more readable and understandable for Python developers.

Remark: I had to use `Self` for the name of the concrete type since this concrete type has to be used to be able to implement `__new__` (currently we cannot write `ndarray[DTYPE]` in the definition of `ndarray[DTYPE]` so we need `Self`)

Remark 2 (detail): it could help to also support `type ArrayData_c = ArrayData[DTYPE]` at the class level (as in Python).

Remark 3: I had to allow isolated module level strings for the example, but this has nothing to do with the subject of this PR and it should be done correctly in another PR.

### DIFF
--- a/examples/myarray.spy
+++ b/examples/myarray.spy
@@ -2,83 +2,91 @@
 This is one of the most advanced examples of SPy so far.
 
 It implements the basics of a generic array type
+
+Note that struct are passed by values (i.e. copied).
 """
 
-from operator import OpSpec, MetaArg
 from unsafe import gc_alloc, gc_ptr
 
+
+@struct
+class ArrayData[DTYPE]:
+    length: i32
+    capacity: i32
+    items: gc_ptr[DTYPE]
+
+
+"""
+This generic class is just syntactic sugar for
 
 @blue.generic
 def ArrayData(DTYPE):
     @struct
-    class _ArrayData:
+    class Self:
         length: i32
         capacity: i32
         items: gc_ptr[DTYPE]
 
-    return _ArrayData
+    return Self
+
+Note that the name of the concrete type is always "Self".
+"""
 
 
-@blue
-def ndarray1(DTYPE):
-    """
-    Currently generics are implemented using blue functions taking a
-    type. Eventually we will support a syntax like `ndarray[i32]`, which will
-    be equivalent to this.
-    """
+@struct
+class Array1d[DTYPE]:
+    # This struct only contains a low level ("__ll__") pointer towards the data leaving on the heap (allocated)
+    __ll__: gc_ptr[ArrayData[DTYPE]]
 
-    @struct
-    class ndarray:
-        __ll__: gc_ptr[ArrayData[DTYPE]]
+    # note: here, Self corresponds to the concrete type Array1d[DTYPE].
 
-        def __new__(length: i32) -> ndarray:
-            data = gc_alloc[ArrayData[DTYPE]](1)
-            data.length = length
-            data.capacity = length
-            data.items = gc_alloc[DTYPE](length)
+    def __new__(length: i32) -> Self:
+        # a pointer towards the array data
+        ll = gc_alloc[ArrayData[DTYPE]](1)
+        ll.length = length
+        ll.capacity = length
+        ll.items = gc_alloc[DTYPE](length)
+        i = 0
+        while i < length:
+            ll.items[i] = 0
+            i = i + 1
+        return Self.__make__(ll)
+
+    def append(self, value: DTYPE) -> None:
+        ll = self.__ll__
+        if ll.length >= ll.capacity:
+            # resize needed - double the capacity
+            new_capacity = ll.capacity * 2
+            if new_capacity == 0:
+                new_capacity = 1
+            new_items = gc_alloc[DTYPE](new_capacity)
+            # copy existing items
             i = 0
-            while i < length:
-                data.items[i] = 0
+            while i < ll.length:
+                new_items[i] = ll.items[i]
                 i = i + 1
-            return ndarray.__make__(data)
+            ll.items = new_items
+            ll.capacity = new_capacity
 
-        def append(self, value: DTYPE) -> None:
-            ll = self.__ll__
-            if ll.length >= ll.capacity:
-                # resize needed - double the capacity
-                new_capacity = ll.capacity * 2
-                if new_capacity == 0:
-                    new_capacity = 1
-                new_items = gc_alloc[DTYPE](new_capacity)
-                # copy existing items
-                i = 0
-                while i < ll.length:
-                    new_items[i] = ll.items[i]
-                    i = i + 1
-                ll.items = new_items
-                ll.capacity = new_capacity
+        ll.items[ll.length] = value
+        ll.length = ll.length + 1
 
-            ll.items[ll.length] = value
-            ll.length = ll.length + 1
+    def __getitem__(self, i: i32) -> DTYPE:
+        ll = self.__ll__
+        if i >= ll.length:
+            raise IndexError
+        return ll.items[i]
 
-        def __getitem__(self, i: i32) -> DTYPE:
-            ll = self.__ll__
-            if i >= ll.length:
-                raise IndexError
-            return ll.items[i]
-
-        def __setitem__(self, i: i32, v: DTYPE) -> None:
-            ll = self.__ll__
-            if i >= ll.length:
-                raise IndexError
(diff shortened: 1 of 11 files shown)


paugier:
I rebased to be able to use new things recently added to the repo with this branch.

paugier:
Just to save this idea somewhere, I add here (not totally appropriate since it is not directly related to this PR) that in order that this syntax be useful to implement real things (in particular things currently in stdlib), it would be nice to support something like:

```python
# ArrayData_generic and func_generic already defined above

@struct
class ndarray[DTYPE, NDIM]:

    type ArrayData = ArrayData_generic[DTYPE, NDIM]
    type func = func_generic[ArrayData]

    @blue
    def __blue_checks__():
        if NDIM>2:
            raise StaticError("NDIM>2 not supported")

    @blue
    def __blue_replace_type__():
        # nonsense but just to get the idea
        if NDIM == 0:
            return f64

    __ll__: gc_ptr[ArrayData]
```

syntactic sugar for

```python
@blue.generic
def ndarray(DTYPE, NDIM):
    ArrayData = ArrayData_generic[DTYPE, NDIM]
    func = func_generic[ArrayData]

    @blue
    def __blue_checks__():
        if NDIM>2:
            raise StaticError("NDIM>2 not supported")

    __blue_checks__()

    @blue
    def __blue_replace_type__():
        # nonsense but just to get the idea
        if NDIM == 0:
            return f64

    returned_type = __blue_replace_type__()
    if returned_type is not None:
        return returned_type

    @struct
    class ndarray:
        __ll__: gc_ptr[ArrayData]
```

paugier:
@antocuni can this PR be considered? It would be convenient for me to be able to use this syntax even when using other branches, to avoid conflicts and also to build upon these changes, in particular with `type ArrayData = ArrayData_generic[DTYPE, NDIM]`.

These changes seem quite reasonable and symmetrical to #437 (the equivalent for functions).

antocuni:
@paugier yes sorry, I want to consider it but I didn't have time to look at it yet.
I'm a bit overwhelmed by the amount of open PRs/issues and I decided to take "focus weeks" to concentrate on new developments (what I'm doing now).

I promise/hope to get back to it soon!

paugier:
> I decided to take "focus weeks" to concentrate on new developments (what I'm doing now).

That's very understandable and useful! Thanks for the news.

examples/myarray.spy
```diff
@@ -2,83 +2,91 @@
...
-    class _ArrayData:
+    class Self:
         length: i32
         capacity: i32
         items: gc_ptr[DTYPE]
 
-    return _ArrayData
+    return Self
 
+Note that the name of the concrete type is always "Self".
```
antocuni, line 32: I think I prefer something like `Impl`, `__impl__` or similar.

I agree it would be nice to use `Self`, but in python typing `Self` is special in presence of subclassing. E.g.:

```
class A:
    def foo(self) -> Self:
        return self

class B(A):
    pass

b = B()
b.foo()  # this has type "B", not "A"
```

But because of the way SPy typing currently works, this doesn't happen, and In SPy `b.foo()` would have a type of `A` (when we get inheritance, that is).
So I think that `Self` is misleading.

Hopefully we will find a way to model the same behavior in a SPy-friendly way, but I don't want to think about it now.

playground/myarray.spy
```diff
@@ -2,83 +2,91 @@
...
-        __ll__: gc_ptr[ArrayData[DTYPE]]
-
-        def __new__(length: i32) -> ndarray:
-            data = gc_alloc[ArrayData[DTYPE]](1)
-            data.length = length
-            data.capacity = length
-            data.items = gc_alloc[DTYPE](length)
+@struct
+class Array1d[DTYPE]:
+    # This struct only contains a low level ("__ll__") pointer towards the data leaving on the heap (allocated)
```
antocuni:
```suggestion
    # this struct is a thin zero-cost wrapper around the "__ll__" pointer. The actual data is heap-allocated.
```

antocuni (approved):
LGTM, but see the note about `Self`.

antocuni:
@paugier sorry for taking so long but as promised I finally got back to it.
It overall looks very good and now that I merged #519 it will automatically keep a link between the specialized type and the factory function via `__origin__`.  

We will need to decide whether `issubclass`/`isinstance` would (improperly) consider the `__origin__` as a "superclass", or whether it's maybe better to add a new function like `implements()` or `has_origin()` for that, but that's something which doesn't need to be discussed now.

---

To answer your other comments:

> ```
> <spy struct try2::Point[i32]::Self({'x': W_I32(2), 'y': W_I32(3)})>
> <spy type 'try2::Point[i32]::Self' (struct)>
> <spy struct _list::list[i32]::_ListImpl({'__ll__': W_Ptr('gc', _list::list[i32]::ListData, 0x1001300, length=1)})>
> <spy type '_list::list[i32]::_ListImpl' (struct)>
> ```
> 
> The `::Self` and `::_ListImpl` do not seem very useful. But again it is not an issue directly related to this syntactic sugar.

I agree and I plan to solve it soon :)

---

> ```python
> 
>     @blue
>     def __blue_checks__():
>         if NDIM>2:
>             raise StaticError("NDIM>2 not supported")
> ```

I agree it's useful, but I'm not sure about the name.
One possibility (which I don't fully like) is:

```python
class Foo[T]:
    @staticmethod
    def __INIT__():
        assert NDIM > 2
```

Another one is `__class_init__`, similar to how CPython already does for `__class_getitem__`. I don't like it either but at least has a precedent in CPython.

Anyway, let's discuss it in a separate issue.

---


> ```python
>     @blue
>     def __blue_replace_type__():
>         # nonsense but just to get the idea
>         if NDIM == 0:
>             return f64
> ```

this looks *very* ad-hoc and obscure. I'm leaning towards declaring that if you need such a thing, you should write a `@blue.generic` function manually. I think it would be useful to see a real life example where it's needed.

examples/myarray.spy
antocuni: oh, **now** I see why you need `Self`. Sorry, at the beginning I though it was only a small implementation detail visible only in th FQN.

Yes, I think that given the current constraints, `Self` is a good idea, but we need to think better as soon as we have inheritance.

Ideally, I would also like to be able to write this:
```python
class MyList[T]:
    def foo(self) -> MyList[T]:
        ...
```

I understand why it doesn't work at the moment. The proper solution is something like this:
1. as soon as we execute `MyList[T]`, we create a fwdecl and we add it to vm.globals_w and to the blue cache
2. then, we can use `MyList[T]` freely in the definition
3. when we return it, we can call fwdecl.define.

This mechanism already exists and it's used in the non-generic case, e.g.:
```python
@struct
class Point:

    # at this stage, Point is a fwdecl
    def foo(self) -> Point:
        ...

# here, it became a "real" type
```

See e.g. the relevant code here (and its callers):
https://github.com/spylang/spy/blob/d369e3bd5ec97dbd5473bd65cfad18a48e7928e8/spy/vm/astframe.py#L416-L427

But I don't think it's a merge blocker, I'm fine with implementing this in a subsequent PR.

paugier:
I think it is ready.

> this looks very ad-hoc and obscure. I'm leaning towards declaring that if you need such a thing, you should write a @blue.generic function manually. I think it would be useful to see a real life example where it's needed.

There are examples in the stdlib but I agree that manual `@blue.generic` seems better for that.

PR #449, opened 2026-04-02
353: Add support `argv` parameter of SPy main function

kanin-kearpimy:
Issue: https://github.com/spylang/spy/issues/353

Hi everyone,

this is implementation of **backend** SPy argv parameter for main function.

I separated argument wrapper to `list.c` and `list.h` then calling it in `cmodwriter`.

### However, there is a tricky part happening

image

if any spy files need argument in main like below
```
def main(argv: list[str]) ...
```

It will include `list.h` to such compiled C file to use `spy_wrap_argv`. However, such implementation is in `list.c` which requires to compile before such current working file got compile.

Rough correct pipeline.
```
1) **pre-compile stage**
_list.c (and other built-in standard lib)

2) **current working files**
-> list.c -> [any current working file with `main(argv: list[str])`]
```

Our current pipeline does compile all standard libraries (e.g. _list.spy) at once. It's challenging to modify current pipeline to receive **customized on-the-fly `.c`** to compile as above pipeline. Also, to compile `list.c` with all `libspy` is challenging because `list.c` need `_list.c` to compiled first.

So, now I juggled by copy content of `list.c` and compile it with [any current working file with `main(argv: list[str])`] instead.

Well, I'm not sure I explained it clearly. Please let me know if you need elaboration.

### DIFF
--- a/spy/backend/c/cmodwriter.py
+++ b/spy/backend/c/cmodwriter.py
@@ -157,17 +157,18 @@ class CModuleWriter:
             w_main = self.ctx.vm.globals_w[fqn_main]
             assert isinstance(w_main, W_ASTFunc)
 
-            w_restype, needs_argv = self.ctx.vm.typecheck_main(w_main)
+            w_restype, has_argv = self.ctx.vm.typecheck_main(w_main)
             returns_i32 = w_restype == B.w_i32
 
-            if needs_argv:
+            if has_argv:
                 self.tbc.wb("""
+                    /* helper code for C->SPy argv wrapping */
+
                     #define spy_list_str spy__list$list__builtins$str$_ListImpl
                     #define spy_list_str_new spy__list$list__builtins$str$_ListImpl$__new__
                     #define spy_list_str_push spy__list$list__builtins$str$_ListImpl$_push
 
-                    spy_list_str
-                    spy_wrap_argv(int argc, const char *argv[]) {
+                    spy_list_str spy_wrap_argv(int argc, const char *argv[]) {
                         spy_list_str lst = spy_list_str_new();
                         for (int i = 0; i < argc; i++) {
                             size_t length = strlen(argv[i]);
@@ -178,22 +179,28 @@ class CModuleWriter:
                         }
                         return lst;
                     }
+
+                    #undef spy_list_str
+                    #undef spy_list_str_new
+                    #undef spy_list_str_push
+
+                    /* end of helper code */
                 """)
 
-            if needs_argv and returns_i32:
+            if has_argv and returns_i32:
                 main_src = f"""
                     int main(int argc, const char *argv[]) {{
                         return {fqn_main.c_name}(spy_wrap_argv(argc, argv));
                     }}
                     """
-            elif needs_argv and not returns_i32:
+            elif has_argv and not returns_i32:
                 main_src = f"""
                     int main(int argc, const char *argv[]) {{
                         {fqn_main.c_name}(spy_wrap_argv(argc, argv));
                         return 0;
                     }}
                     """
-            elif not needs_argv and returns_i32:
+            elif not has_argv and returns_i32:
                 main_src = f"""
                     int main(void) {{
                         return {fqn_main.c_name}();
--- a/spy/tests/test_cli.py
+++ b/spy/tests/test_cli.py
@@ -127,27 +127,6 @@ class TestMain:
             _, stdout = self.run(*argset, self.main_spy)
             assert stdout == "hello world\n"
 
-    def test_exit_code(self):
-        src = """
-        def main() -> i32:
-            print("This main return 99 as exit code")
-            return 99
-        """
-        f = self.write("test.spy", src)
-        res = self.runner.invoke(app, [str(f)])
-        assert res.exit_code == 99
-
-    def test_main_wrong_return_type(self):
-        src = """
-        def main() -> str:
-            return "oops"
-        """
-        f = self.write("test.spy", src)
-        res = self.runner.invoke(app, [str(f)])
-        assert res.exit_code == 1
-        output = decolorize(res.output)
-        assert "`main` has the wrong signature" in output
-
     def test_timeit(self):
         _, stdout = self.run("--timeit", self.main_spy)
         assert "main()" in stdout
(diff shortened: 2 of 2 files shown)


spy/backend/c/cmodwriter.py
```diff
@@ -156,23 +156,41 @@ def init_c(self) -> None:
...
-                int main(void) {{
-                    return {fqn_main.c_name}();
-                }}
-                """)
+
+            needs_argv = len(w_main.w_functype.params) == 1
+            returns_i32 = w_main.w_functype.w_restype == B.w_i32
+
+            if needs_argv:
+                self.tbh_includes.wl('#include "spy/list.h"')
```
antocuni: here, instead of including `spy/list.h`, I'd just copy&paste the code for wrapping argv.
Something like this:

```
self.tbc.wb("""
#define spy_list_str spy__list$list__builtins$str$_ListImpl
#define spy_list_str_new spy__list$list__builtins$str$_ListImpl$__new__
#define spy_list_str_push spy__list$list__builtins$str$_ListImpl$_push

spy_list_str
spy_wrap_argv(int argc, const char *argv[]) {
    ...
}
""")
```

it's not ideal, but it's just a very short amount of code and it saves lots of complications

antocuni (commented):
thanks @kanin-kearpimy .

Congrats to find a way to solve the `list[str]` bootstrapping problem, I didn't think about it and indeed it's annoying.

I think we can solve it in a much easier way by just emitting the code directly in the generated C instead of trying to have it in a separate file.

See me comment below

spy/backend/c/cmodwriter.py
kanin-kearpimy: Yeah,

At first I thought of this as well. Let me resolve it this way.

kanin-kearpimy: Okay, refactored it.

spy/backend/c/cmodwriter.py
```diff
@@ -156,23 +156,58 @@ def init_c(self) -> None:
...
-                raise WIP("`main(argv: list[str])` not supported by the C backend")
-
-            if w_restype == B.w_i32:
-                self.tbc.wb(f"""
-                int main(void) {{
-                    return {fqn_main.c_name}();
-                }}
+
+            needs_argv = len(w_main.w_functype.params) == 1
+            returns_i32 = w_main.w_functype.w_restype == B.w_i32
```
antocuni: why do you need this instead of calling `vm.typecheck_main` as it was before?

spy/backend/c/cmodwriter.py
```diff
@@ -156,23 +156,58 @@ def init_c(self) -> None:
...
+
+            if needs_argv:
+                self.tbh.wb("""
```
antocuni: I think this should go to the tbc, not tbh

spy/backend/c/cmodwriter.py
```diff
@@ -156,23 +156,58 @@ def init_c(self) -> None:
...
+                    #define spy_list_str spy__list$list__builtins$str$_ListImpl
+                    #define spy_list_str_new spy__list$list__builtins$str$_ListImpl$__new__
+                    #define spy_list_str_push spy__list$list__builtins$str$_ListImpl$_push
+
+                    spy_list_str
+                    spy_wrap_argv(int argc, const char *argv[]) {
+                        spy_list_str lst = spy_list_str_new();
+                        for (int i = 0; i < argc; i++) {
+                            size_t size_str = strlen(argv[i]);
+                            spy_Str *allo = spy_str_alloc(size_str);
```
antocuni:
```suggestion
                            size_t length = strlen(argv[i]);
                            spy_Str *s = spy_str_alloc(length);
```

I think it's better to call it `length` for consistency with the field in `spy_Str`.
Also, `allo` was a very werid name, better to use `s` (you need to fix it also below)

spy/backend/c/cmodwriter.py
```diff
@@ -156,23 +156,58 @@ def init_c(self) -> None:
...
+                            char *buf = (char *)allo->utf8;
+                            memcpy(buf, argv[i], size_str);
+                            lst = spy_list_str_push(lst, allo);
+                        }
+                        return lst;
+                    }
                 """)
+
+            if needs_argv and returns_i32:
+                execution_code = f"""
```
antocuni:
```suggestion
                main_src = f"""
```

a bit of nitpick, but I think this is clearer. After all, it's the "source of the main function".

spy/libspy/include/spy/list.h
```diff
@@ -0,0 +1,17 @@
+#ifndef SPY_LIST_H
```
antocuni: I think this file is no longer needed, is it?

spy/libspy/src/list.c
```diff
@@ -0,0 +1,20 @@
+#include "spy/list.h"
```
antocuni: same as above, delete it

spy/tests/test_cli.py
```diff
@@ -235,6 +235,96 @@ def test_cwrite(self):
...
+        return_exit_code_file = self.tmpdir.join("return_exit_code_src.spy")
+        return_exit_code_file.write(textwrap.dedent(return_exit_code_src))
+
+        self.run(
+            "build", "--no-compile", "--build-dir", self.tmpdir, return_exit_code_file
+        )
+        main_c = self.tmpdir.join("src", "return_exit_code_src.c")
+        assert main_c.exists()
+        csrc = main_c.read()
+        assert "return 99;" in csrc
```
antocuni: I think I have already deleted/refactored these tests during the review of #419, see e.g. this commit :)
https://github.com/spylang/spy/pull/419/commits/3ebdaa82b2973f0d7425f84fedf4982863e3fb66

There are few code style problems:
- there is no need for a "cwrite" test for every case. The existing `test_cwrite` checks that the option work, but we shouldn't test cwrite for every single combination of options.

Also, the rest of the file follows different naming convention, e.g. it always use `src = """..."""` for the source to compile. Please follow them for consistency.

spy/tests/test_cli.py
```diff
@@ -235,6 +235,96 @@ def test_cwrite(self):
...
+
+    def test_nested_functions_exit_code_cwrite(self):
```
antocuni: this was also deleted by https://github.com/spylang/spy/commit/3ebdaa82b2973f0d7425f84fedf4982863e3fb66 . It seems to test that you can call a function, but that's already tested elsewhere.

Please look at that commit and refactor/redelete all the tests which I have already fixed :)

spy/backend/c/cbackend.py
```diff
@@ -2,6 +2,7 @@
 
 import py.path
 
+import spy.libspy
```
antocuni: I think this is not needed?

antocuni (changes requested):
there seem to be several leftovers from the previous attempt, but overall looks good 👍

antocuni (approved):

antocuni:
thank you a lot for this @kanin-kearpimy ! Merged

PR #462, opened 2026-04-08
Implement hasattr (blue arg)

### LINKED ISSUE #247: Implement `hasattr()`, `getattr()`, and `setattr()`
They would be useful for compatibility with CPython's builtins.

I guess there is a prior question that needs addressing here, should these dynamic elements of the Python language be ported to SPy?

The problem I'm dealing with now is how to declare a custom `__hash__()` implementation in user classes.
So that the `hash()` builtin can lookup the presence of such method in the type.

In more concrete terms:

```python
@blue.generic
def hash(T):
    def hash_i32(v: i32) -> i32: ...
    def hash_str(v: str) -> i32: ...
    ...

    if T == i32:
        return hash_i32
    if T == str:
        return hash_str
    ...
    if hasattr(T, "__hash__"):  # <-------- this is where hasattr() can be useful
        return T.__hash__

    raise TypeError("unhashable type: 'XXX'")
```

paugier:
Fixes #247

I needed hasattr, and I saw #247 and https://github.com/spylang/spy/pull/399.

I did the simple part with a LLM.

The code in spy/vm/modules/operator/attrop.py has to be improved. I'm going to think about it.

### DIFF
--- a/spy/tests/compiler/test_builtins.py
+++ b/spy/tests/compiler/test_builtins.py
@@ -262,6 +262,52 @@ class TestBuiltins(CompilerTest):
         )
         self.compile_raises(src, "foo", errors)
 
+    def test_hasattr(self):
+        src = """
+        @struct
+        class Point:
+            x: i32
+            y: i32
+
+        def foo() -> bool:
+            p = Point(1, 2)
+            return hasattr(p, 'x')
+        """
+        mod = self.compile(src)
+        assert mod.foo() == True
+
+    def test_hasattr_false(self):
+        src = """
+        @struct
+        class Point:
+            x: i32
+            y: i32
+
+        def foo() -> bool:
+            p = Point(1, 2)
+            return hasattr(p, 'z')
+        """
+        mod = self.compile(src)
+        assert mod.foo() == False
+
+    def test_hasattr_red(self):
+        src = """
+        @struct
+        class Point:
+            x: i32
+            y: i32
+
+        def foo() -> bool:
+            var attr = "x"  # this is red
+            p = Point(1, 2)
+            return hasattr(p, attr)
+        """
+        errors = expect_errors(
+            "expected blue argument",
+            ("this is red", "attr"),
+        )
+        self.compile_raises(src, "foo", errors)
+
     @only_interp
     def test_dir(self):
         src = """
--- a/spy/tests/compiler/test_builtins.py
+++ b/spy/tests/compiler/test_builtins.py
@@ -1,6 +1,12 @@
 import pytest
 
-from spy.tests.support import CompilerTest, expect_errors, no_C, only_interp
+from spy.tests.support import (
+    CompilerTest,
+    expect_errors,
+    no_C,
+    only_interp,
+    skip_backends,
+)
 from spy.vm.builtin import builtin_method
 from spy.vm.opspec import W_MetaArg, W_OpSpec
 from spy.vm.primitive import W_I32, W_Dynamic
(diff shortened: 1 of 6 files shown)


paugier:
I think this PR is now ready for review.

paugier:
A question: actually test_hasattr_custom_getattribute could be in test_attrop.py?

Also, I see that test_getattr_setattr_custom is written building by hand in Python the class with `__getattribute__`.

In contrast, for simplicity I used SPy. Is there a difference? What is better?

spy/tests/compiler/test_builtins.py
```diff
@@ -262,6 +268,108 @@ def foo() -> None:
...
+        class Point:
+            x: i32
+            y: i32
+
+        def foo() -> bool:
+            p = Point(1, 2)
+            return hasattr(p, 'z')
+        """
+        mod = self.compile(src)
+        assert mod.foo() == False
```
antocuni: this test can probably be merged with the previous one.
You can either have two functions one returning `hasattr(p, 'x')` and another `hasattr(p, 'z')`, or you could have a single function returning `tuple[bool, bool]`.

Also, would be nice to have a test for `hasattr(Point, 'x')`.

spy/vm/astframe.py
```diff
@@ -1114,6 +1114,11 @@ def eval_expr_Call(self, call: ast.Call) -> W_MetaArg:
             w_opimpl = self.vm.call_OP(call.loc, OP.w_GETATTR, args_wam)
             return self.eval_opimpl(call, w_opimpl, args_wam)
 
+        elif wam_func.color == "blue" and wam_func.w_blueval is B.w_hasattr:
+            self.special_calls[call] = "hasattr"
+            w_opimpl = self.vm.call_OP(call.loc, OP.w_HASATTR, args_wam)
+            return self.eval_opimpl(call, w_opimpl, args_wam)
+
```
antocuni, line 1121: is it really necessary to special-case `hasattr` as we do for `getattr`?
The point of special-casing `getattr` is to get a nicer error message, but for `hasattr` you don't need that (I think?).

It might be possible to implement it as a fully normal metafunc, unless I'm missing something.

antocuni (commented):
The PR seems mostly good, the only thing is that it might be easier to implement it as a normal metafunction instead of a "full" operator like `w_GETATTR`.

Something like (pseudocode):

```python
@BUILTINS.builtin_func(color="blue", kind="metafunc")
def w_hasattr(vm: "SPyVM", wam_obj: W_MetaArg, wam_name: W_MetaArg) -> W_OpSpec:
     w_opimpl = vm.call_OP(OP.w_GETATTR, [wam_obj, wam_name])
     if w_opimpl.is_null():
         return W_OpSpec.const(B.w_False)
     else:
         return W_OpSpec.const(B.w_True)
```

I *think* it might work, but I might be missing some important details.
The tests that you wrote are very good and comprehensive though, so they will tell us if this technique is enough :)

> A question: actually test_hasattr_custom_getattribute could be in test_attrop.py?

that's a good question. If we make `hasattr` a "normal" metafunc, then I think it should stay where it is.

> Also, I see that test_getattr_setattr_custom is written building by hand in Python the class with __getattribute__.
> In contrast, for simplicity I used SPy. Is there a difference? What is better?

I think it's fine to use SPy. IIRC, the original tests were written when we didn't support custom `__dunder__` methods at app-level at all, that's why I had to write them at interp-level

spy/vm/astframe.py
paugier: I tried with `vm.call_OP(None, OP.w_GETATTR, [wam_obj, wam_name])` but it is not that simple since it raises a SPyError both for the AttributeError and for red name. I can catch the error and it should be possible to inspect it but it seems ugly.

spy/tests/compiler/test_builtins.py
paugier: I simplified these tests. Regarding `hasattr(Point, 'x')`, it fails with "WIP: implement me: descriptor accessed via class".

spy/vm/astframe.py
antocuni: ok, I get it now. I agree it's not worth the extra complexity, it's better to keep it simple and in the same style as getattr

antocuni (approved):

antocuni:
@paugier sorry it took so long to review, LGTM. Merging.

PR #467, opened 2026-04-10
@force_inline to force inlining of red functions

### LINKED ISSUE #464: `@force_inline` decorator
We can already do loop fusion with current SPy (https://github.com/spylang/demos/pull/3/changes).

However, the result of redshift is not readable for humans (it even becomes very messy compared to without loop fusion) since there is no inlining during redshift.

@antocuni wrote

> Yes, I agree it would be nice to have a better way to see it visually. Maybe a `@force_inline` decorator to do inlinining during redshifting would do the job, although I'd say that it's not the highest priority.

I open this issue to be able to think about this and not forgot about it.

It's highly probable that the C compilers will inline these small functions (especially with `--release`) but it is less clear and predictable that if inlining can be done during redshift.

I tend to think that it would indeed be good to have a `@force_inline` decorator, which could be limited to one-statement functions (see https://github.com/paugier/spy-demos/blob/loop-fusion/loop-fusion/lib_ndarray.spy where I added few `# @force_inline` where they would be needed). The advantages I see are:

- clear and more readable redshift results which is good for SPy developers
- more predictable inlining
- nice for communication. For example, I could try to submit an abstract to a conference about array implementation and loop fusion in SPy. Nicer and more readable redshift output would be useful in this perspective.

By limiting `@force_inline` to one-statement functions, it seems to me that the implement could be quite simple, at least for the typed AST. I guess that we would also need to modify other things than the typed AST (symtable, ?).

This would be a simple optimization pass happening at the end of redshift. I don't know if there could potentially be other similar optimization passes at the end of redshift and if we should think about a general mechanism (as for transformations before redshift)?

It would be nice for the redshift command to be able to disable this optimization (so that it can be inspected).

I would be motivated to work on the implementation of this decorator, which fits well in the project of communicating about loop fusion in SPy. Some technical indications on the implementation would of course be welcome.

Anyway, the first question is: is it a good idea to add `@force_inline`!

@antocuni does all this makes sense? What do you think?

paugier:
Fixes #464 

I didn't want to touch too much doppler so here is an implementation with a new file spy/inliner.py.

For now `@force-inline` functions must consist only in one return statement.

It was a bit too difficult alone so it is mostly AI generated.

In particular, Claude told me that it found a bug about precedence in backend/spy.py (one of the test didn't pass).

### DIFF
--- a/examples/force_inline.spy
+++ b/examples/force_inline.spy
@@ -1,0 +1,41 @@
+"""
+@force_inline is a special decorator (like the @blue family) that forces
+inlining of a red function at the end of redshift.
+
+The main use cases are:
+
+1. Readability: write small helper functions without polluting the
+   redshifted output with extra call sites.
+
+2. Predictable inlining: guarantee that the SPy compiler inlines a
+   function regardless of C compiler heuristics.
+
+Currently @force_inline only supports functions with a single `return`
+statement (no intermediate assignments).
+
+To inspect what inlining does, compare:
+
+  $ spy redshift force_inline.spy
+  $ spy redshift force_inline.spy --no-inline
+
+With inlining, the body of `polynomial` becomes a single expression:
+
+  def polynomial(a: i32) -> i32:
+      return a * a + (a + 1) * (a + 1) + 2 * (a + 2)
+"""
+
+
+@force_inline
+def square(x: i32) -> i32:
+    return x * x
+
+
+def polynomial(a: i32) -> i32:
+    return square(a) + square(a + 1)
+
+
+def main() -> None:
+    x = 2
+    assert polynomial(x) == x * x + (x + 1) * (x + 1)
+    # 4 + 9 -> 13
+    print(polynomial(x))
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -544,6 +544,7 @@ class FuncDef(Stmt):
     docstring: Optional[str]
     body: list["Stmt"]
     decorators: list["Expr"]
+    is_force_inline: bool = False
     symtable: Any = field(repr=False, default=None)
 
     def shortrepr(self) -> Optional[str]:
--- a/spy/backend/interp.py
+++ b/spy/backend/interp.py
@@ -47,13 +47,12 @@ class InterpModuleWrapper:
 
         if isinstance(w_obj, W_ASTFunc):
             w_func = w_obj
-            if not w_func.is_valid:
-                # let's find the redshifted version
+            while not w_func.is_valid:
+                # follow the chain: orig -> redshifted -> inlined (if applicable)
                 assert w_func.w_redshifted_into is not None
                 w_func = w_func.w_redshifted_into
-                assert w_func.redshifted
-                # tmp change so that force_inline tests pass
-                # assert self.vm.lookup_global(w_func.fqn) is w_func
+            if w_func.redshifted:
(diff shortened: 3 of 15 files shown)


paugier:
With this version, loop fusion works! `x + x + x` becomes a loop with something like:

```
`unsafe::gc_ptr[f64]::getitem_byval`(
    `unsafe::ptr_getfield_byval[unsafe::gc_ptr[f64]]`(
        `ndarray[f64]::__get___ll____`(
            `LazyAdd[f64, ndarray[f64], ndarray[f64]]::__get_left__`(
                `LazyAdd[f64, LazyAdd[f64, ndarray[f64], ndarray[f64]], ndarray[f64]]::__get_left__`(expr))), 'items', 8), i)
+ `unsafe::gc_ptr[f64]::getitem_byval`(
    `unsafe::ptr_getfield_byval[unsafe::gc_ptr[f64]]`(
        `ndarray[f64]::__get___ll____`(
            `LazyAdd[f64, ndarray[f64], ndarray[f64]]::__get_right__`(
                `LazyAdd[f64, LazyAdd[f64, ndarray[f64], ndarray[f64]], ndarray[f64]]::__get_left__`(expr))), 'items', 8), i)
+ `unsafe::gc_ptr[f64]::getitem_byval`(
    `unsafe::ptr_getfield_byval[unsafe::gc_ptr[f64]]`(
        `ndarray[f64]::__get___ll____`(
            `LazyAdd[f64, LazyAdd[f64, ndarray[f64], ndarray[f64]], ndarray[f64]]::__get_right__`(expr)), 'items', 8), i)
```

Formatted like that, it's nearly readable by common people :slightly_smiling_face:

paugier:
I think I would need feedback since this PR is now good enough for me (reasonable tests, green CI, reasonable architecture from my point of view, and it does what I need for loop fusion).

This is the simplest version of `@force_inline` I could think about. Restricted to functions with only one return (which greatly simplifies things) and clear separation with `doppler.py`, which is not even modified.

One thing that I like is how small are the changes in existing files. Nearly all the logic is in a new filed dedicated to inlining.

antocuni (commented):
hi @paugier.
I did just a quick skim at the PR without looking too much into details.

I'm not sure I like the approach "inline after redshifting".
As you could see you had to write some kind of ad-hoc recursion in `inliner.py`, which is ok but it feels a bit strange since we are ALREADY doing exactly that kind of recursion in `doppler.py`.

In the meantime, I realized that a "full" `@force_inline` is probably what I need to implement a proper `print()` function, so my plan is to try to implement it differently, as part of doppler.
Hopefully this will also eliminate the need for "hacks" like `_effective_precedence`.

A lot of the content of this PR will probably be useful for the "new" PR, especially the tests, so if you don't mind I'll probably cherry-pick some of them.

Sorry for making you doing "unnecessary" work, but I think this is the way to go.
I'm leaving this PR open though, because I still don't know whether my approach will work :)

paugier:
> Sorry for making you doing "unnecessary" work, but I think this is the way to go.

No issue with that! I agree and I realize the limitations and drawbacks of this simple approach. I think most of the tests will be useful though.

Good luck with the proper full `@force_inline`. Happy that this need coming from loop fusion could finally be useful for something totally different like a proper `print()`. I'm eager to see how you will do that!

antocuni:
closing this as it was superseded by #504

PR #469, opened 2026-04-17
Support tuple targets in for loops

### LINKED ISSUE #455: Tuple unpacking in for loops
**WARNING - full AI-assisted solutions are not welcome for this issue**
The point of "Good first issues" is to give a chance to humans to interact and learn about the codebase. Writing code using AI assistant prevents this.

---

The following currently fails:
```python
def main() -> None:
    lst = [(1, 2), (3, 4)]
    for (a, b) in lst:
        print(a + b)
```

```
❯ spy /tmp/x.spy 

ParseError: not implemented yet: complex for loop targets
  | /tmp/x.spy:3
  |     for (a, b) in lst:
  |         |____| this is not supported
```

We should support it.

I think that the easiest way is to improve `ASTFrame._desugar_for`, specifically here:
https://github.com/spylang/spy/blob/1c6ad6bdb59541c44b43a55830b2ff6b8c03427c/spy/vm/astframe.py#L755-L765

Instead of generating an `ast.Assign` node, we should generate an `ast.UnpackAssign` node.

Green00101:
Closes #455 
## Summary

This PR adds support for tuple targets in `for` loops.

Example:
```python
for a, b in pairs:
    pass
```
Before this change, `for` loops only supported a single name target.

## Changes
- allow `ast.For.target` to be either a single name or a list of names
- parse tuple targets in `for` loops
- declare and flatten all names from tuple targets
- emit `UnpackAssign` in `_desugar_For` for tuple targets
- update backend output to print `for a, b in pairs:`

## Notes
This PR supports simple tuple targets only.
Nested unpacking is still rejected.

## Testing

Added tests for parser and backend output.

While testing, I found a separate doppler issue:
`for t in lst` fails when `lst` is `list[tuple[...]]`, with what seems to be a `gc_ref[tuple]` vs `tuple` mismatch.

Because that looks independent from tuple targets in `for` loops, I did not include that compiler-level test in this PR.

Happy to make any changes if needed.

### DIFF
--- a/spy/analyze/scope.py
+++ b/spy/analyze/scope.py
@@ -427,13 +427,18 @@ class ScopeAnalyzer:
         # What is the "type_loc" of i? It's an implicit declaration, and its
         # value depends on the iterator returned by range. So we use
         # "range(10)" as the type_loc.
-        self.define_name(
-            forstmt.target.value,
-            "var",
-            "auto",
-            forstmt.target.loc,
-            forstmt.iter.loc,
-        )
+        if isinstance(forstmt.target, ast.StrConst):
+            targets = [forstmt.target]
+        else:
+            targets = forstmt.target
+        for target in targets:
+            self.define_name(
+                target.value,
+                "var",
+                "auto",
+                target.loc,
+                forstmt.iter.loc,
+            )
 
         # Increment loop depth before processing body
         self.loop_depth += 1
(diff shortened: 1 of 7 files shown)


antocuni:
Hello @Green00101.
Before I review, could you please disclose whether you used any AI assistant and, in case, how much?

Green00101:
Sure.

I first used Opus to understand the codebase and identify the key functions I should focus on. Then, I used ChatGPT to help me understand how those important functions work.

Next, I tried to add code to the codebase by myself. After that, I used ChatGPT to review my code and make changes based on its feedback.

Finally, I used Opus to generate test cases. During this process, I encountered a strange failure that I could not understand. I asked three different AI tools about it, and they all gave me the same answer: it was a Doppler issue. So I mentioned that in the description.

PR #470, opened 2026-04-17
Add docs on the main function, argv and return codes

JeffersGlass:
Adds a documentation page for current info on the `main` function, and optionally providing command line arguments and return codes. Based on the work in #353 and elsewhere.

### DIFF
--- a/docs/src/howto/main_function.md
+++ b/docs/src/howto/main_function.md
@@ -1,14 +1,16 @@
 title: The Main Function
 ---
 
-All SPy programs must have a `main` function. The main function is the entry point of the program, and it is where the execution of the program begins.
+All SPy programs that are run in the interpreter or compiled must have a `main` function. The main function is the entry point of the program, and it is where the execution of the program begins.
 
 ```py
 def main() -> None:
     print("Hello world")
 ```
 
-Not every `.spy` module needs a main function, but the module invoked by, e.g. `spy execute foo.spy` must contain a main function.
+Not every `.spy` module needs a main function, but the module invoked by, e.g. `spy foo.spy` `spy build foo.spy` must contain a main function.
+
+Modules which are compiled as a library (e.g. `spy build --target lib foo.spy` or `--target py-cffi`) do not need a main function.
 
 ## Return Codes
 
@@ -26,9 +28,9 @@ $ echo $?
 ```
 
 
-## Passed Arguments (Interpretter Only)
+## Accessing Command Line Arguments
 
-If the `main` function accepts a list of strings as an argument, the SPy program will accept arguments from the command line, both when running in interpretted
+If the `main` function accepts a list of strings as an argument, the SPy program will accept arguments from the command line, both when running in interpreted
 
 ```py
 #args.spy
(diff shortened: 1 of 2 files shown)


kanin-kearpimy:
Hello @JeffersGlass,

Thank you so much for this doc. Overall content is comprehensive. However, **main() argument** now supports both compiler and interpreter. So, we can compile code with main function arguments regarding https://github.com/spylang/spy/pull/449.

docs/src/howto/main_function.md
```diff
@@ -0,0 +1,53 @@
...
+    return 123
+```
+```
+$ uv run spy retcode.spy
+$ echo $?
+123
+```
+
+
+## Passed Arguments (Interpretter Only)
```
dpdani, line 29:
```suggestion
## Passed Arguments (Interpreter Only)
```

docs/src/howto/main_function.md
```diff
@@ -0,0 +1,53 @@
...
+
+If the `main` function accepts a list of strings as an argument, the SPy program will accept arguments from the command line, both when running in interpretted
```
dpdani, line 31:
```suggestion
If the `main` function accepts a list of strings as an argument, the SPy program will accept arguments from the command line, both when running in interpreted
```

Seems the sentence is not finished?

docs/src/howto/main_function.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+All SPy programs must have a `main` function. The main function is the entry point of the program, and it is where the execution of the program begins.
+
+```py
+def main() -> None:
+    print("Hello world")
+```
+
+Not every `.spy` module needs a main function, but the module invoked by, e.g. `spy execute foo.spy` must contain a main function.
```
antocuni, line 11:
```suggestion
Not every `.spy` module needs a main function, but the module entry point invoked by, e.g. `spy foo.spy` or `spy build foo.spy` must contain a main function.
```

antocuni: I think this is a big more correct, but it become more nuanced, because e.g. `spy build --target lib` and `spy build --target py-cffi` do not require a `main`.

Maybe we can reword the section by starting this way:

> The execution of a SPy program starts from a `main` function; you need a `main` if you try to run the interpreter (`spy foo.spy`) or to build an executable (`spy build foo.spy`). You don't need a `main` if you build a library (e.g. `spy build --target lib foo.spy`).

However, I don't fully like the wording. What do you think?

docs/src/howto/main_function.md
antocuni: as @kanin-kearpimy already pointed out, the arguments works also on compiled mode now 🎉.

Maybe instead of "passed arguments" we can call this section "Accessing command line arguments"?

docs/src/howto/main_function.md
```diff
@@ -0,0 +1,53 @@
...
+#args.spy
+def main(args: list[str]) -> None:
+    print(args[1])
+```
+```
+$ uv run spy args.spy 999
+999
+```
+
+As with CPython, args[0] is the name of the string passed to the uv runtime:
```
antocuni, line 43: probably we should say that this is the equivalent of CPython's `sys.argv`

docs/src/howto/main_function.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+
+## Passed Arguments (Interpretter Only)
+
+If the `main` function accepts a list of strings as an argument, the SPy program will accept arguments from the command line, both when running in interpretted
+
+```py
+#args.spy
+def main(args: list[str]) -> None:
```
antocuni, line 35: I think we should use `argv` as the canonical name, for consistency to CPython

antocuni (approved):
Thank you @JeffersGlass!
I left some minor comments but overall looks good 💪

kanin-kearpimy (approved):

PR #471, opened 2026-04-17
Abs as a metafunc in stdlib to support more types

### LINKED ISSUE #461: abs for float
`abs(1.2)` gives

```
  * bug_abs::main at ~/dev/spy-demos/loop-fusion/bug_abs.spy:4
  |     print(abs(1.2))
  |           |______|

TypeError: mismatched types
  | ~/dev/spy-demos/loop-fusion/bug_abs.spy:4
  |     print(abs(1.2))
  |               |_| expected `i32`, got `f64`

  | ~/dev/spy/spy/vm/modules/builtins.py:34
  | def w_abs(vm: "SPyVM", w_x: W_I32) -> W_I32:
  | |__________________________________________| function defined here
```

paugier:
Fixes #461

@antocuni should I do the same also for min and max in this PR?

### DIFF
--- a/spy/analyze/symtable.py
+++ b/spy/analyze/symtable.py
@@ -190,6 +190,7 @@ class SymTable:
             loc = w_obj.def_loc if isinstance(w_obj, W_BuiltinFunc) else None
             add_sym(attr, ImportRef("builtins", attr), loc)
 
+        add_sym("abs", ImportRef("_builtins", "abs"))
         add_sym("range", ImportRef("_range", "range"))
         add_sym("list", ImportRef("_list", "list"))
         add_sym("tuple", ImportRef("_tuple", "tuple"))
--- a/spy/analyze/symtable.py
+++ b/spy/analyze/symtable.py
@@ -190,8 +190,9 @@ class SymTable:
             loc = w_obj.def_loc if isinstance(w_obj, W_BuiltinFunc) else None
             add_sym(attr, ImportRef("builtins", attr), loc)
 
-        for attr in ("abs", "min", "max"):
-            add_sym(attr, ImportRef("_builtins", attr))
+        add_sym("abs", ImportRef("_builtins", "abs"))
+        add_sym("min", ImportRef("_builtins", "min"))
+        add_sym("max", ImportRef("_builtins", "max"))
 
         add_sym("range", ImportRef("_range", "range"))
         add_sym("list", ImportRef("_list", "list"))
--- a/spy/analyze/symtable.py
+++ b/spy/analyze/symtable.py
@@ -193,7 +193,6 @@ class SymTable:
         add_sym("abs", ImportRef("_builtins", "abs"))
         add_sym("min", ImportRef("_builtins", "min"))
         add_sym("max", ImportRef("_builtins", "max"))
-
         add_sym("range", ImportRef("_range", "range"))
         add_sym("list", ImportRef("_list", "list"))
         add_sym("tuple", ImportRef("_tuple", "tuple"))
--- a/spy/analyze/symtable.py
+++ b/spy/analyze/symtable.py
@@ -190,7 +190,9 @@ class SymTable:
             loc = w_obj.def_loc if isinstance(w_obj, W_BuiltinFunc) else None
             add_sym(attr, ImportRef("builtins", attr), loc)
 
-        add_sym("abs", ImportRef("_builtins", "abs"))
+        for attr in ("abs", "min", "max"):
+            add_sym(attr, ImportRef("_builtins", attr))
+
         add_sym("range", ImportRef("_range", "range"))
         add_sym("list", ImportRef("_list", "list"))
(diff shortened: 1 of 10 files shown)


paugier:
I don't know if [93d8ee8](https://github.com/spylang/spy/pull/471/commits/93d8ee85daf7abe0ae9285064650352a1419e8df) is the right solution but I don't see how I could add a function implemented in stdlib in the `builtins` module.

Is [93d8ee8](https://github.com/spylang/spy/pull/471/commits/93d8ee85daf7abe0ae9285064650352a1419e8df) acceptable? Is it possible to allow `from builtins import abs`?

A solution would be to change the name of `builtins` in `b.py`, use `stdlib/builtins.spy` and import all builtins functions defined in `vm/modules/builtins.py`...

spy/tests/compiler/test_importing.py
```diff
@@ -6,7 +6,8 @@ class TestImporting(CompilerTest):
 
     def test_import(self):
         mod = self.compile("""
-        from builtins import abs as my_abs
+        # abs is implemented in stdlib/_builtins.spy
+        from _builtins import abs as my_abs
```
antocuni: the original meaning of this test was to check that we can import stuff from a builtin module. I actually wrote `abs` specifically to have something to import in this test :).

I think you can use e.g. `from math import sin as my_sin` (or any other `math` function) to keep the spirit of the test intact

antocuni (approved):
LGTM. 

> Is [93d8ee8](https://github.com/spylang/spy/pull/471/commits/93d8ee85daf7abe0ae9285064650352a1419e8df) acceptable? Is it possible to allow from builtins import abs?

I don't care too much about being able to do `from builtins import abs`, I care that the test checks that I can import something :). See the inline comment.

+1 for this approach to get a generic `abs`. Eventually I would like some syntax sugar to do it automatically, but for now it's good enouch.

+1 also for having min and max. In theory we could also try to support the `max(iterable)` case but for now I'm happy with just `max(a, b)` and `min(a, b)`

JeffersGlass:
Feel free to merge this as-is of course, but a little suggestion for the docs: the description of `abs()` in `docs/src/api_reference/python_builtins.md` could be changed to:

```md
Works for any type which is can be compared (less-than) against `0` and accepts unary minus `-` operator. The `__abs__` attribute is not currently supported.
```

paugier:
Should be ready for another review, now with `min` and `max` in .spy.

@JeffersGlass This doc preview is really convenient! I modified `python_builtins.md` so you might want to have a look.

spy/analyze/symtable.py
```diff
@@ -190,6 +190,9 @@ def add_sym(attr: str, impref: ImportRef, loc: Optional[Loc] = None) -> None:
             loc = w_obj.def_loc if isinstance(w_obj, W_BuiltinFunc) else None
             add_sym(attr, ImportRef("builtins", attr), loc)
 
+        for attr in ("abs", "min", "max"):
+            add_sym(attr, ImportRef("_builtins", attr))
+
```
antocuni:
```suggestion
        add_sym("abs", ImportRef("_builtins", "abs"))
        add_sym("min", ImportRef("_builtins", "min"))
        add_sym("max", ImportRef("_builtins", "max"))

```

I like this better as it's more explicit

antocuni (commented):
review WIP, but let me commit this to run tests

spy/tests/stdlib/test__builtins.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_abs_f64(self):
+        mod = self.compile(
+            """
+        def foo(x: f64) -> f64:
+            return abs(x)
+        """
+        )
+        assert mod.foo(10.0) == 10.0
+        assert mod.foo(-20.0) == 20.0
+
```
antocuni, line 15: I leave this comment for posterity, although I'm fixing it locally to save roundtrips.

`abs(i32)` was tested in `test_basic::test_builtin`, but the point of that test was to check that we can call a builtin function. If abs is no longer builtin, better to move the test here.

Similar for min/max: they were in test_basic for historical reasons, but now it makes sense to test everything here where they belong.

Then there is a nitpick: I think that your editor (or codeformatter maybe) does a weird thing with the location/indentation of multiline strings passed to functions. In the rest of the codebase, the style is this:
```python
mod = self.compile("""
    ...
""")
```

This is not the first time it happens, and not your fault: it seems that every tool choose a different way to format these things.

The "new" style which I have been using is this, which seems to work consistently well with all tools:
```
src = """
...
"""
mod = self.compile(src)
```

antocuni: ok, I see that github now shows this comment against my code, so it doesn't make any sense 😂.

The multine thing I was talking about was this:
```
        mod = self.compile(
            """
        def foo(x: f64) -> f64:
            return abs(x)
        """
        )
```

antocuni (approved):

antocuni:
LGTM! Merging now, thanks

Issue #476, opened 2026-04-20
Allow easier dev setup with a modern project manager


paugier:
I see that the instructions for Local development setup (https://github.com/spylang/spy#local-development-setup) mention simply `pip install -e .[dev]` and `make -C spy/libspy`.

I understand that you want to avoid forcing developers to use a package manager.

However, it does not help setting up a development environment and thus using SPy locally.

The current instructions assumes that people know that they should create and activate a virtual env, or worse that they have an global environment where they run `pip install -e .` for different projects.

Currently, I personally use:

```sh
sudo apt install libgc
# purge the repo
uv venv .venv -p 3.12
. .venv/bin/activate
uv pip install -e .[dev]
make -C spy/libspy
```

This works but it is not that great (long and no lock file).

I think there should also have alternative simpler instructions to install SPy and setup a good locked dev environment with all dependencies. Therefore, it seems to me that SPy should have a versioned lock file.

I thought about which project manager could be used for SPy. Of course, UV is a good candidate. However, SPy is not Python and the plan is finally to remove the Python dependency so it seems to me that Pixi would also be a very good choice.

- An advantage is to be able to use both PyPI and conda-forge, for Python and non-Python dependencies (see for example #473),
- Pixi has been chosen by Mojo,
- Pixi is really nice for developers (in particular fast, robust and simple for simple things).

I worked on a proof of concept of using SPy with Pixi and I'm going to soon submit a PR to help the discussion. With this PR, one gets a good dev env with:

```
pixi run make-libspy
pixi shell
```

Note that I don't propose here that SPy development requires Pixi! My point is that people should be able to use Pixi (or UV) to get a nicer development experience for using SPy from its repo.

I see that it would help me with SPy dev and I think it could be useful for other people trying SPy.

antocuni:
I agree with all you said.
There is no big reason for why this is not yet supported, only that I never felt the need because I'm personally happy with my current setup.

I agree that one big advantage of pixi is to be able to handle non-python dependencies such as libgc.

I'm fine with having a pixi-friendly workflow as long as it's not required. My only concern is how to ensure that _both_ supported installation methods work. Probably we need a separate github action to test that the pixi-based setup works and can e.g. run `spy examples/hello.spy`.

antocuni:
let's continue the discussion directly on PR #477

antocuni:
@paugier I merged #477 . I guess we can close this?

paugier:
Yes we can!

PR #477, opened 2026-04-20
Show a simple Pixi setup

paugier:
Related to #476 

Setup a dev env

```sh
pixi run make-libspy
pixi shell
# or from everywhere
pixi shell -m ~/dev/spy
```

Other useful commands (with tab auto-completion)

```sh
pixi run fmt-py
pixi run check-lint
pixi run check-fmt
pixi run doc-serve
pixi run test-xdist
```


.github/workflows/tests-pixi.yml
```diff
@@ -0,0 +1,68 @@
...
+          pixi run check-lint
+
+      - name: Format check
+        run: |
+          pixi run check-fmt
+
+      - name: Run tests
+        run: |
+          pixi run pytest -n auto -m "not mypy" --junit-xml=test-results.xml
+
```
antocuni: if I understand correctly this PR runs the whole test suite twice (one with pixi, one with pip)?

I'd prefer to avoid that, I think it's enough a smoke test such as `spy build -x examples/hello.spy` to ensure that things generally work.

pyproject.toml
```diff
@@ -119,3 +120,37 @@ inline-quotes = "double"
...
+spylang = { path = ".", editable = true }
+
+[tool.pixi.environments]
+default = { features = ["dev"], solve-group = "default" }
+docs = { features = ["docs"], solve-group = "default" }
+
+[tool.pixi.activation.env]
+C_INCLUDE_PATH = "$PIXI_PROJECT_ROOT/.pixi/envs/default/include:$C_INCLUDE_PATH"
+LIBRARY_PATH = "$PIXI_PROJECT_ROOT/.pixi/envs/default/lib:$LIBRARY_PATH"
+DYLD_LIBRARY_PATH = "$PIXI_PROJECT_ROOT/.pixi/envs/default/lib:$DYLD_LIBRARY_PATH"
```
antocuni, line 146: I guess this is needed to be able to link libgc?
If so, it's worth adding a comment

pyproject.toml
```diff
@@ -119,3 +120,37 @@ inline-quotes = "double"
...
+
+[tool.pixi.tasks]
+make-libspy = "make -C spy/libspy"
+test = "pytest spy/tests -v"
+test-xdist = "pytest -n auto -m 'not mypy'"
+check-lint = "ruff check ."
+check-fmt = "ruff format --check ."
+fmt-py = "ruff format ."
```
antocuni: I never used pixi before so I don't know whether these are conventional names or ad-hoc.
If they are ad-hoc, what about usinf `ruff-check`, `ruff-format-check` and `ruff-format`? They look much more intuitive to me :)

antocuni (approved):
I'm super happy about this 🎉.
Left some minor comments.

.github/workflows/tests-pixi.yml
paugier: I simplified a lot the workflow checking Pixi setup. Only a smoke test is run.

pyproject.toml
paugier: Done.

pyproject.toml
paugier: Done!

antocuni (approved):

antocuni:
@paugier after merging, the pixi workflow fails:
https://github.com/spylang/spy/actions/runs/25753173159

do you feel like investigating it?

Issue #416, opened 2026-03-14
Playground examples should be tested


paugier:
Currently, myarray.spy gives in the playground:

```
  | /home/pyodide/test.spy:35
  |             data = gc_alloc(ArrayData(DTYPE))(1)
  |                    |________________________| operator::CALL called here
```

I think it should just be `gc_alloc[ArrayData(DTYPE)]`. I'm going to submit a PR to fix that.

However, this issue is about the fact that it should not happen. Playground examples should be tested and just work 🙂

paugier:
I'm working on this issue. I should soon open a PR for:

- playground examples in the examples directory,
- test all examples,
- examples numbered,
- homogenization of the examples, with a doctring.

One issue I see for the playground is that the tab layout is not adapted to include all the examples and I don't feel that I can be good to improve the playground in this direction.

antocuni:
> One issue I see for the playground is that the tab layout is not adapted to include all the examples and I don't feel that I can be good to improve the playground in this direction.

I think it's fine to add only selected examples to the playground. Eventually we should probably decide more explicitly which examples are worth to keep and which should be removed.

PR #480, opened 2026-04-22
Support isolated module-level str

paugier:
This is a very simple and small PR to support isolated strings at the module level. A commit comes from #448 and I added a simple test.

Isolated strings at the module level are useful for examples and I would need this for my work on examples and playground (related to #416).

Since this corresponds to a very small change, I guess it should be easier/faster to review (and merge when ready).

### DIFF
--- a/spy/tests/test_parser.py
+++ b/spy/tests/test_parser.py
@@ -1703,3 +1703,11 @@ class TestParser:
             )
             """,
         )
+
+    def test_module_level_str(self):
+        mod = self.parse("""
+        "The docstring"
+        a = 1
+        "A module-level string"
+        """)
+        assert mod.docstring == "The docstring"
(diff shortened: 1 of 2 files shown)


spy/tests/test_parser.py
```diff
@@ -1703,3 +1703,11 @@ def foo() -> i32:
...
             """,
         )
+
+    def test_module_level_str(self):
+        mod = self.parse("""
+        "The docstring"
+        a = 1
+        "A module-level string"
+        """)
+        assert mod.docstring == "The docstring"
```
antocuni, line 1713:
```suggestion
    def test_module_level_str(self):
        mod = self.parse("""
        "The docstring"
        a = 1
        "This string will be ignored"
        """)
        assert mod.docstring == "The docstring"
```

antocuni (approved):
LGTM! I'll add a minor change in the test and merge it

PR #481, opened 2026-04-22
Move .spy playground files in examples and test all examples

### LINKED ISSUE #416: Playground examples should be tested

paugier:
Fixes #416


paugier:
This is a first step but it might be enough for now. At least, it can be useful to have feedback on this strategy.

examples/expected_output/dict_example.txt
```diff
@@ -0,0 +1,2 @@
+20
+<spy struct _dict::dict[i32, i32]::_dict({'__ll__': W_Ptr('gc', _dict::dict[i32, i32]::DictData, 0x1001460, length=1)})>
```
antocuni: this is a bit fragile (we should have a better repr for the dict, but that's a different story) because it contains a hardcoded address, which happens to be always the same because `malloc` on WASM is very deterministic.

If we want to be more robust, we probably want to support some kind of regexp/glob syntax in these "expected" files.

Anyway, not a big deal and not a merge blocker, just something to keep in mind for the future.

examples/test_examples.py
```diff
@@ -0,0 +1,57 @@
...
+    assert result.returncode == expected_returncode(spy_file.name), (
+        f"spy exited with code {result.returncode}\n"
+        f"--- stdout ---\n{result.stdout}"
+        f"--- stderr ---\n{result.stderr}"
+    )
+    assert _strip_comments(result.stdout) == _strip_comments(
+        expected_file.read_text()
+    ), (
+        f"Output of {spy_file.name} differs from {expected_file.relative_to(EXAMPLES_DIR)}"
+    )
```
antocuni, line 99: could you use `spy.util.print_diff` to print a readable diff here?

examples/update_expected_output.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+"""Regenerate expected output files for all SPy examples.
+
+Run this from anywhere; it always writes to examples/expected_output/.
+
+    python examples/update_expected_output.py
+
+After running, review the diff and commit the updated files.
+"""
```
antocuni: this works, but I propose a different UI.
What about introducing an option `pytest --update-examples` or similar?

This way:
1. by default, `test_example` checks the output against the expected output
2. with `--update-examples`, it SAVES the new expected output

This has some nice properties like the builtin ability to update only a single test by using pytest standard filtering mechanism.

examples/expected_output/exit_code.txt
```diff
@@ -0,0 +1,3 @@
+argv:
+/home/users/augier3pi/dev/spy/examples/exit_code.spy
+exit code: 88
```
antocuni, line 3: how is it possible that this test passes in CI? Doesn't it print a different path? 🤔

examples/expected_output/vector.txt
```diff
@@ -0,0 +1,100 @@
+0
+1
+4
+9
```
antocuni, line 5: I think it's fine to modify the example to print less numbers here :)

antocuni (commented):
thank you a lot for this!
I generally like the approach and I'm super happy to add this functionality.
I have added some minor comments.

One additional discussion point: have you considered to embed the expected output directly in the example files? Something like this:

```python
def main() -> None:
    print("hello")

# ==== expected output ====
# hello
```

This would make the update logic slightly more complex but it has the advantage of avoiding creating a lot of new tiny files.

I'm not 100% sure it's a good idea though, happy to hear your thoughts about it

examples/expected_output/vector.txt
paugier: Done

examples/expected_output/exit_code.txt
paugier: The tests were not run in the CI because of 

```
 [tool.pytest.ini_options]
testpaths = ["spy"]
```

I added an explicit step in the CI for the examples.

examples/update_expected_output.py
paugier: I did that and I agree it's better!

examples/test_examples.py
paugier: done

examples/expected_output/dict_example.txt
paugier: Done with a regexp (`line = re.sub(r"0x[0-9a-fA-F]+", "<ADDR>", line)`).

examples/expected_output/exit_code.txt
paugier: Note that I didn't change `testpaths = ["spy"]`, meaning that just running pytest in the root directory does not run the examples. It seems to me that it is a reasonable behavior. To run the example tests locally:

```
pytest examples
# or
cd examples
pytest
```

paugier:
> One additional discussion point: have you considered to embed the expected output directly in the example files?

I don't love the idea. I think the advantage (avoiding few small files in `examples/expected_output`) is not that interesting. This is only one file per examples, and I don't see how these files would cause any issue to anyone.

Then I thought about the "advantage" of having the expected output directly in the examples in terms of pedagogy. However, it is not a very nice way to present the output, in particular when the examples will be used in the doc. We could of course trim this part for the doc but it seems to me that it's going to be hacky.

It is also not unreasonable to have in future examples with relatively long and not very interesting output, which is not a problem if the output is not embedded in the source file.

Therefore, I propose that we keep it simple with standard golden files in `examples/expected_output`. Would it be fine for you?

antocuni:
@paugier I was ready to merge, but I got an idea: what do you think about moving the `pytest examples` command into the `Playground tests` workflow (and maybe rename it to `Playground and example tests`)?

That workflow usually finishes much earlier than the main one, so it should make everything a bit faster.

paugier:
> @paugier I was ready to merge, but I got an idea: what do you think about moving the `pytest examples` command into the `Playground tests` workflow (and maybe rename it to `Playground and example tests`)?
> 
> That workflow usually finishes much earlier than the main one, so it should make everything a bit faster.

Done. With #516, it should be all green!

antocuni:
@paugier there seem to be a failure after merging this (although I admit I didn't investigate whether it's actually related to this PR):
https://github.com/spylang/spy/actions/runs/25760377162/job/75659662735

PR #484, opened 2026-04-24
Demo web without PyScript and JS + improve and fix jsffi

paugier:
Fixes #483 and co-authored with Claude.

Fixes #492

I didn't clean the history (commits). Please tell me @antocuni if you think it is useful.

**Description produced with Claude:**

This PR (fixes #483) adds the primitives needed to drive 60fps canvas animations from SPy/WebAssembly without relying on PyScript, Pyodide, or any hand-written JavaScript. The changes span the C/WASM runtime (`libspy`), the SPy Python-level module (`jsffi.py`), new tests, and two canvas demo applications.

### 1. New `JsVal` tagged-union type (C side)

The old design passed JavaScript values around purely as `JsRef` (an opaque integer handle into a JS object table). The new design introduces a `JsVal` struct — a tagged union encoding a value along with its type tag (`JSREF`, `F64`, `I32`, `STR`, `BOOL`, `FUNCPTR`). This avoids unnecessary JS object allocations for primitive arguments. Each `JsVal` is split into `(int32_t tag, double payload)` when crossing the WASM→JS boundary via `EM_JS`.

### 2. `jsffi_call_method.h/.c` — variadic method calls (0–6 args)

Previously only `call_method_1` existed. A new header/source pair generates `jsffi_call_method_0` through `jsffi_call_method_6`, all taking `JsVal`-based arguments. The C file is `#include`-d directly into `jsffi.c` so all `EM_JS` functions share the same translation unit and can access the `jsffi` global object set up by `jsffi_init()`.

### 3. `jsffi.c` — extended JS runtime object table and new primitives

- `jsffi_init()` now pre-populates well-known singleton values: `document` (id 2), `undefined` (3), `null` (4), `true` (5), `false` (6).
- `jsffi.from_jsval(tag, val)` is added to decode `JsVal` pairs on the JS side.
- New WASM exports: `jsffi_f64`, `jsffi_drop_ref`, `jsffi_setattr` (now takes a `JsVal` pair instead of a `JsRef`), `jsffi_u8array_from_ptr`, `jsffi_new_ImageData`, `jsffi_to_i32`, `jsffi_to_f64`, `jsffi_request_animation_frame`, `jsffi_debug_n_jsrefs`.
- `jsffi_wrap_func` is removed and replaced by `JsVal`-based function pointer passing.
- `$wasmMemory` is added to `EM_JS_DEPS` (needed for `Uint8ClampedArray` from a raw WASM pointer).

[...]

PR #488, opened 2026-04-25
Fix var NULL ptr global and the tests

paugier:
@antocuni, this is a very simple PR just to fix `var BUF: gc_ptr[u8] = gc_ptr[u8].NULL` at the global level.

The diff was given by @antocuni in https://github.com/spylang/spy/pull/484#issuecomment-4317123524 and I just modified two existing tests.

It seems to me that it is not very useful to have a constant null global pointer so I added `var` to the related tests, so that they fail.

Is it correct?


paugier:
This one is another very small and simple PR. Just slightly modify a test to get the SPy bug and the fix, about a quite evident logic bug (diff by @antocuni).

antocuni (approved):

PR #489, opened 2026-04-27
fix bug setattr register

paugier:
This is a 1-character-long fix with fixes a typo leading to a strange bug for this code:

```python
from unsafe import gc_alloc

@struct
class Point:
    x: i32
    y: i32

def main() -> None:
    ptr_point = gc_alloc[Point](1)
    setattr(ptr_point[0], "x", 1)

    point = Point(1, 2)
    print(getattr(point, "x"))
```

### DIFF
--- a/spy/tests/compiler/test_builtins.py
+++ b/spy/tests/compiler/test_builtins.py
@@ -262,6 +262,24 @@ class TestBuiltins(CompilerTest):
         )
         self.compile_raises(src, "foo", errors)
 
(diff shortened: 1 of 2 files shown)


antocuni:
I'll be happy to merge it but it needs an accompanying test

paugier:
> I'll be happy to merge it but it needs an accompanying test

Here it is :slightly_smiling_face:

antocuni (approved):

Issue #494, opened 2026-04-28
Improve documentation: tested examples, tutorial structure, and LLM reference file


paugier:
No action needed from maintainers.

I'd like to work on SPy's documentation, in particular before a presentation I will give on SPy on June.

I have now a detailed plan about this work. I open this issue mostly to declare that I will work on this and to present the plan to potentially get feedback. Howerver, no need to respond unless you have concerns.

---

**Text generated with Claude:**

## Motivation

There are two related goals:

**1. Better human documentation**

The current documentation is very sparse, with no tutorial or smooth progressive introduction to the language. Someone coming to SPy for the first time has no guided path — no ordered sequence of concepts to follow from "hello world" to the blue/red model. The playground exists for interactive experimentation, but there is no written narrative to accompany it.

The goal of this work is to produce something that a human can follow in order to learn SPy: a tutorial-style documentation with source code and command output shown side by side, where the output is actually executed as part of the doc build rather than typed by hand. This means the documentation stays in sync with the language automatically — broken examples fail CI rather than silently misleading readers.

**2. A reference file for LLM-assisted development: `SPY_REFERENCE.md`**

The existing `CLAUDE.md` in the repo is aimed at contributors working *on* SPy itself (code style, test commands, repo conventions). There is no equivalent document for people who want to *develop with* SPy — writing `.spy` programs, using the CLI, understanding the type system and the blue/red model.

This matters concretely for tasks like #490 (implementing `random.random()`), where the natural workflow is to give an LLM enough context about SPy's language semantics and stdlib conventions to implement or iterate on a feature. Right now there is no good artifact to use as that context.

The proposal is to auto-generate a `SPY_REFERENCE.md` file as a build artifact from the same documentation sources. It would be a single flat Markdown file — covering the CLI, type system, blue/red model, and stdlib — that can be dropped directly into an LLM context window. Being generated rather than hand-maintained, it cannot drift from the actual examples. It would also be served as part of the documentation site itself, so it is easy for anyone to find and use.

## Proposed approach

The core idea is that **`.spy` files in `examples/` are the source of truth**. Documentation, tests, and `SPY_REFERENCE.md` all derive from them — nothing is maintained separately.

### 1. Rationalize `examples/`

[...]

PR #495, opened 2026-04-28
Add Documentation Previews to PRs

JeffersGlass:
Adds 3 workflows to automatically build a preview copy of the documentation whenever the docs are touched in a PR; this should make it more comfortable to review documentation PRs (epecially since it seems like there's excitement around the docs, aka #494). It hopefully will help us catch other docs regressions sooner as well, like what happened with #470.

You can see an example of what this looks like [over on my fork](https://github.com/JeffersGlass/spy/pull/7). The action only ever creates one 'sticky' comment in the PR with a link to the docs preview, which gets edited. That way the PR thread isn't cluttered with bot comments.

When the PR is closed, the cleanup action deletes its files and edits the comment one last time to clear the link. Yes the files live in git history forever, but given how they're going to be 98% identical to other files, the addition to clone size should be negligible. And if it really does get bit in the long term, we could do some [`git-filter-repo` kung-foo](https://stackoverflow.com/questions/10067848/remove-folder-and-its-contents-from-git-githubs-history) to actually remove the old previews.

### DIFF
--- a/.github/workflows/docs-preview-build.yml
+++ b/.github/workflows/docs-preview-build.yml
@@ -1,0 +1,53 @@
+name: Build PR Docs Preview
+
+# Splitting the preview docs build into an (untrusted) build step (this one)
+# and a trusted deployment step helps prevent a class of issues that allow 
+# secret exfiltration from a repo.
+# See https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ 
+permissions:
+  contents: read
+
+# When a new udpate is made to this PR, cancel any still-running builds from previous pushes
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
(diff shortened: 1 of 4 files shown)


antocuni (approved):
that's incredibly cool and useful! :)
Thanks!

antocuni:
@JeffersGlass something failed after merging this:
https://github.com/spylang/spy/actions/runs/25082650567

not sure whether it's just a one off failure which we can ignore or something which will happen again.

JeffersGlass:
@antocuni hmmmm yes, the cleanup workflow failed when tried to edit the sticky post on *this* PR…. because this PR doesn’t have a sticky post. And neither will any PR that doesn’t touch the docs...

On the one hand, it’s a harmless failure, so feel free to keep closing/merging other PRs. But of course having a failing action on *most* merges is bad - I’ll find a way to fix it :thinking:

Issue #457, opened 2026-04-07
Add the subcommand "spy fmt"


antocuni:
I think that spy should have a `spy fmt` subcommand to format one or more files, similar to `go fmt` or `cargo fmt`.

Ideally, it should be a wrapper around ruff, but it needs to be "smart".

SPy syntax is not 100% compatible with Python. In particular, currently we have `var` and `const`, but we might decide to introduce more differences in the future (we try not to, but we will if it makes sense):
```python
var x: int = 0
const y: int = 1
```

`ruff` (or any other formatter) fails to parse those and thus cannot reliably format .spy files.

SPy has already support for "fixing" spy files in order to be parsed by the CPython parser, see `magic_py_parse.py`:
https://github.com/spylang/spy/blob/1c6ad6bdb59541c44b43a55830b2ff6b8c03427c/spy/magic_py_parse.py#L1-L30

I think that `spy fmt` should:
1. run `magic_py_parse.preprocess` on the .spy file to get a "fixed" file
2. run ruff on the "fixed" file
3. "undo" the fixes at point (1) to reintroduce `var` and `const` in the formatted source

Point (3) is the hard part.

kanin-kearpimy:
Hi, 

I'm taking care of this.

JeffersGlass:
With #496 merged, closing this as completed.

PR #496, opened 2026-04-29
Add the subcommand "spy format"

kanin-kearpimy:
**Issue:** https://github.com/spylang/spy/issues/457

## Description
Hello, this PR introduce SPy formatter via `spy fmt` command. SPy actually extends Python syntax with extra language grammar. Technically, we can use Python formatter. However, we need to handle SPy-specific `const` and `var` declaration.

`spy format` is calling:

1) preprocess `.spy` code to `python.
2) in `preprocess`, we mark SPy grammar (e.g. `var x`, `const     y` with MIDDLE DOT like `var·x` and `const·y`.
3) Ruff formatting
4) reintroduce SPy grammar by converting `var·x` and `const·y` to `var x` and `const y`

## Note
This PR cover only **single file formatter**. Adding **multiple files** or **directory** are not that hard. To concentrate only on SPy formatter, I propose we leave those features for another issue (I'll create it later).  Such issue is also good for first time contributor.

## DEPRECATED DO NOT READ
`spy fmt` is calling:

1) Read `.spy` file content and tokenize.
2) Perform **scoping tracker** SPy-specific declaration `const` and `var` << (Please see below for more detail)
3) Cleaning SPy-specific `const` and `var` out.
4) Parse (3) to Python tokens.
5) perform `ruff` formatter.
6) Reintroduce SPy-specific declaration back.
7) change file content.

## SPy-specific declaration: Scoping Tracker
image

We scope source code by function, class, and some control-flow branch such as `if`, `else`, `while`, `for`. Tracker push/pop those scope into stack. These scopes are not affected by `ruff` formatter. So, we can track nth-variable by each scope.

In picture above, we have two scope: `module_def-main` and `module_def-main_if-1`. So, dictionary tracker has
- `module_def-main__x-1`
- `module_def-main_if-1__y-1`

If we found SPy-specific like `const`, `var`, we marked them in dictionary. When we perform step (4) to reintroduce SPy-specific, we perform scope tracker against step (1) dictionary. 

Please note we also track *occurrence* of variables and keywords (`if`, `else`, `while`, `for`) against its scope.


spy/cli/commands/format.py
```diff
@@ -0,0 +1,28 @@
...
+@dataclass
+class Fmt_Args(Base_Args):
+    filename: Path = Argument(help="", show_default=False)
+
+
+async def fmt(args: Fmt_Args) -> None:
+    """Format SPy file or directory"""
+    if args.filename.suffix != ".spy":
+        print(
+            f"Error: {args.filename} is not a .spy file or directory.",
```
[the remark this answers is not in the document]
kanin-kearpimy: Leave it as another ticket.

spy/magic_py_parse.py
```diff
@@ -81,6 +81,89 @@ def get_tokens(src: str) -> list[TokenInfo]:
...
+
+        if tokens[i].type == NAME:
+            prefix = f"{'_'.join(scope_stack)}__{tokens[i].string}"
+            occurrence = occurrences.get(prefix, 1)
+            occurrences[prefix] = occurrence + 1
+
+            identifier = _make_identifier(scope_stack, tokens[i].string, occurrence)
+            if identifier in spy_grammar_tracker:
+                tokens.insert(i, spy_grammar_tracker[identifier])
+
```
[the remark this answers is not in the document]
kanin-kearpimy: Good.

spy/cli/cli.py
```diff
@@ -30,3 +31,4 @@
 app.spy_command(imports, name="imports")
 app.spy_command(symtable, name="symtable")
 app.spy_command(cleanup, name="cleanup")
+app.spy_command(fmt, name="fmt")
```
JeffersGlass:
```suggestion
app.spy_command(fmt, name="format | fmt")
```

spy/magic_py_parse.py
```diff
@@ -81,6 +81,127 @@ def get_tokens(src: str) -> list[TokenInfo]:
...
+            identifier = _make_identifier(scope_stack, tokens[i].string, occurrence)
+            if (
+                i > 0
+                and tokens[i - 1].string in ("const", "var")
+                and identifier not in spy_grammar_tracker
+            ):
+                string = tokens[i - 1].string
+                spy_grammar_tracker[identifier] = tokens[i - 1]._replace(
+                    string=string + " "
```
JeffersGlass: Why is the trailing space necessary here?

spy/magic_py_parse.py
```diff
@@ -81,6 +81,127 @@ def get_tokens(src: str) -> list[TokenInfo]:
...
+                    t = tokens[j]
+                    tokens[j] = t._replace(
+                        start=(line_no, t.start[1] + offset),
+                        end=(line_no, t.end[1] + offset),
+                        line=new_line,
+                    )
+                    j += 1
+
+        i += 1
+    # print(spy_grammar_tracker)
```
JeffersGlass: Leftover debugging comment =)

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,63 @@
...
+    construct_SPy_specific_grammar,
+    preprocess,
+    reinsert_spy_specific_grammar,
+)
+from spy.parser import Parser
+from spy.vendored import untokenize
+from spy.vm.vm import SPyVM
+
+
+def format_python_source_with_ruff(source: str) -> str:
```
JeffersGlass: It would be nice to show the results of `ruff format` on success, .e.g. `2 file reformatted, 3 file left unchanged`. Maybe this function could return a [str, str] tuple of source and stdout?

JeffersGlass (approved):
HI @kanin-kearpimy! This is really cool! A few little comments dropped in.

My biggest though is that I think this feature should be called `format`, not `fmt`, and just have `fmt` be an alias in the CLI. I think that will make it more clear across the code what the intent is. So, `fmt.py` -> `format.py`, `test_fmt.py` -> `test_format.py`, etc.

spy/cli/cli.py
kanin-kearpimy: Oh right! Thank man! Renamed it here.

spy/magic_py_parse.py
kanin-kearpimy: Yeah, at first it's for variable offset. Without space, `var x` become `varx` after formatting. 

I didn't go deep into this yet. What do you think?

spy/magic_py_parse.py
kanin-kearpimy: A very good catch. Thank you!

spy/tool/spyformatter.py
kanin-kearpimy: Ohhh, You're right. Let me see this improvement this weekend.

antocuni (commented):
@kanin-kearpimy sorry for taking so long to review this, the last month has been crazy.
So, first of all: congratulations for having invented this way to solve the problem. It's very clever and I didn't think about it.

That said, I think it's _too much_ clever and I am a bit worried about the extra complexity added to `magic_py_parse`.

While reviewing this, I got a different idea which would make `spy fmt` much easier to implement. I prototyped this idea in #540.

Basically, instead of "stripping" var/const and keeping track of `varkind_locs` in a separate table, we can replace `var x` with a single identifier `var·x`, using the special char MIDDLE DOT.

If we have that, `spy fmt` would become as easy as:
1. use magic_py_parse to produce the "pythonized" source
2. use a regex to replace `var·····x` with `var·x`
3. run `ruff fmt` on it
4. use a regex to replace `var·x` back to `var x`
The nice part is that `len('var x') == len('var·x')`, so `ruff fmt` should automatically do the right thing w.r.t. e.g. long lines.

What do you think?

kanin-kearpimy:
Hi, I refactored code to use new `preprocess`.

@antocuni

JeffersGlass (commented):
This might be a nit, but I would love to see the filenames and function names all changed from `fmt` to `format`. (`fmt.py` -> `format.py`, `test_fmt.py` -> `test_format.py`, etc.). There's nothing really gained by the abbreviation here and it would be nice to be consistent with the rest of the command filenames and functions.

Other than that, LGTM!

spy/tool/spyformatter.py
antocuni: it seems that this improvement has been lost during the refactoring?
Re-opening it.

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,59 @@
+import os
```
antocuni, line 1: I don't think that this file belongs to `spy/analyze`.
Let's move it to `spy/tool/spyformatter.py`.

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,59 @@
...
+from spy.magic_py_parse import (
+    preprocess,
+    reintroduce_spy_grammar,
+)
+from spy.parser import Parser
+from spy.vm.vm import SPyVM
+
+
+def format_python_source_with_ruff(source: str) -> str:
+    repo_root = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
```
antocuni: you can use `import spy; spy.ROOT` to get the repo ROOT.
However, I'm not sure it's needed, see below.

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,59 @@
...
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            f.write(source)
+        try:
+            subprocess.run(
+                ["ruff", "format", tmp_path],
+                check=True,
+                capture_output=True,
+                text=True,
+                cwd=repo_root,
```
antocuni: why do you need `cwd=repo_root` here?

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,59 @@
...
+            )
+        except FileNotFoundError as exc:
+            raise RuntimeError("ruff executable not found") from exc
+        except subprocess.CalledProcessError as exc:
+            message = exc.stderr.strip() or exc.stdout.strip() or str(exc)
+            raise RuntimeError(f"ruff format failed: {message}") from exc
```
antocuni: currently if you run `spy format` on a system where `ruff` is not installed, you get this which is a bit ugly:

```
[...]
  File "/home/antocuni/anaconda/spy/spy/cli/_runners.py", line 46, in _run_command
    return await user_func(args)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/antocuni/anaconda/spy/spy/cli/commands/fmt.py", line 29, in fmt
    formatter.format(args.filename.stem)
  File "/home/antocuni/anaconda/spy/spy/analyze/fmt.py", line 56, in format
    formatted_py_src = format_python_source_with_ruff(py_src)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/antocuni/anaconda/spy/spy/analyze/fmt.py", line 29, in format_python_source_with_ruff
    raise RuntimeError("ruff executable not found") from exc
RuntimeError: ruff executable not found
```

I think it's better to raise a `SPyError("W_Exception", message)` so that it's automatically caught and displayed by cli/_runners/_run_command.

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,59 @@
...
+        if os.path.exists(tmp_path):
+            os.unlink(tmp_path)
+
+
+class SPyFormatter:
+    def __init__(self, vm: "SPyVM") -> None:
+        self.vm = vm
+
+    def format(self, modname: str) -> None:
+        spyfile = self.vm.find_file_on_path(modname)
```
antocuni: `modname` looks wrong here. I think it should take a filename and bypass `find_file_on_path`

spy/tool/spyformatter.py
```diff
@@ -0,0 +1,59 @@
...
+        if spyfile is None:
+            return
+        if not spyfile.exists() or not spyfile.isfile():
+            return
+        parser = Parser.from_filename(str(spyfile))
+        parser.parse()
+
+        spy_src = parser.src
+        py_src = preprocess(spy_src, parser.filename)
```
antocuni: this dance looks a bit complicated. You create a parser, parse the file, *discard the result*, then call `preprocess` on the original source code.
I think that this logic can be greatly simplified by doing something like this:
```
def format(self, filename: Path):
    spy_src = filename.read()
    py_src = preprocess(spy_src, filename)
```

spy/cli/commands/format.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@dataclass
+class Fmt_Args(Base_Args):
+    filename: Path = Argument(help="", show_default=False)
+
+
+async def fmt(args: Fmt_Args) -> None:
+    """Format SPy file"""
+    # TODO: implement directory formatting.
```
antocuni, line 20: it would be nice to have support for
1. directoris, e.g. `spy format /tmp/somedir`)
2. multiple files, e.g. `spy format /tmp/*.spy`)

But I think it's fine to do this in a follow-up PR :)

spy/cli/commands/format.py
```diff
@@ -0,0 +1,29 @@
+import sys
```
antocuni, line 1: I agree with @JeffersGlass that this file should be called `format.py`, for consistency with all the rest

spy/tests/test_cli.py
```diff
@@ -345,6 +345,23 @@ def test_imports(self):
         _, stdout = self.run("imports", self.main_spy)
         assert stdout.startswith("Import tree:")
 
+    def test_fmt_file(self):
+        mod = textwrap.dedent(
+            """def main( ) -> None:
+                print( \"hello world\" )
+            """
+        )
+        f = self.write("ugly.spy", mod)
```
antocuni:
```suggestion
        src = """
        def main( ) -> None:
            print( "hello world" )
        """
        src = textwrap.dedent(src)
        f = self.write("ugly.spy", src)
```

Some changes here:
1. it's better to call it `src`, as in all the other tests
2. when you pass a multiline string to a function, the indentation becomes very ugly. Better to define a `src = """...` variable first, then use it later. This is the preferred style nowadays (although some old tests still use the old style).
3. You don't need `\"` inside the multiline string, you can use `"` directly

spy/tests/test_cli.py
```diff
@@ -345,6 +345,23 @@ def test_imports(self):
...
+
+        self.run("format", f)
+        expected = textwrap.dedent(
+            """def main() -> None:
+    print("hello world")
+            """
+        )
+
```
antocuni, line 357: same comment as above about using a `src` variable

spy/magic_py_parse.py
```diff
@@ -63,6 +66,14 @@ def get_tokens(src: str) -> list[TokenInfo]:
     return list(tokenize(readline))
 
 
+def reintroduce_spy_grammar(src: str) -> str:
+    """
+    Convert merged SPy declarations like `var·x` / `const·name` back to
+    `var x` / `const name`.
+    """
+    return SPY_DECL_RE.sub(r"\1 \2", src)
```
antocuni:
```suggestion
def undo_preprocess(src: str) -> str:
    """
    The opposite of preprocess.
    
    If preprocess converts e.g. `var x` into `var·x`, this does the opposite.  
    Used e.g. by 'spy format' to reconstruct the user code after passing it to ruff.
    """
    return SPY_DECL_RE.sub(r"\1 \2", src)
```

antocuni (commented):
thanks @kanin-kearpimy .
This is much easier and simpler than the previous approach, good job :).

I left some comments: many are style issues and nitpicks, but two are real:

1. `spy format` should operate on filenames, not modnames
2. it should raise `SPyError` for messages which are meant to be shown to the user, not `RuntimeError`.

kanin-kearpimy:
```
it seems that this improvement has been lost during the refactoring?
Re-opening it.
```

What do you mean here sorry? @antocuni

antocuni:
> ```
> it seems that this improvement has been lost during the refactoring?
> Re-opening it.
> ```
> 
> What do you mean here sorry? @antocuni

the original comment asked to return `[str, str]`, it was marked as resolved but it still returns `str`.

paugier:
Side remark: it would be very convenient if `spy format` could format result of `spy redshift`. I'm wondering if a trick like `var·x` could be used for that. Redshift output contains names like "`_print::println[str]`", which could be turned into equivalent with other special uncommon characters. I will test this idea with a dedicated Python script.

antocuni:
> Side remark: it would be very convenient if `spy format` could format result of `spy redshift`. I'm wondering if a trick like `var·x` could be used for that. Redshift output contains names like "`_print::println[str]`", which could be turned into equivalent with other special uncommon characters. I will test this idea with a dedicated Python script.

that's an interesting idea.

Not 100% it would make the code more readable though, because redshifted code tends to have very long identifiers. Maybe we can make it working by using a longer line length than usual?

spy/tool/spyformatter.py
kanin-kearpimy: Okay. Edited!

spy/tool/spyformatter.py
kanin-kearpimy: It's for config in `pyproject.toml`. However, I adjusted it a bit to explicitly adding config file on command.

```
            cmd = ["ruff", "format", tmp_path]
            if os.path.exists(config_path):
                cmd.extend(["--config", config_path])
            result = subprocess.run(
                cmd,
                check=True,
                capture_output=True,
                text=True,
            )
```

spy/tests/test_cli.py
kanin-kearpimy: Thank you for this insight!

spy/tests/test_cli.py
kanin-kearpimy: # Update

I got it! I miss `/`. 

```
        expected = """\
        def main() -> None:
            print("hello world")
        """
```
OMGG

---
# Deprecated

Seem like my `expected` get affected by indent regardless of `dedent` function. So, I have to ugly do it like below.

```
        expected = """def main() -> None:
    print("hello world")
        """
        expected = textwrap.dedent(expected)
        assert file.read() == expected
```



when I beautify it as

```
        expected = """
        def main() -> None:
            print("hello world")
        """
```

it get error

```
E       assert 'def main() -> None:\n    print("hello world")\n' == '\ndef main() -> None:\n    print("hello world")\n'
E         
E         - 
E           def main() -> None:
E               print("hello world")

spy/tests/test_cli.py:363: AssertionError
```

Not sure why... still working on it.

spy/cli/commands/format.py
kanin-kearpimy: Of course. Let me share new PR with you soon.

kanin-kearpimy:
Hello @antocuni ,

most of comments has been resolved. For multiple files / dictionary, New PR is coming.

spy/cli/commands/format.py
```diff
@@ -0,0 +1,29 @@
...
+
+async def format(args: Fmt_Args) -> None:
+    """Format SPy file"""
+    # TODO: implement directory formatting.
+    if args.filename.suffix != ".spy":
+        print(
+            f"Error: {args.filename} is not a .spy file or directory.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
antocuni, line 26: not a merge blocker, but a nice improvement.
Both `ruff format` and `black` don't bother checking the extension:
```
❯ ruff format /tmp/a.txt 
1 file left unchanged

❯ black /tmp/a.txt 
All done! ✨ 🍰 ✨
1 file left unchanged.
```

spy/tests/test_format.py
```diff
@@ -0,0 +1,29 @@
+import textwrap
+
+from spy.magic_py_parse import undo_preprocess
```
antocuni: if this function lives in `magic_py_parse`, it should be tested by `test_magic_py_parse`.
I fixed it by cfa78a1b

antocuni (approved):
Looks good!
Sorry for taking so long.

I improved a minor thing and tests are running now.
I'll merge as soon as they are green.

PR #504, opened 2026-05-07
Add the @force_inline decorator

antocuni:
Add a `@force_inline` decorator.

Inlining happens during redshift. At the current moment it is possible to apply the decoator only to functions which are "simple enough": in particular, they must a `return` as the last statement of the body, and cannot have other `return`s anywhere.

Code written by claude with *heavy* supervision and editing by me.


antocuni:
@JeffersGlass there seem to be still problems with the `cleaup-repo` job:
https://github.com/spylang/spy/actions/runs/25559113149/job/75025723104

PR #512, opened 2026-05-11
Docs: Add Pages For SPy Builtins and `__spy__` module

JeffersGlass:
Adds two new pages - `SPy Builtins` and `The __spy__ Module`.

Not every last thing is detailed here - I omitted things that feel 'temporary enough' to be not worth mentioning for now.

- Module `__INIT__()` feels like an implementation detail for builtin types... but may be worth documentating?
- `_stdio_write()` seems transient enough to not be worth documenting yet.
- `is_compiled()` feels like it's just for SPy's tests

Of course, happy to add these if wanted!

### DIFF
--- a/docs/src/api_reference/spy_builtins.md
+++ b/docs/src/api_reference/spy_builtins.md
@@ -1,0 +1,91 @@
+title: SPy Builtins
+
+<style> {
+  font-family: "Lucida Console", "Courier New", monospace;
+}
+</style>
+
+
+
+/// warning
+XXXXXXXXXXXXx
+///
+
+
+## Class and Callable Decorators
+
+### __@struct__
+
+:   Used to create a struct from a class definition. See the [low level memory section on structs](/llmem/#stack-allocated-structs) for more details.
+
+### __@blue__
+:   Declares that a function should be executed at [redshift time](https://antocuni.eu/2025/10/29/inside-spy-part-1-motivations-and-goals/#redshifting); that is, at the point when method and function lookups are resolved, and prior to compilation to C or WASM, if any.
+
+:   `@blue` decorated functions are not required to specify types for their parameters, nor are they required to specify return types.
+
+### __@blue.generic__
+:   Denotes that the decorated function accepts a type as its first argument. Generic functions should return a function object.
+
+:   Generic functions are called using the spy-specific syntax `func[Type](args)`.
+
+```python
+  @blue.generic
+  def add(T):
+      def impl(x:T, y: T) -> T:
+          return x + y
+      return impl
+
+  def main() -> None:
+      print(add[i32](1, 2))
+      print(add[str]('hello ', 'world'))
+```
+
+### __@blue.metafunc__
+:   Unlike `blue.generic` functions, `metafunc`s accept one or more arguments, and return an [OpSpec]() appropriate to those arguments based on their static type.
+
+```python
+  from operator import OpSpec
+
+  @blue.metafunc
+  def myprint(m_x):
+      if m_x.static_type == int:
+          def myprint_int(x: int) -> None:
+              print(x)
+          return OpSpec(myprint_int)
+
+      if m_x.static_type == str:
+          def myprint_str(x: str) -> None:
+              print(x)
+          return OpSpec(myprint_str)
+
+      raise TypeError("don't know how to print this")
+
+  def main() -> None:
+      print(42)
+      myprint("hello")
+      myprint(5.2)  # raises TypeError
+```
+
+### @force_inline
+
+:   Causes the decorated functino to be inlined **during redshifting**. 
+
+Only functions with a single return statement at the end of the funciton can be forced inline. Blue functions cannot be forced inline, nor can forced-inline statements be used recursively.
+
+```py
+#inline_demo.spy
+from __spy__ import force_inline
+
+@force_inline
+def inc(x: i32) -> i32:
+    return x + 1
+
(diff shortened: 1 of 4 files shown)


docs/src/api_reference/spy_builtins.md
```diff
@@ -0,0 +1,115 @@
...
+## Class and Callable Decorators
+
+### __@struct__
+
+:   Used to create a struct from a class definition. See the [low level memory section on structs](../llmem.md#stack-allocated-structs) for more details.
+
+### __@blue__
+:   Declares that a function should be executed at [redshift time](https://antocuni.eu/2025/10/29/inside-spy-part-1-motivations-and-goals/#redshifting); that is, at the point when method and function lookups are resolved, and prior to compilation to C or WASM, if any.
+
+:   `@blue` decorated functions are not required to specify types for their parameters, nor are they required to specify return types.
```
antocuni:
```suggestion
:   type annotations for `@blue` functions are optional. If omitted, the arguments and return types default to `dynamic`.
```

docs/src/api_reference/spy_builtins.md
```diff
@@ -0,0 +1,115 @@
...
+    for i in range(terms):
+        sum += 1/factorial(i)
+    return sum
+
+def main() -> None:
+    print(calc_e(10))
+```
+
+### __@blue.generic__
+:   Denotes that the decorated function accepts a type as its first argument. Generic functions should return a function object.
```
antocuni: This is wrong, they accept arbitrary arguments.
The only difference between `@blue` and `@blue.generic` is that the latter is called via `[]` instead of `()`.

docs/src/api_reference/spy_builtins.md
```diff
@@ -0,0 +1,115 @@
...
+
+:   Generic functions are called using the spy-specific syntax `func[type](args)`.
```
antocuni: this is also semi wrong. Potentially a `@blue.generic` function might return something which is not a function, in that case you cannot call like that.

At the same time, the use case that you are document *is* the most important use case for blue.generic functions, so it's still worth mentioning.

docs/src/api_reference/spy_module.md
```diff
@@ -0,0 +1,30 @@
...
+</style>
+
+The `__spy__` module provides functions for introspecting and manipulating SPy objects' at colors and types at runtime. It also contains some objects - `interp_list`, `interp_dict`, and `interp_tuple` -  that act as fallback implementations for those data structures that only function within the interpreter.
+
+/// warning
+The contents of the `__spy__` module and SPy's builtins form the API surface of a language that's rapidly evolving. All of the constructs, names, functions, or decorators here are likely to change!
+///
+
+### __COLOR__(object) -> Literal["red", "blue"] { #markdown data-toc-label='Color()' }
+:   Returns the current color of the passed object. Mainly useful in tests, but may be useful to give users a view into the color of objects during development.
```
antocuni: `COLOR` returns the color of an *expression*, not an object (object themselves don't have a color).
Take e.g. this example:
```
from __spy__ import COLOR

def main() -> None:
    a = 'hello'
    print(COLOR(a))
    var b = a
    print(COLOR(b))
```

`a` is blue, `b` is red, although they both point to the same object.

docs/src/api_reference/spy_module.md
```diff
@@ -0,0 +1,30 @@
...
+
+### __as_red__(object) -> Literal["red", "blue"] { #markdown data-toc-label='as_red()' }
+:   Returns a copy of the object as a red object. May be useful during metaprograming ensure that blue objects which are equal do not get optimized into the same object at redshift time. See, for example, the [implementation of `exal_expr_List`](https://github.com/spylang/spy/blob/main/spy/vm/astframe.py#L1161).
```
antocuni: it's a copy only for value types; for reference types (e.g. strings) it's basically an identity function (but it changes the color)

docs/src/api_reference/spy_module.md
```diff
@@ -0,0 +1,30 @@
...
+
+### __STATIC_TYPE__(object) { #markdown data-toc-label='\_\_STATIC_TYPE\_\_()' }
+:   Returns the type of the object determined in a static context. Useful in tests, and is used in some of SPy's internal machinery.
```
antocuni: again, it's the type of the *expression*, not the object

antocuni (changes requested):
thank you for this.
It's good, but see my comments.

As a meta note, I'm not sure that the _explanation_ of what is a blue function belongs to an "API reference" document.
E.g. if we adopted [Diataxis](https://diataxis.fr/), it would belong to the "Explanation" category.

Still better to have it here than nowhere else, though :).

About `is_compiled()`: I think it should be documented as it's useful.
E.g. I introduced it because in some benchmark I wanted to be able to change the number of iterations depending on `is_compiled()` (because the interpreter is too slow)

JeffersGlass:
> As a meta note, I'm not sure that the explanation of what is a blue function belongs to an "API reference" document.
E.g. if we adopted [Diataxis](https://diataxis.fr/), it would belong to the "Explanation" category.

> Still better to have it here than nowhere else, though :).

This was my thinking exactly =). Over time, these "API references" should get shorter and tighter, and be able to reference the more conceptual and tutorial parts of the docs, as they get written.

Thank you for all the comments! Working through them now.

antocuni (approved):
LGTM, thanks!

PR #516, opened 2026-05-12
pixi lock for pygments 2.19.1

paugier:
Should fix the Pixi workflow.

The pyproject.toml was modified (dep. pygments 2.19.1), so we need to relock the Pixi environment.


antocuni:
thank you!
FWIW, the current `[dependencies]` section in pyproject.toml is basically a poor's man lockfile.
I wonder whether we should just move to be entirely uv/pixi based and use proper lockfiles.

PR #519, opened 2026-05-14
Add type.__origin__ and function.__origin__

antocuni:
Add `__origin__` to types and functions. This is used to keep track of which `@blue.generic` function a concrete object come from.

For example:

```python
@blue.generic
class MyList(T):
    @struct
    class impl:
         ...

    return impl


assert MyList[T].__origin__ is MyList
```

This becomes even more useful as soon as we merge #448 because then we can write `class list[T]: ...` but we still have the notion of "this type is a list".

The `__origin__` is set ONLY:
1. upon returning something from a `@blue.generic` function
2. if the returned value was defined INSIDE the function
3. if the `__origin__` has not been set already

This means that e.g.:
```python
@blue.generic
def foo(T):
    return list[T]

# origin was already set, don't overwrite it
assert foo[T].__origin__ is list 
```

And:
```python
@struct
class Point:
    ...

@blue.generic
def bar(T):
    return Point

# don't set the origin because Point was not defined inside bar
assert Point.__origin__ is None
```

This is also the groundwork for a subsequent PR to display "better" FQNs: e.g. we will declare that `_list::list[i32]::_ListImpl` can be shown as `list[i32]`.

### DIFF
--- a/spy/doppler.py
+++ b/spy/doppler.py
@@ -58,7 +58,7 @@ def make_const(vm: "SPyVM", loc: Loc, w_val: W_Object) -> ast.Expr:
         items = [make_const(vm, loc, w_item) for w_item in w_val.items_w]
         res = ast.Tuple(loc, items, w_T=w_T)
 
-    elif w_T.fqn.match("_tuple::tuple[*]::_tup"):
+    elif vm.is_tuple_type(w_T):
         # transform the struct into a syntactical ast.Tuple node, so that we can put it
         # in the AST without necessarily create a FQN
         assert isinstance(w_val, W_Struct)
--- a/spy/tests/compiler/test_basic.py
+++ b/spy/tests/compiler/test_basic.py
@@ -956,6 +956,75 @@ class TestBasic(CompilerTest):
         assert mod.foo() == 3
         assert mod.bar() == "hello world"
 
+    @only_interp
+    def test_generic_type_origin(self):
+        mod = self.compile("""
+        @blue.generic
+        def MyType(T):
+            @struct
+            class _impl:
+                data: T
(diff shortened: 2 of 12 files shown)


JeffersGlass:
Just thinking about docs for later, would it be accurate to say (when this PR is complete :smile: )

> If a generic function returns a class or a function, that class or function has its `__origin__` attribute set to the original generic function.

antocuni:
> Just thinking about docs for later, would it be accurate to say (when this PR is complete 😄 )
> 
> > If a generic function returns a class or a function, that class or function has its `__origin__` attribute set to the original generic function.

I think this is correct but also very hard to read.
I probably deserves a separate section explaining `__origin__`, something like that:

```
# `__origin__`

If a type or a function has `__origin__` set, it means it was created by and returned from a `@blue.generic` function:

@blue.generic
class MyList(T):
    @struct
    class impl:
        ...
     return impl

assert MyList[T].__origin__ is MyList

This makes it possible to write a generic "is this type a list?" check.

Note that this also applies for PEP 695-style declarations:

def add[T](...):
    ...

```

And then if we have this section, in the docs for types/blue.generic/whatever we can explain `__origin__` by linking to that section. I don't fully like the exact wording which I use, but you get the idea.

Probably it makes sense to wait until #448 lands, so we can write examples for both types and functions.

antocuni:
@JeffersGlass I merged this PR, and explained some more details in the PR description. Hopefully will make your life easier to write the doc (thanks for that!)

JeffersGlass:
Thanks @antocuni! The explanation you added is super helpful - once the other PR lands, I can get some docs together on this.

Issue #407, opened 2026-02-27
Add support for the `in` operator and the `__contains__` method


antocuni:
This is a **GOOD FIRST ISSUE** which is meant to be done by humans, as a way to learn the codebase. If I plan to use a code assistant to do all the work, please don't grab this issue and let some human to do it (it is fine to use them to *assist* the work).

We want to be able to write e.g.:
```python
def main() -> None:
    a = [1, 2, 3]
    print(1 in a)
```

this requires:
1. modify the parser to support the `in` operator
2. add the `w_CONTAINS` operator to `vm/modules/operator/binop.py`
3. fix the ASTFrame and/or DopplerFrame as needed
4. add `__contains__` to `str`, `list` and `dict` (and possibly other types where it makes sense).

PR #520, opened 2026-05-14
Add in operator

sbundaz:
Issue: #407 

This PR adds the `in` operator.


JeffersGlass:
In general this looks good to me! A few thoughts:

1. It would be nice to have `_dict.__contains__` and `_tuple.__contains__` for completeness in the stdlib
1. A bunch of the string methods [got moved to App Level](https://github.com/spylang/spy/pull/556) (`stdlib/_str.spy`) earlier this week. For consistency `str.__contains__` may want to move there
1. There isn't a firm story around `Iterator` or `Iterable` yet... in CPython if there's no `__contains__` method it falls back to `__iter__` (SPy's `__fastiter__`?) and then `__getitem__` if they exist. This is probably overkill for now, but maybe a good extension for later.

JeffersGlass:
After a little more testing, I think you'll need to add some additional traps for non-matching types of generic containers.. For instance, this is currently an error:
```py
def main() -> None:
    x = [1,2,3]
    "1" in x
```
```py
TypeError: mismatched types
  | /workspaces/spy/a.spy:4
  |     print("1" in x)
  |           |_| expected `i32`, got `str`

  | /workspaces/spy/stdlib/_list.spy:364
  |         def __contains__(self, value: T) -> bool:
  |         |______________________________________| function defined here

  | /workspaces/spy/a.spy:4
  |     print("1" in x)
  |           |______| operator::IN called here
```
Whereas it probably wants to just return False.

antocuni:
> After a little more testing, I think you'll need to add some additional traps for non-matching types of generic containers.. For instance, this is currently an error:
> 
> ```python
> def main() -> None:
>     x = [1,2,3]
>     "1" in x
> ```
> 
> ```python
> TypeError: mismatched types
>   | /workspaces/spy/a.spy:4
>   |     print("1" in x)
>   |           |_| expected `i32`, got `str`
> 
>   | /workspaces/spy/stdlib/_list.spy:364
>   |         def __contains__(self, value: T) -> bool:
>   |         |______________________________________| function defined here
> 
>   | /workspaces/spy/a.spy:4
>   |     print("1" in x)
>   |           |______| operator::IN called here
> ```
> 
> Whereas it probably wants to just return False.

I'm happy to be convinced otherwise, but I think this should be an error, unless it's list[object] or list[dynamic].

I made the same choice for e.g. ==: `1 == ”hello"` it's a TypeError

antocuni:
@sbundaz sorry it's taking so long to review, I hope to get to it soon!

spy/tests/stdlib/test__list.py
```diff
@@ -799,3 +799,39 @@ def foo() -> list[i32]:
...
+            lst = list[str]()
+            lst.append("hello")
+            lst.append("world")
+            return "foo" in lst
+        """)
+
+        assert mod.test_i32_found() == True
+        assert mod.test_i32_not_found() == False
+        assert mod.test_str_found() == True
+        assert mod.test_str_not_found() == False
```
antocuni: very minor improvement/nitpick, feel free to ignore.
You could write the test to have just `i32_contains` and `str_contains`, like this:
```python
def i32_contains(x: i32) -> bool:
    ...
    lst.append(1)
    lst.append(2)
    lst.append(3)
    return x in lst
```

spy/tests/stdlib/test__range.py
```diff
@@ -98,3 +98,43 @@ def fmt3(a: int, b: int, step: int) -> str:
...
+
+        assert mod.c3(1) == True
+        assert mod.c3(4) == True
+        assert mod.c3(2) == False
+
+        assert mod.c4(10) == True
+        assert mod.c4(5) == True
+        assert mod.c4(0) == False
+        assert mod.c5(8) == True
+        assert mod.c5(7) == False
```
antocuni: same as above; I think `c1`, `c2`, `c3` etc a bit hard to read because you always need to jump up and down to follow the logic.
What about (pseudocode)

```
def range1_contains(x, n):
    return x in range(n)

def range3_contains(x, start, stop, step):
    return x in range(start, stop, step)
```

stdlib/_tuple.spy
```diff
@@ -50,4 +50,44 @@ def tuple(*items_T):
...
+
+                    def check_or_rest(self: _tup, v: T) -> bool:
+                        if getattr(self, attr) == v:
+                            return True
+                        return rest(self, v)
+                    return check_or_rest
+
+                return rest
+
+            return OpSpec(build(0), [m_self, m_v])
```
antocuni, line 108: oh wow, that's very clever and very advanced. I'd even image that such a thing is already possible today :)
If we decide to keep it, it's worth adding a docstring/comment to explain _what_ this code does and what is the result after redshift.

I'm not 100% sure that `contains` for heterogeneous tuple make sense to exist. The way I see it:
  - fixed-size heterogeneous tuples are basically "structs in disguise". If you have a struct `Person(name, age, address)`, does it make sense to say `"Antonio" in p`?
  - variable-size homogeneous tuples like `tuple[int, ...]`. For them, it totally makes sense, but are not supported yet (they will)
  - fixed-size homogeneous tuples like `tuple[int, int, int]`: I'm undecided, but we should probably implement it because it's easy enough and might make some existing python code less surprising.

Can we think of any real-world use case in which we might want `contains` on heterogeneous tuples _AND_ that it's not solvable by using `tuple[dynamic, ...]` (when we will have it)?

Btw, in case we decide to discard it, I think this code should still go in examples/4_advanced` because it's very cool :)

stdlib/_list.spy
```diff
@@ -331,6 +331,17 @@ def list(T):
...
 
+        def __contains__(self, value: T) -> bool:
+            ll = self.__ll__
+            i = 0
+            while i < ll.length:
+                if ll.items[i] == value:
+                    return True
+                i = i + 1
+            
+            return False
```
antocuni, line 342: not a merge blocker but might be an interesting follow-up:
1. we should check whether a modern C compiler is able to automatically vectorize this loop
2. if not, we might want to research some existing SIMD/SSE/AVX/whatever C library to vendor which offers the equivalent of the (non-existing) `memchr32`, `memchr64`, etc.
3. if it doesn't exist, we could at least expose `memchr` and use it for `list[i8]`, `list[u8]`, `bytes` and `str` (for `str` it will be harder in case of "true" unicode, but for now we can ignore it). See e.g. `vm/modules/unsafe/mem.py` where we have memcpy, memcmp, etc.

stdlib/array.spy
```diff
@@ -74,6 +74,15 @@ def _array1(DTYPE):
...
             return self.__ll__.l
 
+        def __contains__(self, v: DTYPE) -> bool:
+            ll = self.__ll__
+            i = 0
+            while i < ll.l:
+                if ll.items[i] == v:
+                    return True
+                i = i + 1
+            return False
```
antocuni, line 84: see the comment above about `memchr`&co. (For the followup)

spy/vm/modules/operator/__init__.py
```diff
@@ -93,6 +93,7 @@
     "is not": OP.w_ISNOT,
     "<universal_eq>": OP.w_UNIVERSAL_EQ,
     "<universal_ne>": OP.w_UNIVERSAL_NE,
+    "in": OP.w_IN,
```
antocuni, line 96: good, but this misses a test in `tests/compiler/operator/test_binop.py`.

spy/vm/str.py
```diff
@@ -117,3 +117,10 @@ def w_NEW(vm: "SPyVM", wam_cls: W_MetaArg, *args_wam: W_MetaArg) -> "W_OpSpec":
     @staticmethod
     def w_str(vm: "SPyVM", w_s: "W_Str") -> "W_Str":
         return w_s
+
+    @builtin_method("__repr__")
+    @staticmethod
+    def w_repr(vm: "SPyVM", w_s: "W_Str") -> "W_Str":
+        assert isinstance(w_s, W_Str)
+        ptr = vm.ll.call("spy_str_repr", w_s.ptr)
+        return W_Str.from_ptr(vm, ptr)
```
antocuni: I'm confused. We already have `_str::methods::__repr__`, why is this needed?

spy/libspy/src/str.c
```diff
@@ -37,6 +37,88 @@ spy_str_eq(spy_StrObject *a, spy_StrObject *b) {
...
+    return res;
+}
+
+int32_t
+spy_str_len(spy_StrObject *s) {
+    return (int32_t)s->length;
+}
+
+spy_StrObject *
+spy_str_repr(spy_StrObject *s) {
```
antocuni: see the related comment in `str.py`. This shouldn't exist.

spy/libspy/src/str.c
```diff
@@ -37,6 +37,88 @@ spy_str_eq(spy_StrObject *a, spy_StrObject *b) {
...
+    }
+    if (i >= l || i < 0) {
+        spy_panic("IndexError", "string index out of bound", __FILE__, __LINE__);
+        return NULL;
+    }
+    spy_StrObject *res = spy_str_alloc(1);
+    char *buf = (char *)spy_StrObject_UTF8(res);
+    buf[0] = spy_StrObject_UTF8(s)[i];
+    return res;
+}
```
antocuni: this is already implemented in `_str.spy`, why do you need it here?

spy/libspy/include/spy/str.h
```diff
@@ -107,6 +107,9 @@ int32_t WASM_EXPORT(spy_str_hash)(spy_StrObject *s);
 #define spy_operator$str_eq spy_str_eq
 #define spy_operator$str_ne spy_str_ne
 #define spy_operator$str_to_complex128 spy_str_to_complex128
+#define spy_builtins$str$replace spy_str_replace
+#define spy_builtins$str$__getitem__ spy_str_getitem
+#define spy_builtins$str$__len__ spy_str_len
```
antocuni: remove (see below)

antocuni (commented):
Sorry for the late review @sbundaz .
The PR looks very good!

I left some comments: some are nitpicks which are not merge blockers, some are potential follow-ups.

One thing which must be solved is to remove the unnecessary code in str.c, which I guess it's leftover from an earlier attempt before we had applevel _str.spy?

Then, we need to decide what to do with `tuple.__contains__`. See the inline comment.

Let's summon @JeffersGlass as he's the one requesting it in the first place, happy to hear the thoughts of you both.

stdlib/_tuple.spy
JeffersGlass: +1 on the implementation for homogeneous tuples, both variable-size and fixed-size, just as a straightforward way to minimize differences from CPython.

On the fixed-size heterogeneous tuples front... I guess I would be +0.5 to have just as another ease-of-transition feature for those coming from CPython.

In all the cases I can imagine, the better answer would be "Change your data structure so you don't have heterogeneous tuples", but that might not be the kindest case for non-greenfield code. Sort of depends how much we want to encourage effective re-writing.

antocuni: > In all the cases I can imagine, the better answer would be "Change your data structure so you don't have heterogeneous tuples", but that might not be the kindest case for non-greenfield code. Sort of depends how much we want to encourage effective re-writing.

Generally speaking, we want to encourage "sane" typing. Not all Python patterns work on SPy and that's fine, it's by design and it's a deliberate choice.  If the patter is not "sanely typable", then we should just reject it and ideally emit a nice error message explaining how to fix (in this case it could be `plese use tuple[dynamic, ...]`... when we have it :)).

It's a similar situation as what to do for `==`. We could easily say that if two types are incompatible, `==` always return False, but I think it's wrong: it smells of early javascript in which the language tries to guess the intent.  The call which I made is that `==` between two incompatible types is a TypeError, UNLESS they are of type `dynamic` (in which case you opt-in for CPython-style behavior).

Now that I think about this, you could think of `__contains__` as a series of `==`: in this case, the more natural choice is to disallow it, for similarity with `==`.

So, +0.9 to implement `__contains__` only for homogeneous tuples, unless someone convinces me otherwise :).

stdlib/_list.spy
sbundaz: I'm not very familiar with these topics, but I can try to do some research after this PR.

spy/vm/str.py
sbundaz: My fault when I merged the changes :/

spy/tests/stdlib/test__tuple.py
```diff
@@ -56,3 +56,60 @@ def bar() -> bool:
...
+    def test_contains_str(self):
+        mod = self.compile("""
+        from _tuple import tuple
+
+        def c(v: str) -> bool:
+            t = tuple[str, str]("hello", "world")
+            return v in t
+        """)
+        assert mod.c("world") == True
+        assert mod.c("foo") == False
```
antocuni, line 89: I think that all of these could be condensed into a single test, as you did for the others.
Not a merge blocker, but feel free to submit a small follow-up if you want.

antocuni (approved):
LGTM, merging soon.
Thanks for doing this!

PR #526, opened 2026-05-18
Turn `@struct` assert into WIP Error 

JeffersGlass:
Small change - since the error that `only @struct classes are permitted` is visible to users, it's nicer to have as an Error rather than a bare assert.

### DIFF
--- a/spy/vm/astframe.py
+++ b/spy/vm/astframe.py
@@ -412,7 +412,12 @@ class AbstractFrame:
         if classdef.kind == "struct":
             return W_StructType
         else:
-            assert False, "only @struct classes are supported for now"
+            raise SPyError.simple(
+                "W_WIP",
+                "only @struct classes are supported for now",
+                "class defined here",
+                classdef.loc,
+            )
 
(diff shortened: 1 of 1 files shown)


paugier:
+1!

Generally, it seems to me that there are many other cases like that in the code base. Good WIP errors are much nicer than obscure asserts for "beta testers" :slightly_smiling_face:

But this might be because I tend to try to do relatively advanced things with SPy!

antocuni:
> +1!
> 
> Generally, it seems to me that there are many other cases like that in the code base. Good WIP errors are much nicer than obscure asserts for "beta testers" 🙂

Agreed, that's why I introduced the `WIP` exception in the first place. Generally speaking, `AssertionError` should mean "bug in the compiler" not "the user did something wrong".

Thanks @JeffersGlass for doing that.
 
> But this might be because I tend to try to do relatively advanced things with SPy!

please continue to do so, and feel free to open a "WIP" PR any time you encounter an `AssertionError` :pray:

PR #530, opened 2026-05-19
Fix dict_iterator

antocuni:
dict_iterator was buggy because it didn't skip empty entries. So far
the tests worked by chance because they were reading empty items whose
key/value was 0 by chance.

The fix uncovers and existing bug in unwrap_dict. Fix it.

### DIFF
--- a/spy/tests/stdlib/test_dict.py
+++ b/spy/tests/stdlib/test_dict.py
@@ -127,20 +127,22 @@ class TestDict(CompilerTest):
         src = """
         from _dict import dict
 
-        def test() -> int:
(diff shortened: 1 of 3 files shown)


dpdani:
No wait, I have a much simpler fix. Though it took me a while to figure it out.

It seems that the literal dictionaries relied on the bug though for some reason.

antocuni:
@dpdani ah sorry I didn't understand you were working on it.
Ok, waiting for it :)

antocuni:
thank you for the better fix and explanation @dpdani :pray:

dpdani:
Of course 👍

PR #531, opened 2026-05-20
fix(parser): support untyped global var/const

### LINKED ISSUE #501: Nicer syntax for `var BUF: gc_ptr[u8] = gc_ptr[u8].NULL`
For code using jsffi, one needs to declare `var` global null pointers.

```python
var PARTICLES: gc_ptr[Particle] = gc_ptr[Particle].NULL
var PARAMS: gc_ptr[Params] = gc_ptr[Params].NULL
var JS_REFS: gc_ptr[JsRef] = gc_ptr[JsRef].NULL
```

This is what is currently supported.

I find this syntax a bit heavy so I'd like to support something nicer.

Maybe just:

```python
var PARTICLES = gc_ptr[Particle].NULL
var PARAMS = gc_ptr[Params].NULL
var JS_REFS = gc_ptr[JsRef].NULL
```

? It seems to me that the type of these null pointers is clear enough and that type annotations are not really useful is this case.

Or alternatively just

```python
var PARTICLES: gc_ptr[Particle] = 0
var PARAMS: gc_ptr[Params] = 0
var JS_REFS: gc_ptr[JsRef] = 0
```

or

```python
from unsafe import Null

var PARTICLES: gc_ptr[Particle] = Null
var PARAMS: gc_ptr[Params] = Null
var JS_REFS: gc_ptr[JsRef] = Null
```

What do you think about this @antocuni?

I see that there is a note related to `from unsafe import Null` in https://github.com/spylang/spy/blob/main/spy/vm/modules/unsafe/ptr.py#L136

This is a small thing I'd like to work after my work on jsffi has settled.

SAY-5:
Fixes #501.

Per @antocuni in the issue: `from_py_stmt_Assign` already returns a `VarDef` for `var x = ...` / `const x = ...`, so `from_py_global_Assign` now uses that directly instead of asserting `Assign`.

### DIFF
--- a/spy/parser.py
+++ b/spy/parser.py
@@ -506,18 +506,22 @@ class Parser:
         return spy.ast.Return(py_node.loc, value)
 
     def from_py_global_Assign(self, py_node: py_ast.Assign) -> spy.ast.VarDef:
-        assign = self.from_py_stmt_Assign(py_node)
-        assert isinstance(assign, spy.ast.Assign)
+        stmt = self.from_py_stmt_Assign(py_node)
+        # `var x = ...` and `const x = ...` are already lowered to VarDef
+        # by from_py_stmt_Assign; only bare `x = ...` returns Assign.
+        if isinstance(stmt, spy.ast.VarDef):
+            return stmt
+        assert isinstance(stmt, spy.ast.Assign)
         assert len(py_node.targets) == 1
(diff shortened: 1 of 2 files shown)


paugier (approved):
These changes seem very reasonable to me.

PR #533, opened 2026-05-23
[Docs] CLI Documentation

JeffersGlass:
Add a page which documents the SPy CLI. Documentation is generated by the `mkdocs-typer2` extension.

[See the current preview build here](https://spylang.github.io/docs/pr-preview/pr-533/reference/cli/).

~I think the formatting of this page is fine but not amazing. I like the optional ["native+pretty" formatting](https://syn54x.github.io/mkdocs-typer2/cli-pretty-native/) option that the plugin provides, but there's currently a bug that omits the possible options for Choices. The current formatting ("legacy") I think is less clear than the tables, but actually has all the information.~

~I think it's still worth dropping in a CLI docs page for now. Once that bug is fixed, a follow-up PR can update the formatting.~

 _Edit: The plugin author fixed the bugs mentioned and I've updated the formatting to reflect that. The proposed formatting is live in the dosc preview build._

Also, propose changing the category `API Reference` to just `Reference`.


JeffersGlass:
Just a note for anyone who hits the same issue: the `pixi` checks were in this PR because the pixi.lock file is out of date... because I used `uv` to add a dependency instead of using pixi, which I'm unfamiliar with. To update that lockfile, I had to [install pixi](https://pixi.prefix.dev/latest/installation/) and then run `pixi install`. @paugier 

edit: Pixi checks are still failing… any recomendations?

JeffersGlass:
… and somehow the build for the new page itself failed in CI, probably something to do with the new dependency being in the wrong spot in ‘pyproject.toml`. Converting this to draft until that’s solved.

paugier:
> Just a note for anyone who hits the same issue: the `pixi` checks were in this PR because the pixi.lock file is out of date... because I used `uv` to add a dependency instead of using pixi, which I'm unfamiliar with. To update that lockfile, I had to [install pixi](https://pixi.prefix.dev/latest/installation/) and then run `pixi install`. @paugier
> 
> edit: Pixi checks are still failing… any recomendations?

I see that the Pixi lock file changes have not been committed, which explains why it still fails.

Note also that in that case, you just need to run `pixi lock` (and not `pixi install`) so that pixi does not create the local environment and download the packages.

JeffersGlass:
Thank you @paugier! I committed the lockfile, and also bumped pixi from `0.67.1 -> 0.69.0`, since apparently the lockfile changed formats in `0.68.0`

Still have build issues in CI, but not locally... the `mkdocs-typer2` extension does seem to be installing properly in CI... will continue to debug.

JeffersGlass:
This is fixed up and ready for review! The relevant page on the preview docs is [reference/cli](https://spylang.github.io/docs/pr-preview/pr-533/reference/cli/).

Of note - added a mechanism to declare that we're in a docs build (`spy.platform.IS_DOCS_BULD`, ~set by an environment variable~ set by detecting whether `mkdocs` is in `sys.modules`) so we don't have to load libspy/llmod just to build the docs.

JeffersGlass:
`mkdocs-typer2` just issued a release that fixes both of the formatting issues mentioned in the above post - updating this PR to include the `native+pretty` formatting, which I think looks much nicer.

syn54x/mkdocs-typer2#30
syn54x/mkdocs-typer2#31

.github/workflows/docs-preview-build.yml
```diff
@@ -42,6 +43,7 @@ jobs:
 
       - name: Build and mkdocs
         run: |
+          # don't load libspy/llmod
```
antocuni: I don't understand what this comment means?

antocuni (approved):
LGTM.
I love the fact that CLI docs are autogenerated.

I cannot merge because of some conflicts.
Feel free to merge when the conflict is solved.

.github/workflows/docs-preview-build.yml
JeffersGlass: Oops - it was orphaned from an earlier commit. I had been using an environment variable to prevent loading llmod when we were just building docs... but then I realized we could just check if `mkdocs` was in `sys.modules` at the time.

PR #536, opened 2026-05-27
[GHA] Preview-docs-deploy should succeed if the docs haven't changed

### LINKED ISSUE #534: [GHA] Docs Preview Action Fails on Empty Commit
The github action that generates docs previews tries to commit an update to the relevant build whenever a new commit is added to that branch. However, if that commit didn't actually change the docs, there is nothing to commit, and so the whole workflow fails.

A harmless failure, but annoying and noisy. I will fix.

JeffersGlass:
Small change - the docs-previews get rebuilt for any new commits in a relevant PR, but if that commit didn't actually cause the outputted docs to change, the `docs-previews-deploy` action would fail when it tried to create a new commit with no changes. Ths should allow that job to succeed, even if creating that commit fails.

Fixes #534

### DIFF
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -72,6 +72,7 @@ jobs:
 
       - name: Commit artifact
         working-directory: /tmp/spy-docs
+        continue-on-error: true
         run: |
           git add pr-preview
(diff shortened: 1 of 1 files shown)


antocuni (approved):
I'm not sure I totally understand what's going on and I don't have energy to investigate, but I trust you on this topic :).

Merging

PR #540, opened 2026-06-01
Simplify magic_py_parse

antocuni:
Greatly simplify `magic_py_parse`. The ultimate goal is to be able to re-use the python parser also with this spy-specific syntax:
```python
var x = 1
const y = 2
```

The old way detected `var x`, replaced with a single `x` and kept a side table with info about whether this particular NAME was by chance a `var/const`, and the location info.

The new approach is much simpler: the code above becomes:
```python
var·x = 1
const·y = 2
```

The `·` is "U+00B7 MIDDLE DOT", and Python considers it a valid char for an identifier. This means that `var·x` is seen as a single identifier, so the python parser parses it correctly, and then the SPy parser detect this special name.

As usual, special care is needed to ensure that the source location of the rewritten file are kept the same as in the original. E.g.:
```
    var x: i32 = 100
    var    y     : i32 = 200
```

becomes:
```
    var·x: i32 = 100
    var····y     : i32 = 200
```

### DIFF
--- a/spy/ast_dump.py
+++ b/spy/ast_dump.py
@@ -85,8 +85,6 @@ class Dumper(TextBuilder):
         name = "py:" + node.__class__.__name__
         fields = list(node.__class__._fields)
         fields = [f for f in fields if f not in self.fields_to_ignore]
-        if isinstance(node, py_ast.Name):
-            fields.append("spy_varkind")
         # Use turquoise text_color to distinguish from blue in colorize command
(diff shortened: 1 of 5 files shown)


antocuni:
@kanin-kearpimy this PR should make it much easier to implement `spy fmt`.
For reader: see also #496 (and [this comment](https://github.com/spylang/spy/pull/496#pullrequestreview-4405195060) and #457 .

/cc @JeffersGlass as he reviewed #496 .

PR #544, opened 2026-06-04
Update examples

paugier:
This is the first step of the work that I'd like to do on examples and documentation (as described in #494).

This PR is only to start with a cleaner state.

Notes on the next steps:

- I tend to think that it would be good to group the examples in directories corresponding to categories. I'm not sure about the categories but it could be something like "end-users", "advanced" and "low-level". An advantage is that it could help to include the examples in the playground.

- I'm not sure about what to do with vector.spy. It does not seem very natural code given what can be currently done with SPy. It might be that it could be just "merged" with myarray.spy.


antocuni:
> This is the first step of the work that I'd like to do on examples and documentation (as described in #494).
> 
> This PR is only to start with a cleaner state.

is the PR ready to review or still WIP?

> Notes on the next steps:
> 
> * I tend to think that it would be good to group the examples in directories corresponding to categories. I'm not sure about the categories but it could be something like "end-users", "advanced" and "low-level". An advantage is that it could help to include the examples in the playground.

+1 for grouping in categories. I don't have any particular opinion on what is the best categorization though.

> * I'm not sure about what to do with vector.spy. It does not seem very natural code given what can be currently done with SPy. It might be that it could be just "merged" with myarray.spy.

+1 for deleting it

examples/README.md
```diff
@@ -0,0 +1,47 @@
+# SPy examples
+
+The main examples are organized into four directories, ordered by complexity:
+
+| Directory | Description |
+|---|---|
+| [`1_high_level/`](1_high_level/) | High-level SPy code that looks close to Python |
+| [`2_blue/`](2_blue/) | Blue functions, metafunctions, and compile-time computation |
```
antocuni: maybe we should call this "metaprogramming" instead of "blue"?

paugier:
> is the PR ready to review or still WIP?

Sorry for that. It was in a review state but then I went further. I think it can be reviewed now.

antocuni (approved):
LGTM.

Thank you for doing that, I think it will help immensely to make spy more approachable by people

antocuni:
@paugier feel free to merge it if/when you are done (I assume you have merging rights? If not, let me know and I'll add you)

paugier:
I added two examples on memcpy and friends and str/bytes/encode/decode.

The next step is to modify the playground to show the different categories of examples. I'm going to have a look at that.

PR #547, opened 2026-06-05
Introduce type.__name__, __qualname__, __fqn__ and __full_fqn__

antocuni:
They are all derived by the fqn, with varying grades of precision. It
is not possible to do a 1:1 mapping to what CPython shows but we tried
to give reasonable results for `__name__` and `__qualname__`.

`__fqn__` and `__full_fqn__` are new and should be preferred.

The easiest way to explain the difference is to look at a concrete example (see also test_basic::test_type_name_attributes):

It is easier to explain with an example:

```python
@struct
class Foo[T]:
    pass

def main() -> None:
    print("__full_fqn__ ", Foo[i32].__full_fqn__)
    print("__fqn__      ", Foo[i32].__fqn__)
    print("__qualname__ ", Foo[i32].__qualname__)
    print("__name__     ", Foo[i32].__name__)
```

```
❯ spy /tmp/x.spy 
__full_fqn__  x::Foo[i32]::Self
__fqn__       x::Foo[i32]
__qualname__  x::Foo[i32]
__name__      Foo[i32]
```

`x:Foo[i32]::Self` is the most complete name used internally. It's what `spy redshift` uses to show types/function names when you pass `__full_fqn__`.

`__fqn__` is the "human readable" version of it. The biggest difference is that it honors generics, so you get `x::Foo[i32]` instead.  `__qualname__` is just an alias to it.

`__name__` is like `__qualname__` but without the module name. It's a bit different than CPython because it includes the qualifiers (e.g. `Foo[i32]` instead of `Foo`) but I think it's good to include them in e.g. error messages.

### DIFF
--- a/examples/expected_output/type_name_and_fqn.txt
+++ b/examples/expected_output/type_name_and_fqn.txt
@@ -1,0 +1,4 @@
+__full_fqn__  type_name_and_fqn::Foo[i32]::Self
+__fqn__       type_name_and_fqn::Foo[i32]
+__qualname__  type_name_and_fqn::Foo[i32]
+__name__      Foo[i32]
--- a/examples/type_name_and_fqn.spy
+++ b/examples/type_name_and_fqn.spy
@@ -1,0 +1,9 @@
+@struct
+class Foo[T]:
+    pass
+
+def main() -> None:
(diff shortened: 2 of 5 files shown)


antocuni:
@JeffersGlass would be nice to have some docs about this if you feel like adding them :)

PR #550, opened 2026-06-08
Implement `str.__getitem__(slice)`

JeffersGlass:
Implements passing slices to `str.__getitem__`. Adds tests. 

Also adds the `convertible()` utility to `typechecker.py`. This is a shallow wrapper around `CONVERT_maybe` which tests whether a given MetaArg is convertible to the given type. It's probably not the most efficient way to make that determination, but it seems nice to at least have a single function for it, so we can localise improvements there.


spy/analyze/scope.py
```diff
@@ -581,6 +581,11 @@ def flatten_Tuple(self, tup: ast.Tuple) -> None:
         for item in tup.items:
             self.flatten(item)
 
+    def flatten_Slice(self, slc: ast.Slice) -> None:
+        self.mod_scope.implicit_imports.add("_slice")
+        for item in (slc.start, slc.stop, slc.step):
+            self.flatten(item)
+
```
antocuni, line 588: I wonder why we didn't need until now? 🤔

spy/tests/compiler/test_str.py
```diff
@@ -65,6 +66,53 @@ def foo(a: str, i: i32) -> str:
         with SPyError.raises("W_IndexError", match="string index out of bound"):
             mod.foo("ABCDE", -6)
 
+    def test_getitem_slice(self):
+        # Because we can't wrap slices and pass them into the test, and
+        # because we want to only compiler the module once, we do this dance
+        # where we build a bunch of names functions inside module source
+        # Before it's compiled
```
antocuni: if it makes the test simpler, we can easily add support for passing slices into the test.

This seems to work:
```diff
diff --git a/spy/vm/vm.py b/spy/vm/vm.py
index bcde28bc..0bbfaddb 100644
--- a/spy/vm/vm.py
+++ b/spy/vm/vm.py
@@ -652,6 +652,8 @@ class SPyVM:
             return W_Bytes(self, value)
         elif T is Loc:
             return W_Loc(value)
+        elif T is slice:
+            return self.wrap_slice(value)
         elif T is UnwrappedStruct:
             return value.spy_wrap(self)
         elif isinstance(value, FunctionType):
@@ -683,6 +685,15 @@ class SPyVM:
             vm.call_w(w_push, [w_res, vm.wrap(item)], color="red")
         return w_res
 
+    def wrap_slice(self, s: slice) -> W_Object:
+        vm = self
+        self.import_("_slice")
+        w_T = self.lookup_global(FQN("_slice::Slice"))
+        w_start = vm.wrap(s.start)
+        w_stop = vm.wrap(s.stop)
+        w_step = vm.wrap(s.step)
+        return vm.call_w(w_T, [w_start, w_stop, w_step])
+
     def unwrap(self, w_value: W_Object) -> Any:
         """
         Useful for tests: magic funtion which wraps the given app-level w_
```

spy/tests/compiler/test_str.py
```diff
@@ -65,6 +66,53 @@ def foo(a: str, i: i32) -> str:
...
+        # Generate a function (who's name we know) that takes no arguments
+        # and returns the slice.indices() of the given slice.
+        # Append that function to the source code for the module
+        # we're about to compile.
+        for s_in, slc, _ in eq_list:
+            src += dedent(f"""
+            def {args_to_func_name(s_in, slc)}() -> str:
+                s: slice = slice({slc.start}, {slc.stop}, {slc.step})
+                return "{s_in}"[s]
+            """)
```
antocuni: es, I'm not a big fan to tests which build the source dynamically because they are hard to follow and to grep (although I understand why we need it here). If we can avoid this by implementing `vm.wrap(slice)`, I'm all for it!

spy/vm/str.py
```diff
@@ -45,6 +45,7 @@ class W_Str(W_Object):
     __spy_lazy_attributes__ = {
         "isascii": FQN("_str::methods::isascii"),
         "upper": FQN("_str::methods::upper"),
+        "__getitem_slice__": FQN("_str::methods::_getitem_slice_impl"),
```
antocuni: maybe call it `_getitem_slice`?
Else people might thing that it's a special dunder method called automatically by SPy on slices.

Also, I think it should have the same name on both sides: so:
```
"_getitem_slice": FQN("_str::methods::_getitem_slice")
```

spy/vm/typechecker.py
```diff
@@ -142,6 +142,28 @@ def typecheck_opspec(
...
+    try:
+        CONVERT_maybe(vm, wam_expT, wam)
+    except SPyError as exc:
+        if exc.match(W_TypeError):
+            return False
+        else:
+            raise exc
+
+    return True
+
```
antocuni: two notes:
1. I'd call it `is_convertible_to`
2. maybe we can put it as a method of `SPyVM`? Fun fact: we used to have `vm.is_compatible_type` and `vm.can_assign_from_to`, they were killed by commit 82c71b74 (PR #27 )

stdlib/_str.spy
```diff
@@ -88,3 +88,33 @@ class methods:
...
+                current += step
+
+        if not len(result_list):
+            return ""
+        
+        result_str = ""
+        for r in result_list:
+            result_str = result_str + r
+                   
+        return result_str
```
antocuni: note that this implementation is very slow, for two reasons:
1. `result_list` start small (capacity 4) and the grows while we .append, so we will allocate/copy multiple times
2. constructing `result_str` this way is O(n²): CPython has a special case for optimizing this particular construct, but we don't.

The "proper" way to do this is to do it at the `StrObject` level. The even "more proper" way is to wait to have `memcpy_slice` :)

stdlib/_str.spy
```diff
@@ -88,3 +88,33 @@ class methods:
             else:
                 llnew.utf8[i] = ch
         return StrObject.to_str(llnew)
+        
+    def _getitem_slice_impl(self: str, slc: slice) -> str:
+        # Fast path for trivial slices:
+        if (slc.start == 0 or slc.start_is_none) and (slc.stop == -1 or slc.stop_is_none) and (slc.step == 1 or slc.step_is_none): return self
```
dpdani: These forms of `my_str[:]` in Python mean make a copy of the string, not sure returning self is what we want here

spy/analyze/scope.py
JeffersGlass: I think all the tests I wrote for `list.__getitem__(slice)` used numeric literals. Since they don't need flattening, the just passed.

I'll add tests that require flattening to `set.__getitem__(slice)`, and I can go back and add them for list as well. Probably something for us to be aware of when adding new nodes to the Parser - we'll need to write tests that verify not just their functionality but that they shift/flatten their subnodes.

stdlib/_str.spy
antocuni: Is there any visible difference though, considering that strings are immutable?

dpdani: I stand corrected:

```
>>> a = "asdf"
>>> a[:]
'asdf'
>>> a[:] is a
True
```

dpdani: It's probably only relevant for `my_list[:]` or other mutable types

spy/tests/compiler/test_str.py
JeffersGlass: This does seem to mostly work, along with a small modification to `WasmFuncWrapper.py2wasm` so that objects passed it can be converted to unwrapped structs before being passed to the compiled module.

spy/vm/str.py
JeffersGlass: Agreed and done.

spy/tests/compiler/test_str.py
JeffersGlass: Agreed - the changes just pushed allow for passing slices in directly to the tests.

spy/vm/typechecker.py
JeffersGlass: Sounds good!

stdlib/_str.spy
JeffersGlass: Now that `ptr_copy_slice` is in, reimplemented in a way that only allocates the low level string once, since we can calculate its size from the slice.

JeffersGlass:
@antocuni I think this is ready for another look, at your leisure

spy/analyze/scope.py
antocuni: makes sense. No need to write list-specific tests, but we should probably add the relevant test to `test_scope.py`

spy/tests/compiler/test_str.py
```diff
@@ -65,6 +66,36 @@ def foo(a: str, i: i32) -> str:
         with SPyError.raises("W_IndexError", match="string index out of bound"):
             mod.foo("ABCDE", -6)
 
+    def test_getitem_slice(self):
+        # Because we can't wrap slices and pass them into the test, and
+        # because we want to only compiler the module once, we do this dance
+        # where we build a bunch of names functions inside module source
+        # Before it's compiled
```
antocuni: this comment is now outdated I think?

spy/tests/compiler/test_str.py
```diff
@@ -65,6 +66,36 @@ def foo(a: str, i: i32) -> str:
...
+        ]
+
+        mod = self.compile("""
+            def get_slice(s: str, slc: slice) -> str:
+                return s.__getitem__(slc)
+        """)
+
+        for in_, slc, out in eq_list:
+            assert mod.get_slice(in_, slc) == out
+
```
antocuni: *very* nitpick so feel free to ignore, but it's a good occasion to spread the word 😅.

I know that this style of testing is popular thanks to `pytest.parametrize` but I personally find it very hard to read, especially in case of failure because the traceback won't tell me which case is failing.

I find it much better this style:
```python
assert mod.get_slice("abc", slice(0, 3, None)) == "abc"
assert mod.get_slice("abs", slice(...)) == ...
...
```

there are a couple of examples in the spy codebase, see e.g. this:
https://github.com/spylang/spy/blob/5081e719e90f26ace47e3556bfbff7b28597849b/spy/tests/compiler/unsafe/test_div.py#L64-L71

One day I will write a blog post about it :).

In the meantime... feel free to ignore my rant 😂

spy/tests/wasm_wrapper.py
```diff
@@ -107,8 +107,17 @@ def py2wasm(self, pyval: Any, w_T: W_Type) -> Any:
...
             # a flat sequence of fields. This is the opposite of what we do in
             # to_py_result. It works only for flat structs with simple types.
+
+            # If we get a plain object, try to wrap convert it to a spy struct
+            # and then pull out the unwrapped version
+            if not isinstance(pyval, UnwrappedStruct):
+                try:
+                    pyval = self.vm.unwrap(self.vm.wrap(pyval))
+                except Exception as exc:
+                    raise TypeError(f"Unsupported type: {w_T}") from exc
```
antocuni: I think this is overly-generic. I think it's much easier/more readable/more debuggable to be explicit and do it only for slices. If and when we need a more general solution, we can think about that.
I modified this code in 38ceb7bb

stdlib/_str.spy
```diff
@@ -109,3 +110,55 @@ class methods:
...
+        # Fast path for simple slices
+        if slc.step == 1:
+            length = stop - start
+            if length <= 0:
+                return ""
+            llnew = StrObject.alloc(length)
+            ptr_copy_slice(llnew.utf8, 0, length, ll.utf8, start, stop)
+            return StrObject.to_str(llnew)
+
+        # Calculate number of elements in final list
```
antocuni:
```suggestion
        # Calculate length of the final string
```

antocuni: and probably the variable name should be something different than `num_members`?

antocuni (approved):
LGTM!
I left some comments/nitpicks but as usual they are not required to be fixed.

I also tweaked/simplified a bit the code in py2wasm, see the inline comment.

Feel free to merge it :).

spy/tests/compiler/test_str.py
JeffersGlass: No worries! That's a straightforward change to make in this case - I've adjusted all of those to the `assert mod.getslice(...) == ...` form.

spy/tests/compiler/test_str.py
JeffersGlass: Agreed and removed.

PR #555, opened 2026-06-09
fix playground: do not print traceback for SystemExit

### LINKED ISSUE #535: Playground Execute Error
The following error raises when any spy file is executed in the playground.

```python
Hello from SPy 🥸
30
  File "/lib/python3.13/site-packages/spy/cli/_runners.py", line 35, in _pyodide_main
    await _run_command(user_func, args)
  File "/lib/python3.13/site-packages/spy/cli/_runners.py", line 46, in _run_command
    return await user_func(args)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/spy/cli/commands/execute.py", line 27, in execute
    execute_spy_main(vm, w_mod, argv, redshift=False, _timeit=args.timeit)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/spy/cli/_runners.py", line 118, in execute_spy_main
    sys.exit(0)
    ~~~~~~~~^^^
SystemExit: 0
```

paugier:
Fixes #535.

~~It seems that we don't actually need to call `sys.exit(0)` at the end of `execute_spy_main`.~~

### DIFF
--- a/spy/cli/_runners.py
+++ b/spy/cli/_runners.py
@@ -29,13 +29,15 @@ GLOBAL_VM: Optional[SPyVM] = None
 
 async def _pyodide_main(user_func: Callable, args: "Base_Args") -> None:
     """
-    For some reasons, it seems that pyodide doesn't print exceptions
-    uncaught exceptions which escapes an asyncio task. This is a small wrapper
-    to ensure that we display a proper traceback in that case
+    For some reasons, it seems that pyodide doesn't print uncaught exceptions
+    which escapes an asyncio task. This is a small wrapper to ensure that we
+    display a proper traceback in that case.
     """
     try:
         await _run_command(user_func, args)
-    except BaseException:
+    except BaseException as exc:
+        if isinstance(exc, SystemExit):
+            return
         traceback.print_exc()
 
 
--- a/examples/1_high_level/exit_code.spy
+++ b/examples/1_high_level/exit_code.spy
@@ -1,3 +1,10 @@
+"""
+Things to notice:
+  - `main()` can take `argv` as a `list[str]` and can return an `i32` exit code
+  - the exit code is propagated to the OS
+  - `argv` contains the script/executable name and the script arguments
+"""
+
 def main(argv: list[str]) -> i32:
     print("argv:")
     for arg in argv:
--- a/playground/main.py
+++ b/playground/main.py
@@ -32,13 +32,6 @@ console.log("[Python] SPy imports complete")
 libspy.LIBSPY_WASM = str(URL.new("./libspy.mjs", document.baseURI))
 
 
-def spy_main(argv):
-    try:
-        spy.cli.app(argv)
-    except SystemExit:
-        pass
-
-
 # =========== GUI code ==========
 
 
@@ -152,7 +145,7 @@ def run_spy_file_with_args(argv: list[str]):
     with open(display_filename, "w") as f:
         f.write(text)
 
-    spy_main(argv)
+    spy.cli.app(argv)
 
 
 def run_click(event):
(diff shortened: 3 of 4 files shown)


spy/cli/_runners.py
```diff
@@ -115,7 +115,8 @@ def execute_spy_main(
         sys.exit(vm.unwrap_i32(w_res))
     else:
         assert w_res is B.w_None
-        sys.exit(0)
+        if sys.platform != "emscripten":
+            sys.exit(0)
```
antocuni: I vaguely understand why it might be needed but from a casual reader it's not particularly clear.

As I said on discord, it somehow feels wrong to invoke `sys.exit` from here. A better fix might be to return the exit code from here, and propagate up in the caller chain until we find the place which actually should call sys.exit.

If we decide to keep sys.exit here, I'd like a comment to explain why we don't want it on emscripten.

spy/cli/_runners.py
```diff
@@ -112,10 +112,25 @@ def execute_spy_main(
...
+
+
+def maybe_sys_exit(exit_code: int) -> None:
+    """
+    Call sys.exit(exit_code), unless we are running on emscripten/Pyodide
+    and the exit code is 0, in which case we return normally.
+
+    On emscripten, sys.exit() raises an unhandled exception instead of
+    terminating cleanly, which is fine for non-zero codes (it signals an
+    error) but misleading for a successful exit.
```
antocuni:
```suggestion
    Call sys.exit(exit_code), unless we are running on emscripten/Pyodide
    and the exit code is 0, in which case we return normally.

    THIS IS A HACK!
    In playground/main.py we have this code:
        def spy_main(argv):
            try:
                spy.cli.app(argv)
            except SystemExit:
                pass

    In theory, SystemExist should be caught by it but for some reason it's
    not and the traceback ends up in the playground terminal.
    
    We should investigate more properly, but for now it's good enough.
```

antocuni (approved):
LGTM.
I tried to improve the docstring of `maybe_sys_exit` so explain that we don't actually know what's happening :).

Feel free to accept or reject the comment, and feel free to merge.

spy/cli/_runners.py
paugier: I think I start to understand: the `try`/`except` does not work because of async. The async function `execute` is called after `spy.cli.app(argv)` returns!

spy/cli/_runners.py
paugier: I removed that, which was not actually necessary.

spy/cli/_runners.py
paugier: Actually the try/except in main.py is useless (and I think it cannot work because of async) since `SystemExit` is already caught in `_pyodide_main`!

The real issue was in `_pyodide_main`, which is now fixed.

spy/cli/_runners.py
```diff
@@ -29,13 +29,15 @@
...
+    For some reasons, it seems that pyodide doesn't print uncaught exceptions
+    which escapes an asyncio task. This is a small wrapper to ensure that we
+    display a proper traceback in that case.
     """
     try:
         await _run_command(user_func, args)
-    except BaseException:
+    except BaseException as exc:
+        if isinstance(exc, SystemExit):
+            return
```
antocuni, line 40: oh, that was it!
Thanks for finding it :)

antocuni (approved):
LGTM.

Sorry for having insisted/pushed to find the real solution but I think this is much better than the original hack 🎉 ❤️ 

Feel free to merge.

PR #557, opened 2026-06-10
unsafe.sizeof: support f32 and u32

seibert:
Also added unit test for sizeof of all primitive numeric types.

(Patch created with AI assistance.)

### DIFF
--- a/spy/tests/compiler/unsafe/test_sizeof.py
+++ b/spy/tests/compiler/unsafe/test_sizeof.py
@@ -1,19 +1,12 @@
-import pytest
-
 from spy.vm.b import B
 from spy.vm.modules.unsafe.misc import sizeof
 
 
-@pytest.mark.parametrize(
-    "w_T, expected",
-    [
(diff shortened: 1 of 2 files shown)


spy/tests/compiler/unsafe/test_sizeof.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.mark.parametrize(
+    "w_T, expected",
+    [
+        (B.w_i8, 1),
+        (B.w_u8, 1),
+        (B.w_i32, 4),
+        (B.w_u32, 4),
+        (B.w_f32, 4),
+        (B.w_f64, 8),
+    ],
```
antocuni, line 16: I know it's an unpopular opinion but I really dislike `parametrize` (I should write a blog post about it one day :)).
I rewrote it using explicit asserts.

antocuni (approved):
LGTM. I tweaked the test and added support for bool while I'm at it.

PR #558, opened 2026-06-10
add i64 / u64 types

seibert:
(Based on #557.  Merge that one first)

This adds the 64-bit integer types, which are frequently needed to interface with external C APIs.

The PR also includes a change to how integer literals are handled to ensure they are not cast down to int32 before being assigned to a 64-bit integer type.

(AI assistance was used to implement this PR.)

### DIFF
--- a/spy/libspy/include/spy/operator.h
+++ b/spy/libspy/include/spy/operator.h
@@ -16,6 +16,8 @@
 #define u8 uint8_t
 #define i32 int32_t
 #define u32 uint32_t
+#define i64 int64_t
+#define u64 uint64_t
 #define f32 float
 #define f64 double
 
@@ -35,6 +37,16 @@ DEFINE_CONV(u8, f64)
 DEFINE_CONV(u32, i32)
 DEFINE_CONV(u32, f64)
 
+// 64-bit int conversions (see convop.py registrations).
+DEFINE_CONV(i32, i64)
+DEFINE_CONV(i32, u64)
+DEFINE_CONV(i64, i32)
+DEFINE_CONV(u64, i32)
+DEFINE_CONV(i64, u64)
+DEFINE_CONV(u64, i64)
+DEFINE_CONV(i64, f64)
+DEFINE_CONV(u64, f64)
+
 DEFINE_CONV(f32, f64)
 
 DEFINE_CONV(f64, f32)
@@ -43,6 +55,8 @@ DEFINE_CONV(f64, f32)
 #undef u8
 #undef i32
 #undef u32
+#undef i64
+#undef u64
 #undef f32
 #undef f64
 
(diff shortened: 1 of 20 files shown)


spy/vm/astframe.py
```diff
@@ -891,6 +901,31 @@ def eval_expr_Const(self, const: ast.Const) -> W_MetaArg:
...
+        """
+        if varname is None or varname not in self.locals:
+            return None
+        w_T = self.locals[varname].w_T
+        pyclass = self._INT_LITERAL_WRAPPERS.get(w_T)
+        if pyclass is None:
+            return None
+        w_val = pyclass(const.value)
+        return W_MetaArg(self.vm, "blue", w_T, w_val, const.loc)
+
```
antocuni, line 928: this seems overly complicated. I think we should declare that int literals which don't fit into an i32 are automatically wrapped into an `W_I64`.

This is temporary anyway: the long term plan is to have a notion of "comptime int" which is "sizeless", and gets assigned a concrete type only when you need to create a variable out of it.
This solves ANOTHER problem which is that if you have this:
```python
x: i8 = 10
x + 1
```
currently `1` is typed as `i32`, and thus `x + 1` as well. With a "comptime int", we would be able to keep `x + 1` to be an i8.
That's what zig does.

antocuni (commented):
I didn't do a full review, but let me comment on the "int 64 literal" problem

spy/libspy/include/spy/operator.h
```diff
@@ -372,6 +386,52 @@ spy_unsafe$u32_unchecked_mod(uint32_t x, uint32_t y) {
     return x % y;
 }
 
+// ---- 64-bit int div / floordiv / mod ----
+// Generated for {i64, u64} via a macro to avoid ~12 near-identical functions
+// each. `div` returns double (Python `/` semantics, matching i32/u32);
+// floordiv/mod return the integer type. SIGNED guards the (x^y)<0 floor
+// correction, which is meaningless for unsigned.
+// NAME is the SPy type name used in the symbol (i64/u64); CT is the C type.
+#define SPY_DEF_INT64_DIVMOD(NAME, CT, SIGNED)                                         \
```
antocuni, line 395: I'm not necessarily opposed to this but:
1. if we do this, we should do for ALL div/mod/floordiv functions, not just the i64 ones
2. the full name of the function should be spelled explicitly in the macro: this makes it much more easy to grep. Something like:

```c
DEFINE_DIV(spy_operator$i64_div, i64, int64_t)
DEFINE_DIV(spy_operator$i32_div, i32, int32_t)
...
DEFINE_UNCHECKED_DIV(spy_operator$i64_unchecked_div, i64, int64_t)
...
```

there is more code duplication than with this approach, but also much less magic IMHO.

seibert:
closing this PR in favor of what Antonio is working on

PR #559, opened 2026-06-10
Add support for out-of-tree builtin modules

antocuni:
Out-of-tree builtin modules use the same mechanism as the modules in `spy/vm/modules/`, but as the name suggests they live out of the main spy repo, and can be compiled separately.

The main use use at the moment is to write bindings for external C libraries. The naming convention is to call the `spyvm_MODNAME`.
One problem is that the interpreter needs a .wasm version of the library: we don't have a package manager for wasm yet, so we need ad-hoc solutions for now.

We provide two examples:
- `examples/out-of-tree/spyvm_qrcodegen`. This *vendors* the library `qrcodegen`. The provided makefile builds the lib for the whole matrix of SPy target/build_type (`wasi`, `native`, `native-static`, `emscripten` * `release`, `debug`). Since we build the wasm version, it works also in the interpreter.
- `examples/out-of-tree/spyvm_libmagic`: this wraps the *system* `libmagic.so`: as such, we don't have a wasm version and it just works after compilation.

See the related README for a more detailed explanation.

---

To achieve this goal, we had to implement a number of features:

- WASM bundles: SPyVM needs to be able to call (via `vm.ll.call("funcname", ...)`) arbitrary WASM functions. Until now, we had only `libspy.wasm` with all the needed functions. Now, we also have things like `build/wasi/debug/libspyvm_qrcodegen.a`, which in turn can call functions from libspy (e.g. `spy_str_alloc`). This is a job for a linker, and we just reuse clang: at interpreter startup, we determine which combination of spyvm modules we need and we link all their `.a` WASM archives into a single wasm bundle, which is loaded by wasmtime. WASM bundles are cached.

- `build_info`: a spyvm module can declare extra cflags, ldflags, includes, etc. to be forwared to the C compiler. We added plumbing across all the toolchain to write them to `build.ninja`.

- add `python -m spy.build.flags [--cc|--cflags|--ldflags|...]`: this is intended to be called by `Makefiles` to get the correct flags for a given target/build_type pair.  This helps to simplify libspy's `Makefile`.

- add the option `spy --extra-vm-module /path/to/mod`

- add `spy.toml`, which can be used to automatically pass multiple `--extra-vm-module` for a given spy script


seibert:
I think this looks reasonable.  I'm trying to decide if there are any drawbacks to putting the libraries and flags inside the Python registry for the module, rather than setting them externally (like on the spy command line).  I'm wondering if setting on the spy command line might be needed if the C library is distributed separate from the SPy module, and its location varies depending on system setup.  (Ex: Maybe I'm using the system installation of libxml, maybe I'm using the one in my conda environment?)

antocuni:
> I'm wondering if setting on the spy command line might be needed if the C library is distributed separate from the SPy module, and its location varies depending on system setup. (Ex: Maybe I'm using the system installation of libxml, maybe I'm using the one in my conda environment?)

that's a good point and I don't know what is the best tradeoff.
The following is a kind of "thinking out loud" section:

My reasoning is that "which concrete C lib to link" is a property of the extension module, not of the final consumer.

The CPython parallel is this:
1. I want to build/install numpy from source
2. the numpy configure step (or equivalent) decides where to find blas&co.
3. I build a wheel
4. from now, the C lib choice is fixed and every python interp can import numpy without thinking about it

The parallel doesn't fully apply to spyvm_modules because it's not clear whether we should treat them as source distributions or "compiled" artifacts.
Also, remember that the interpreter needs a .wasm, which is unlikely to be distributed with your system package manager.

In the qrcodegen example, you must go there and type `make` in order to build the actual lib (which builds the .wasm for wasi and emscripten, the .so, the .a, etc) this is the equivalent of point (3) above, and point (2) is hardcoded in the `__init__`.

So _probably_, a fully (over?) engineered system would be:
1. the `spyvm_module` exist in source form in a github repo or on disk
2. you have an install/build step which creates/links binary artifacts (if needed)
3. the result of (2) is a manifest which contains the path to (1) plus all the cflags, ldflags, .wasm locations, etc. (or if we want the equivalent of a wheel, a zip file containing all of this)
4. the end user refers to (3) when using `spy --extra-vm-module`

However, I'm not sure that at this stage we need this level of complexity, especially keeping in mind that out-of-tree vm modules might not be the proper long term solution).

I think we can get 90% of the benefits and 1% of complexity by keeping the build info in the `__init__`, and if you really need to configure them externally you use env variables.

Happy to be convinced otherwise :)

antocuni:
@seibert it took much longer than expected but now I have something which seem to work.
It it still marked as WIP as I haven't fully decided all the small design issues, but the basic concept seems to work.

To run `examples/out-of-tree`:
1. build the module first:
```
❯ make -C examples/out-of-tree/spyvm_qrcodegen/`
```

2. run the demo in the interpreter. This is VERY slow but the slowness comes from the printing loop (it's nested loop which needs to print things pixel-by-pixel, and in full interpreted mode is awfully slow):
```
❯ spy examples/out-of-tree/demo/main.spy hello
```

3. compile the demo:
```
❯ spy build --release examples/out-of-tree/demo/main.spy 
[release] examples/out-of-tree/demo/build/main 

❯ ./examples/out-of-tree/demo/build/main hello
[...]
```

PR #560, opened 2026-06-11
Nicer playground

paugier:
This PR contains several relatively small improvements of the playground.

The most important change is the fact that the playground now includes all examples in the four numbered categories (1_high_level, 2_metaprogramming, 3_low_level and 4_advanced).

Note: I used Claude for this!


playground/tests/test_spyast_js.py
```diff
@@ -69,15 +69,17 @@ def __call__(self, *args: str) -> str:
...
+    ) -> Generator[Rodney, None, None]:
         rodney_home = str(tmp_path_factory.mktemp("rodney"))
         rodney = Rodney(rodney_home=rodney_home)
         rodney("start")
         yield rodney
         rodney("stop")
 
     @pytest.fixture
-    def init(self, tmpdir, _rodney):
+    def init(self, tmpdir: pytest.TempdirFactory, _rodney: Rodney) -> None:
```
antocuni: I think this annotation is wrong: `tmpdir` should be a `py.path.local`  I think

antocuni: but also, note that for files in `spy/tests/` we have a special mypy.ini config specifically made to be able to use pytest fixtures without going crazy with annotations:
https://github.com/spylang/spy/blob/515c013fb73db034dae1ee665fcd4d2830d3f504/mypy.ini#L7-L9

so maybe we should do the same for this test file?

antocuni (approved):
LGTM, apart the annotations comments.
If possible, I think it would be better to keep its mypy config consistent with all the other test files, and then we don't need the annotations.

Feel free to merge after doing it!

PR #562, opened 2026-06-12
Implement `str.split()`

JeffersGlass:
Adds `str.split(sep)` to the stdlib. ~Note that the separator has to be explicitly specified; this does _not_ implement `split()` which splits on whitespace, which should probably wait until we have default arguments.~

This also implements the behavior of Python's `str.split()` (splitting on whitespace) by using `str.split(None)` via metafunc. I'm struggling with how to make the behavior I want work, where `str.split()` just works, but see below.

Also implements `WasmFuncWrapper._to_pylist_str()` to get lists back out of wasm tests.


antocuni:
I haven't reviewed it yet but please note that we DO have support for default args.

However, we don't have str|None yet, so str.split is still not easily typeable (but we could do it with a metafunc)

JeffersGlass:
> we DO have support for default args

Oh! I honestly don't think I knew that... exciting!

Don't sweat renewing this - I'm not sure if I've broken something by changing things in `WasmFuncWrapper` or if this is exposing a bug in `str.__getitem__(slice)` that I didn't know about... will do some legwork there

JeffersGlass:
Here's the strucutre I'd need to make `str.split()` (split on whitespace): 

```py
from operator import OpSpec
from types import NoneType

@blue.metafunc
def split(wam=None):
    if wam.static_type == str:
        def inner_sep(x: str) -> list[str]:
            return ['Split', 'on', 'sep', 'impl']
        return OpSpec(inner_sep, [wam])

    if wam.static_type == NoneType:
        def inner_none() -> list[str]:
            return ['Split', 'on', 'whitespace', 'impl']
        return OpSpec(inner_none, [])
    return OpSpec.NULL

def main() -> None:
    print(split('a'))  # works
    print(split(None)) # works
    print(split())     # weird error
```

Unfortunatly, this comes back with a baffling error:
```
TypeError: this function takes from 0 to 1 arguments but 0 arguments were supplied
  | /workspaces/spy/a.spy:5
  | def split(wam=None):
  | |__________________| function defined here

```

I tried making the default argument a MetaArg:
```
@blue.metafunc
def double(wam=MetaArg("blue", NoneType, None)
```
But the error is the same

spy/tests/wasm_wrapper.py
```diff
@@ -215,6 +220,50 @@ def _to_pylist_i32(self, pyres: UnwrappedStruct) -> list[Any]:
...
+        #     i32 length;
+        #     i32 capacity;
+        #     ptr[StrObject] items;  // represented as {addr, length} in debug mode
+        # }
+
+        addr = ll_ptr.addr
+        list_length = self.ll.mem.read_i32(addr)
+        items_addr = self.ll.mem.read_i32(addr + 8)
+
+        def _read_str_from_addr(s_addr: int) -> str:
```
antocuni: any particular reason why you didn't use `ll.read_str`?

stdlib/_str.spy
```diff
@@ -138,7 +139,6 @@ class methods:
                 i = 0
                 if step > 0:
                     while current < stop:
-                        # ptr_copy_slice(dst, dstart, dend, src, sstart, send)
```
antocuni, line 141: this comment was there to document the signature of the function called the line below :)

stdlib/_str.spy
```diff
@@ -367,4 +367,134 @@ class methods:
             else:
                 llnew.utf8[i] = ch
         return StrObject.to_str(llnew)
+    
+    
+    @blue.metafunc
+    def split(self, *args_m):
+        if len(args_m) == 1 and args_m[0].static_type == str:
+            def split_sep(_self: str, sep: str) -> list[str]:
```
antocuni: nitpick, but I think for readability it's better to keep individual implementations out of the metafunc: we avoid a nesting level, and we probably also get a better FQN

stdlib/_str.spy
```diff
@@ -367,4 +367,134 @@ class methods:
             else:
                 llnew.utf8[i] = ch
         return StrObject.to_str(llnew)
+    
+    
+    @blue.metafunc
+    def split(self, *args_m):
```
antocuni: the convention is to use `m_self` here (since it's a metaarg)

stdlib/_str.spy
```diff
@@ -367,4 +367,134 @@ class methods:
...
+                    if end_of_string:
+                        break
+                    elif matched:
+                        # If we matched a separator, copy the items from start_of_chunk up to the 
+                        # separator as the latest list item
+                        new_len = search_position-start_of_chunk
+
+                        llnew = StrObject.alloc(new_len)
+                        for j in range(new_len):
+                            llnew.utf8[j] = llself.utf8[start_of_chunk+j]
```
antocuni: can't we use ptr_copy here?

stdlib/_str.spy
```diff
@@ -367,4 +367,134 @@ class methods:
...
+        if len(args_m) == 1 and args_m[0].static_type == str:
+            def split_sep(_self: str, sep: str) -> list[str]:
+                # Based on CPython stringlib/split.h
+                len_self = len(_self)
+                if len_self == 0:
+                    return [""]
+
+                sep_len_check = len(sep)
+                if sep_len_check == 0:
+                    raise ValueError("empty separator")
```
antocuni: I think that CPython has a special case for when `sep` is a single char.
It might be worth adding the same for SPy (but doesn't need to be this PR, can be a follow up)

stdlib/_str.spy
```diff
@@ -367,4 +367,134 @@ class methods:
...
+
+                llself = StrObject.from_str(_self)
+                llsep = StrObject.from_str(sep)
+                sep_len = llsep.length
+                
+                # For each starting position in the target string, look through len(sep) characters
+                # If all of them match the characters in sep, we have a match and should split here
+                # If not, increment starting position and try again
+                result = list[str]()
```
antocuni: ideally, we should pre-allocate a list of a reasonable/exact capacity, to avoid unnecessary copies. CPython does that.
However, I don't think that your list type supports it.

It's worth putting a comment to flag it though.

stdlib/_str.spy
```diff
@@ -367,4 +367,134 @@ class methods:
...
+                    ptr_copy_slice(llnew.utf8, 0, llnew_len, llself.utf8, rear_pointer, lead_pointer)
+                    result.append(StrObject.to_str(llnew))
+
+                return result
+
+            return OpSpec(split_whitespace)
+
+        return OpSpec.NULL
+
+    def isspace(self: str) -> bool:
```
antocuni, line 383: worth adding a comment to specify that this is broken on "real" unicode strings :)_

antocuni (approved):
LGTM, but see my comments, in particular the one about `ll.read_str` and the one about ptr_copy.
The others are mostly style/adding comments.

The PR description still says that you are having troubles with `.split()`, but I think you solved the problem, didn't you?

antocuni:
@JeffersGlass also, another idea which I got 10 seconds after pressing "submit review" :).

We will probably want to have also `bytes.split()`, and we don't want to duplicate code.
When it's time for it, we can probably use SPy's superpower to have a nice implementation for both. Something like this:

```python
# stdlib/_split.spy

@blue.generic
def split_whitespace[T](s: T) -> list[T]:
    ...

@blue.generic
def split_sep[T](...):
    ...

# stdlib/_str.spy

class methods:

    @blue.metafunc
    def split(...):
        return OpSpec(split_whitespace[str], ...)
```

CPython does something similar but needs to rely on hacky C macros, SPy will be able to do it the proper way :)

spy/tests/wasm_wrapper.py
JeffersGlass: ... I didn't know it existed, so i ended up reimplementing it 🤣 I'll switch to that!

stdlib/_str.spy
JeffersGlass: Makes sense! I sprinkled that same line around while I was writing the split code - this one must have gotten caught in the crossfire while I was cleaning up.

stdlib/_str.spy
JeffersGlass: For sure - you mean move `def split_sep` outside of the metafunc yeah? Should it also move outside of the `methods` class?

stdlib/_str.spy
JeffersGlass: Even better, it can be a single `ptr_copy_slice`!

stdlib/_str.spy
JeffersGlass: It does indeed! Is this the sort of thing that's worth memorializing in the code somehow? Like a WIP comment or similar?

stdlib/_str.spy
antocuni: > For sure - you mean move def split_sep outside of the metafunc yeah? 

yes

> Should it also move outside of the methods class?

I don't know what "looks" better. Both choices are good, your call :)

stdlib/_str.spy
antocuni: yes, a WIP or TODO comment look good

JeffersGlass:
Thank you for the review @antocuni ! I think i've addressed everything, except I have a question about where the implementation functions should go once they're not in the metafunc (see thread above).

Very neat idea about reusing the metafunc for bytes and strings 🤯

antocuni (approved):
🔝👍

stdlib/_str.spy
JeffersGlass: Sounds good! I think outside of `methods` is cleaner, just pushed that

stdlib/_str.spy
JeffersGlass: Done!

PR #563, opened 2026-06-14
fix playground build

paugier:
I broke the playground with my PR #560. This should fix it.


paugier:
I take the responsibility to merge this fix in main. It's simple and I just introduced the bug (sorry for that)!

The failing  pyodide-test-ubuntu-latest-3.12 has nothing to do with these changes.

PR #565, opened 2026-06-15
Create a benchmarks directory and add two simple benchmarks

paugier:
This PR creates the directory benchmarks that can be used to gather few benchmarks while SPy is still very unstable.

A very simple benchmark for a simple array library is added. Python-Numpy is compared to SPy.

The benchmark showed that the creation of arrays (gc_alloc of a relatively large f64 buffer) is very slow.

~~I propose an improvement: by using `GC_MALLOC_ATOMIC` for the allocation of types that do not contain pointer, the slowness compared to Numpy disappears.~~

~~Claude helps me for the  `GC_MALLOC_ATOMIC` part.~~

Edit: I extended this PR to test the benchmark code. For SPy benchmarks, the expected output are saved and compared as for examples. For code in other languages, they are just run in the CI so that we get an error if they fail.

Edit2: I separated the`GC_MALLOC_ATOMIC` part in #594 to ease review.


paugier:
Note that this PR has a strong impact on the benchmarks for [spy-demos/loop-fusion](https://github.com/paugier/spy-demos/tree/loop-fusion/loop-fusion):

**Before:**

```
(.venv) me@meige7ltpa212:~/dev/spy-demos/loop-fusion$ ./build/bench_spy 
add1:
<spy type 'lib_array1d_simple::ndarray[f64]::Self' (struct)> 2 0.235501
<spy type 'lib_array1d::ndarray[f64]::Self' (struct)> 2 0.197771
<spy type 'lib_ndarray::ndarray[f64]::Self' (struct)> 2 0.205385
add4:
<spy type 'lib_array1d_simple::ndarray[f64]::Self' (struct)> 5 1.0396
<spy type 'lib_array1d::ndarray[f64]::Self' (struct)> 5 0.207149
<spy type 'lib_ndarray::ndarray[f64]::Self' (struct)> 5 0.20734
```

**After:**

```
(.venv) augier3pi@meige7ltpa212:~/dev/spy-demos/loop-fusion$ ./build/bench_spy 
add1:
<spy type 'lib_array1d_simple::ndarray[f64]::Self' (struct)> 2 0.0819983
<spy type 'lib_array1d::ndarray[f64]::Self' (struct)> 2 0.077879
<spy type 'lib_ndarray::ndarray[f64]::Self' (struct)> 2 0.0804534
add4:
<spy type 'lib_array1d_simple::ndarray[f64]::Self' (struct)> 5 0.36166
<spy type 'lib_array1d::ndarray[f64]::Self' (struct)> 5 0.0797122
<spy type 'lib_ndarray::ndarray[f64]::Self' (struct)> 5 0.0774825
```

To be compared with the Numpy times:

```
$ python bench_numpy.py 
add1:
0.075 s
add4:
0.214 s
```

benchmarks/array_simple/lib_array1d.spy
```diff
@@ -0,0 +1,67 @@
+from unsafe import gc_alloc, gc_ptr, ptr_setbytes
+from operator import OpSpec, MetaArg
```
antocuni, line 2: any reason why we cannot use `stdlib/array.spy`?

antocuni: if array.spy is missing features needed for the benchmark, maybe we could just add them?

benchmarks/array_simple/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Before usage of `GC_MALLOC_ATOMIC`, the runtime was highly dominated by allocations.
+
+```
+./build/bench
+spy add: 0.166543 s
+spy add inplace: 0.0124178 s
+spy empty: 0.105744 s
+spy zeros: 0.131674 s
+```
```
antocuni, line 28: this is an interesting fact but probably doesn't belong to this readme?
If we merge this PR before the GC_MALLOC_ATOMIC one, it probbly make sense to put these numbers in the malloc_atomic PR description

benchmarks/fibo/README.md
```diff
@@ -0,0 +1,33 @@
...
+## The surprising result: why is Julia 2x slower than Codon/SPy?
+
+Julia is a compiled language with a mature LLVM backend and `-O3` explicitly enabled, so
+one might expect it to match Codon and SPy. It does not, and the reason is instructive.
+
+Julia compiles functions on first call via a JIT (Just-In-Time) compiler. Even though
+Julia's LLVM backend is the same technology as clang's, a JIT operates under a hard
+constraint: **compilation time is paid at runtime**. Julia's LLVM pipeline is therefore
+deliberately tuned to be conservative — it limits inlining budgets and the depth of
+interprocedural optimisation passes in order to keep latency acceptable.
```
antocuni, line 29: I think it would be more fair to report cold and warm performance for all the JITted languages

antocuni (commented):
+1 for starting to collect benchmarks.
I'm -0.5 to put them in this repo though. Maybe it's better to start now from `spylang/benchmarks`? Happy to create the repo is you don't have enough permissions.

The biggest concern is that benchmarks might break while we change spy, but this will probably be true until we have spy 1.0, which is likely many years in the future.
A good middle ground could be to have a spy CI job which clones the benchmarks repo and runs the tests, so we will have visibility when a benchmark breaks.

My other remark is that if we want the numbers to be useful, we should probably have a better way to compute them.
For compiled languages maybe we should use `hyperfine` (or similar) which automatically runs benchmark multiple times and accout for noise.

For interpreted/JITted ones, we cannot use hyperfine but there might be something similar. It's a tedious task but it looks like claude could help here.  This doesn't need to be done now but it's good to keep in mind.

JeffersGlass:
@paugier now that Spylang/benchmarks exists, is it ok to close this PR?

paugier:
Yes, I close it!

PR #567, opened 2026-06-15
playground: avoid SystemExit being turned into an error

### LINKED ISSUE #566: Unclosable Popup in New Playground
@paugier I went to have a poke at #564, but now when I run _any_ code in the new playground, I get this unclosable-popup:

Image

paugier:
Fixes #566 

Actually the try/except removed in https://github.com/spylang/spy/pull/555 was useful.


paugier:
I merge this since it's very simple and it should the playground for Chrome users!

PR #569, opened 2026-06-17
546 argv passthrough

### LINKED ISSUE #546: argv passthrough when running `spy build -x`
`spy build -x program.spy` is convenient to build and run a spy program, but now that SPy has argv support, we need a way to pass arguments to the spy program in this command line.  Right now, there is no way to do that:

https://github.com/spylang/spy/blob/1292f43e6386df0169ea8acc0c06b33b5f945221/spy/cli/commands/build.py#L185

I'm not sure which convention to draw from here.  One option is to require the `-x program.spy` option to be last, and then everything else must be an argument to `program.spy`.  It seems like that would be easy to mess up though, so maybe we need  a `--` separator argument to distinguish the compiler arguments from the application arguments.

JeffersGlass:
This PR makes 3 changes to the CLI. Let me illustrate them with some sample code:

1. `spy build -x` now passes any values after the filename to the newly-built executable as agrv when it is run
1. The `execute`, `redshift`, and `build` commands treat all values after the filename as values to be passed to`argv`. (Redshift is included since it also has execution mode, but it's an easy change if we want to not include it.)
1.  For those 3 commands, if any of the values after the filename exactly match an option flag for the command, a warning is printed.

Here's how that last options appears in practice:

```py
#foo.spy
def main(argv: list[str]) -> None:
    for s in argv:
        print(s)


$ spy build -x foo.spy a b --timeit
'--timeit' passed to 'spy build' as part of argv; to pass it as a flag, it must proceeed the .spy file name
[debug] build/foo 
build/foo
a
b
--timeit
```

@seibert Does this suit your needs?

Closes #546

### DIFF
--- a/spy/cli/spy_typer.py
+++ b/spy/cli/spy_typer.py
@@ -19,8 +19,18 @@ from spy.cli.commands.shared_args import Base_Args
 from spy.vendored.dataclass_typer import dataclass_typer
 
 
-class TyperDefaultCommand(typer.core.TyperCommand):
+class TyperStrictParseCommand(typer.core.TyperCommand):
+    def make_parser(self, ctx: click.Context) -> click.OptionParser:
+        """Creates the underlying option parser for this command."""
+        ctx.allow_interspersed_args = False
+        return super().make_parser(ctx)
+
+
+class TyperDefaultCommand(TyperStrictParseCommand):
     """Type that indicates if a command is the default command."""
+
+
+class TyperStrictDefaultCommand(TyperDefaultCommand, TyperStrictParseCommand): ...
 
 
 class SpyGroupConfig(TyperGroup):
@@ -110,6 +120,10 @@ class SpyGroupConfig(TyperGroup):
                 return cmd.name
         return lookup_name
 
+    def parse_args(self, ctx, args):
+        ctx.obj = {"args": args}
+        return super().parse_args(ctx, args)
+
 
 class SpyTyper(typer.Typer):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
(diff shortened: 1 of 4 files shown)


spy/cli/spy_typer.py
```diff
@@ -19,10 +19,46 @@
...
+        """
+        ctx.allow_interspersed_args = False
+        return super().make_parser(ctx)
+
+    def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
+        """
+        Show a hint if any of the values in argv match any of the possible option flags
+
+        >>> spy execute a.spy --timeit
+        '--timeit' passed to 'spy execute' as part of argv; to pass it as a flag, it must proceeed the .spy file name
```
antocuni: I'm not sure it's a good idea.
Some of the options that we have have pretty generic names like `--format` and `--dump`: this means that if I write a SPy program which accepts its own `--dump`, I will always see this warning?

I'm +0.5 to just kill the warning; that's the behavior of `python`, after all.

spy/tests/test_cli.py
```diff
@@ -127,6 +127,21 @@ def test_execute(self):
...
+                print(s, "+") # Separate argv so we're sure we're not just seeing the input on the command line
+        """
+        test = self.write("test.spy", src)
+        argsets = [["execute"], []]  # No subcommand is equivalent to execute command
+        for argset in argsets:
+            _, stdout = self.run(*argset, test, "1", "--timeit", "2")
+
+            # --timeit appears in argc and not as flag
+            assert "1 +\n--timeit +\n2" in stdout
+            assert "seconds" not in stdout
```
antocuni: I'm not a fan of `for` loops in tests. Better to unroll the loop, even if it causes a small duplication of `asserts`.

antocuni (approved):
LGTM. See my comments about the warning

spy/cli/spy_typer.py
paugier: Shouldn't it be "precede"?

paugier (approved):
The warning is nice but at the same time, it could be annoying if the script and the spy command really share an option name.

Maybe the explicit `--` separator should also be supported (without checks and warning)?

spy/cli/spy_typer.py
JeffersGlass: It sure should 😅 . But per Antonio, I"m happy to just kill this warning. I learned some new things implementing it though, which was nice.

JeffersGlass: Makes sense to me - I'll kill it.

spy/tests/test_cli.py
JeffersGlass: Sounds good! I'll unroll.

PR #570, opened 2026-06-18
Implement function keyword args

shalabhc:
Adds support for keyword arguments in SPy function calls.

* only spy funtions supported (no builtin functions)
* `**kwargs` not supported
* methods WIP

Demo:

```python
> cat examples/1_high_level/kwargs.spy   
"""
Things to notice:
  - keyword args can be passed in any order
  - after redshift, keyword args are reordered
"""

def compute(x: i32, y:i32, z: i32) -> i32:
    return 100 * x + 10 * y + z

def main() -> None:
    print("Hello from SPy 🥸")
    print(compute(1, z=5, y=3))

> spy examples/1_high_level/kwargs.spy   
Hello from SPy 🥸
135

> spy % spy rs examples/1_high_level/kwargs.spy
def compute(x: i32, y: i32, z: i32) -> i32:
    return 100 * x + 10 * y + z

def main() -> None:
    `_print::println[str]`('Hello from SPy 🥸')
    `_print::println[i32]`(`kwargs::compute`(1, 3, 5))
```

High level changes, described bottom up:

1. The function for default value resolution replaced with a new `W_ASTFunc._bind_args` that handles both default args and keyword args reordering. This needs to identify keyword args, so the metaarg now carries an optional `kwarg_name` field.
2. `ASTFrame.eval_expr_Call` now evaluates kwarg expressions on the `Call` node and tags their metaarg the keyword name.
3.  Parser now accepts keyword args and records them as a new `kwargs` field on the `Call` node Expr]]`. It still rejects `**kwargs` with a clear error. 

A side effect of the new binding method `_bind_args` is that missing args even when positional have a clear error:
```
  * x::main at /Users/shalabh/src-local/spy/examples/1_high_level/x.spy:6
  |     f()
  |     |_|

TypeError: f() missing required argument `a`
  | /Users/shalabh/src-local/spy/examples/1_high_level/x.spy:2
  | def f(a: i32) -> i32:
  |       |____| missing required argument

  | /Users/shalabh/src-local/spy/examples/1_high_level/x.spy:6
  |     f()
  |     |_| operator::CALL called here
```

This change in error required patching a whole bunch of tests.

### DIFF
--- a/examples/1_high_level/hello_sub.spy
+++ b/examples/1_high_level/hello_sub.spy
@@ -1,0 +1,13 @@
+"""
+Things to notice:
+  - keyword args can be passed in any order
+  - after redshift, keyword args are reordered
+"""
+
+
+def sub(x: i32, y:i32, z: i32) -> i32:
+    return x - (y - z)
+
+def main() -> None:
+    print("Hello from SPy 🥸")
+    print(sub(100, z=10, y=20))
--- a/examples/1_high_level/kwargs.spy
+++ b/examples/1_high_level/kwargs.spy
@@ -4,10 +4,9 @@ Things to notice:
   - after redshift, keyword args are reordered
 """
 
-
-def sub(x: i32, y:i32, z: i32) -> i32:
-    return x - (y - z)
+def compute(x: i32, y:i32, z: i32) -> i32:
+    return 100 * x + 10 * y + z
 
 def main() -> None:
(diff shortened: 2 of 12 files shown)


shalabhc:
closed because tests need fixing, but will do it in draft

shalabhc:
Superseded by #575

PR #572, opened 2026-06-18
`type` object docs

JeffersGlass:
Adds docs for the builtin type `type`, including notes on the special name attrbibutes from #547


antocuni:
LGTM

PR #575, opened 2026-06-19
Implement keyword arguments for spy functions

shalabhc:
Adds support for keyword arguments in SPy function calls.

* only spy functions and methods support kwargs
* metafuncs, builtin funcs, dynamic calls and `__call__` dont support keyword args and compilation fails
* `**kwargs` not supported

## Demo

```python
> cat examples/1_high_level/kwargs.spy   
"""
Things to notice:
  - keyword args can be passed in any order
  - after redshift, keyword args are reordered
"""

def compute(x: i32, y:i32, z: i32) -> i32:
    return 100 * x + 10 * y + z

def main() -> None:
    print("Hello from SPy 🥸")
    print(compute(1, z=5, y=3))

> spy examples/1_high_level/kwargs.spy   
Hello from SPy 🥸
135

> spy % spy rs examples/1_high_level/kwargs.spy
def compute(x: i32, y: i32, z: i32) -> i32:
    return 100 * x + 10 * y + z

def main() -> None:
    `_print::println[str]`('Hello from SPy 🥸')
    `_print::println[i32]`(`kwargs::compute`(1, 3, 5))
```

## High level changes for `op_CALL`

1. The parser now accepts keyword arguments, and the `Call` nodes now hold a new `kwargs` field. `**kwargs` is rejected by the parser.
2. The `op_CALL` op now accepts exactly two parameters `[wam_func, wam_funcargs]` where funcargs is a new `W_FuncArgs` object that contains all positional and keyword arguments. Source order for keyword arguments is preserved. Previously it accepted variadic args `[wam_func, wam_arg1, wam_arg2, ...]`. This required many call sites to change even when they only use positional args (eg when `op_GETITEM` calls into `op_CALL`). 
3. `ASTFrame.eval_expr_Call` now evaluates the values in the `Call.kwargs` field and builds a new `FuncArgs` to pass into `vm.call_op(op_CALL, ...)`.
4. A new `W_ASTFunc._bind_args(funcargs)` method replaces `_fill_defaults_maybe` since default handling and keyword mapping needs to happen at the same time. It returns an ordered list of inner args, reordering keyword args and injecting defaults. 
5. The redshift pass flattens kwargs to positional in parameter order, so the C backend needs no changes.

A side effect of the new binding method `_bind_args` is that missing args even when positional have a different error:
```
TypeError: f() missing required argument `a`
  | /Users/shalabh/src-local/spy/examples/1_high_level/x.spy:2
  | def f(a: i32) -> i32:
  |       |____| missing required argument
```

This change in error required patching some more tests.

## `op_CALLMETHOD` and `op_METACALL`

`op_CALLMETHOD` also follows the same design and implements keyword args for spy methods.
`op_METACALL` signature now changed to accept funcargs, but it rejects compilation if any kwargs are present.

## Future

* keyword args for metacalls and builtin functions - seem possible, needs extra work
* `__call_method__` and `__call__` - maybe we can change this to accept funcargs directly, allowing richer custom dispatch.

### DIFF
--- a/examples/1_high_level/kwargs.spy
+++ b/examples/1_high_level/kwargs.spy
@@ -1,0 +1,12 @@
+"""
+Things to notice:
+  - keyword args can be passed in any order
+  - after redshift, keyword args are reordered
+"""
+
+def compute(x: i32, y:i32, z: i32) -> i32:
+    return 100 * x + 10 * y + z
+
+def main() -> None:
+    print("Hello from SPy 🥸")
+    print(compute(1, z=5, y=3))
--- a/examples/expected_output/kwargs.txt
+++ b/examples/expected_output/kwargs.txt
@@ -1,0 +1,2 @@
+Hello from SPy 🥸
+135
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -411,6 +411,7 @@ class Call(Expr):
     precedence = 16
     func: Expr
     args: list[Expr]
+    kwargs: list[tuple["StrLiteral", Expr]] = field(default_factory=list)
 
 
 @astnode
@@ -427,6 +428,7 @@ class CallMethod(Expr):
     target: Expr
     method: StrLiteral
     args: list[Expr]
+    kwargs: list[tuple["StrLiteral", Expr]] = field(default_factory=list)
 
 
 @astnode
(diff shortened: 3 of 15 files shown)


paugier:
You should be able to fix the error in the CI related to examples test by running `pytest --update-examples -k kwargs` in the examples directory. Check that the new file in expected_output in correct and commit it.

shalabhc:
> You should be able to fix the error in the CI related to examples test by running `pytest --update-examples -k kwargs` in the examples directory. Check that the new file in expected_output in correct and commit it.

done and it passes locally. also rebased on recent changes.

but i'm not sure how to re-trigger the test suite.

PR #580, opened 2026-06-19
fix playground: export individual uint32_t fields for str/bytes layout

### LINKED ISSUE #578: Playground: TypeError because layout is None
We now get in the playground:

```
  File "/lib/python3.13/site-packages/spy/cli/_runners.py", line 38, in _pyodide_main
    await _run_command(user_func, args)
  File "/lib/python3.13/site-packages/spy/cli/_runners.py", line 51, in _run_command
    return await user_func(args)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/spy/cli/commands/redshift.py", line 68, in redshift
    vm = await init_vm(args)
         ^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/spy/cli/_runners.py", line 155, in init_vm
    vm = await SPyVM.async_new()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/spy/vm/vm.py", line 183, in async_new
    ll = await LLSPyInstance.async_new(llmod)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.13/site-packages/spy/llwasm/emscripten.py", line 77, in async_new
    return cls(llmod, hostmods, instance=instance)
  File "/lib/python3.13/site-packages/spy/libspy/__init__.py", line 152, in __init__
    self.str_layout = StrLayout(*layout)
                      ~~~~~~~~~^^^^^^^^^
TypeError: spy.libspy.StrLayout() argument after * must be an iterable, not NoneType
```

paugier:
Fixes #578

The issue was understood by Claude, who also proposed the fix.

4 calls instead of one but at least it works also for Emscripten and therefore it fixes the playground!

---

**Description by Claude**

### Root cause

`_spy_StrObject_layout()` and `_spy_BytesObject_layout()` are C functions that return a struct by value (4 × size_t fields = 16 bytes on wasm32).

In WebAssembly, structs larger than a primitive scalar can't be returned directly. Emscripten uses the sret (struct return) ABI: the caller allocates space and passes a hidden pointer as the first argument; the function writes the struct there and returns void. From the Python/Pyodide side, the function returns None/undefined — so `StrLayout(*layout)` blows up with `TypeError: argument after * must be an iterable, not NoneType`.

This was never a problem on the wasmtime backend because wasmtime handles multi-value returns differently (it unpacks struct returns into a tuple).

### Fix

Replaced the two struct-returning functions with individual uint32_t-returning accessor functions — one per field — in both the C headers and implementations:

```
_spy_StrObject_size/length_offset/hash_offset/utf8_offset(void) → uint32_t
_spy_BytesObject_size/length_offset/hash_offset/data_offset(void) → uint32_t
```

`uint32_t` is a scalar type and crosses the WASM/Emscripten boundary correctly in both backends. The Python side now calls each function individually to build the StrLayout/BytesLayout dataclasses.

The `_spy_StrObject_Layout` typedef can be removed from the headers entirely since nothing references it anymore (done).

### DIFF
--- a/spy/libspy/__init__.py
+++ b/spy/libspy/__init__.py
@@ -116,7 +116,7 @@ class LibSPyHost(HostModule):
 
 @dataclass
 class StrLayout:
-    # See also the struct _spy_StrObject_layout in str.h
+    # See also spy_StrObject in str.h
     size: int
     length_offset: int
     hash_offset: int
@@ -125,7 +125,7 @@ class StrLayout:
 
 @dataclass
 class BytesLayout:
-    # See also the struct _spy_BytesObject_layout in bytes.h
+    # See also spy_BytesObject in bytes.h
     size: int
     length_offset: int
     hash_offset: int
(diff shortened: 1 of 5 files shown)


antocuni:
> This was never a problem on the wasmtime backend because wasmtime handles multi-value returns differently (it unpacks struct returns into a tuple).

I fear this might become a problem later. If we cannot find a way for emscripten to support multi-value, this seems a reasonable workaround. But ideally, I'd like to find a way to have emscripten to follow the same ABI as clang/wasi, because it seems easier to manage.

It _should_ be somehow possible, because under the hood emcc still uses clang.
What is strange is that we *do* pass the flag to emscripten already; maybe it's swallowed by emcc and not forwarded to clang?
https://github.com/spylang/spy/blob/47692e7ab6d08cfa410dcdb572e754265548cfc4/spy/build/flags.py#L31-L50

paugier:
Superseded by #583

antocuni:
closing as this has been superseded.

Issue #564, opened 2026-06-14
PanicError: ptr_copy_slice out of bounds (playground)


paugier:
In the playground,

```python
def main() -> None:
    print("a" + "b")
```

gives:

```
Traceback (most recent call last):
  * str_bytes::main at /home/pyodide/str_bytes.spy:3
  |   - str supports: +, *, ==, !=, [], len(), replace(), upper(), isascii()
  |           |_______|
  * _str::methods::__add__ at /lib/python3.13/site-packages/stdlib/_str.spy:162
  |         ptr_copy_slice(out.utf8, 0, a.length, a.utf8, 0, a.length)
  |         |________________________________________________________|

PanicError: ptr_copy_slice out of bounds
```

antocuni:
Closing this as the bug is done in recent versions of the playgrond.
Probably it was fixed by #584

PR #584, opened 2026-06-19
playground: use debug build of libspy

paugier:
Workaround #564 and #578 (no -flto in debug build)

The playground has to use the debug build of libspy. The interpreter is only tested with the debug build. This is what works!

#564 is the manifestation of a bug related to using the interpreter with the release mode, for which pointers have no length.

### DIFF
--- a/playground/Makefile
+++ b/playground/Makefile
@@ -3,7 +3,7 @@
 # Paths relative to playground directory
 ROOT_DIR := ..
 LIBSPY_DIR := $(ROOT_DIR)/spy/libspy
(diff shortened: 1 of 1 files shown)


antocuni:
happy to merge this, but then maybe we should revert #583?

antocuni (approved):
LGTM. Feel free to merge.

paugier:
I merge since it should have a direct good effect on the playground!

Let's take the time to think a bit more if https://github.com/spylang/spy/pull/583 should be reverted.

PR #586, opened 2026-06-20
Playground layout: file explorer instead of buttons

paugier:
As suggested by @antocuni, the buttons at the top are removed and replaced with a UI looking like a file explorer.

It looks quite nice :slightly_smiling_face: 

Done with Claude.

### DIFF
--- a/examples/1_high_level/factorial.spy
+++ b/examples/1_high_level/factorial.spy
@@ -1,7 +1,7 @@
 """
 Things to notice:
   - `i32` is a concrete 32-bit integer type (unlike Python's arbitrary-precision `int`).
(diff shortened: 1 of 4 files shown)


antocuni:
I like it!
Two things: `hello.spy` says this:
> The playground presents example files organized in 4 categories. You can discover
SPy by following the proposed progression from left to right.

we should probably turn into "from top to bottom".

The second is that the new layout is not mobile friendly:
image

This is not necessarily a merge blocker, but if claude can find a quick&easy fix, might be worth doing it.

paugier:
It's now responsible.

I merge since it's a net improvement.

Issue #474, opened 2026-04-20
Improve readability of spy redshift output


paugier:
For my experiment on loop fusion with SPy, I need to get a nice, human-readable redshift output.

First, it should be possible to ask for a simplified short version of the redshift output. For the loop fusion demo, I used (https://github.com/paugier/spy-demos/blob/loop-fusion/loop-fusion/simplify_rs_output.py):

```python
# `name` is the file name without `.spy`.
for name, content in files.items():
    files[name] = (
        content.replace(name + "::", "")
        .replace("::Self", "")
        .replace(str(Path.home()), "~")
    )
```

`Self` is the name of the concrete types created in type factories. We don't need `ndarray[f64, 3]::Self`, especially when it's repeated several times. This is about the representation of generic function returning a type. A simple solution would be to treat the word `Self` as special and not display "::Self" in this case.

Second, the redshift output should be formatted to avoid very long lines to obtain something like this rather that the one-line version:

```
`unsafe::gc_ptr[f64]::getitem_byval`(
    `unsafe::ptr_getfield_byval[unsafe::gc_ptr[f64]]`(
        `ndarray[f64]::__get___ll____`(
            `LazyAdd[f64, ndarray[f64], ndarray[f64]]::__get_left__`(
                `LazyAdd[f64, LazyAdd[f64, ndarray[f64], ndarray[f64]], ndarray[f64]]::__get_left__`(expr))), 'items', 8), i)
+ `unsafe::gc_ptr[f64]::getitem_byval`(
    `unsafe::ptr_getfield_byval[unsafe::gc_ptr[f64]]`(
        `ndarray[f64]::__get___ll____`(
            `LazyAdd[f64, ndarray[f64], ndarray[f64]]::__get_right__`(
                `LazyAdd[f64, LazyAdd[f64, ndarray[f64], ndarray[f64]], ndarray[f64]]::__get_left__`(expr))), 'items', 8), i)
+ `unsafe::gc_ptr[f64]::getitem_byval`(
    `unsafe::ptr_getfield_byval[unsafe::gc_ptr[f64]]`(
        `ndarray[f64]::__get___ll____`(
            `LazyAdd[f64, LazyAdd[f64, ndarray[f64], ndarray[f64]], ndarray[f64]]::__get_right__`(expr)), 'items', 8), i)
```

antocuni:
I agree that very long FQNs are not very readable but I cannot think of a "general" fix to improve the situation.

[...]

PR #590, opened 2026-06-20
Redshift: no current module name in FQN by default?

paugier:
Related to #474

It seems to me that the current module name should not be displayed in FQN by default.

For real world code (longer than simple examples), it considerably simplify the output which becomes much more readable.

Of course, the full FQN can be obtained with `--full-fqn`.

For example, applied to one of the example:

```
# before
$ spy rs examples/4_advanced/myarray.spy | wc -m
14156
# with this PR
$ spy rs examples/4_advanced/myarray.spy | wc -m
13211
```

And for a typical function:

```
# before
def `myarray::Array1d[i32]::__ne__`(a: `myarray::Array1d[i32]`, b: `myarray::Array1d[i32]`) -> bool:
    return `operator::bool_not`(`unsafe::gc_ptr[myarray::ArrayData[i32]]::__eq__`(`myarray::Array1d[i32]::__get___ll____`(a), `myarray::Array1d[i32]::__get___ll____`(b)))
 
# with this PR
def `Array1d[i32]::__ne__`(a: `Array1d[i32]`, b: `Array1d[i32]`) -> bool:
    return `operator::bool_not`(`unsafe::gc_ptr[ArrayData[i32]]::__eq__`(`Array1d[i32]::__get___ll____`(a), `Array1d[i32]::__get___ll____`(b)))
```

### DIFF
--- a/spy/backend/spy.py
+++ b/spy/backend/spy.py
@@ -77,8 +77,8 @@ class SPyBackend:
             ):
                 aliases.append((attr, w_obj))
         for attr, w_obj in aliases:
-            fqn_str = w_obj.fqn.human_name(self.vm)
-            self.out.wl(f"{attr} = `{fqn_str}`")
+            fqn_str = self.fmt_fqn(w_obj.fqn)
+            self.out.wl(f"{attr} = {fqn_str}")
         if aliases:
             self.out.wl()
 
@@ -145,7 +145,7 @@ class SPyBackend:
         if isinstance(w_obj, W_Type) and issubclass(w_obj.pyclass, W_InterpList):
             # this is a ugly special case for now, we need to find a better
             # solution
-            return w_obj.fqn.human_name(self.vm)
+            return self.fmt_fqn(w_obj.fqn)
         #
         # this assumes that w_obj has a valid FQN
         fqn = self.vm.reverse_lookup_global(w_obj)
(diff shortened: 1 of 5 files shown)


paugier:
Alternatively, this behavior could be a new triggered by a new redshift option?

antocuni:
+1 for the goal, but I'm not sure this is the right approach: we already have `FQN.human_name`, let's try to make it even more human.

I propose the following:

- `FQN.human_name` takes an optional `parent` argument, which is another FQN, and strips the parent from the output;
- the spy backend keeps track of the current parent, and pass it to `human_name`.

Two things which I like of this approach are:
1. we can strip the parent "properly", without having to resort to fragile regexps.
2. it works for any kind of nesting, not only for module names

One open question: how do we format a stripped FQN? Imagine this:
```python
# mod.spy
@struct
class T:
    def bar(self) -> None:
        pass
```

One option is to just strip the parent and be happy; another is to use some kind of "relative to here" marker:
```python
@struct
class `./T`:
    def `./bar`(self) -> None:
        pass
```

Just to be clear, I don't like the `./` syntax, but it's to show what I mean.
The problem is that I cannot find any syntax which I like. Some proposals:
- `./T`
- `.::T`
- `::T`
- `.T`
- `^T`

Of these, I think I like `.T` a bit more than the others.

paugier:
I changed the implementation with `_human_render(self, parent)`.

I didn't change yet `class T:` in 

```
class `.T`:
```

I'm not sure that it is so useful. What does it add to have \`.T\` instead of `T`? Isn't it implicit that `T` is defined in the current module? Simple names (without `::`) are either defined in the current module or builtins. It's simple and clear. And we have `--full-fqn` when needed.

PR #592, opened 2026-06-20
README: list optional deps, gcc/clang and emcc

### LINKED ISSUE #541: Add emscripten as a prerequisite in README
I tried to use pixi to make spy, but emscripten was not installed. It should be added to README. After installing emscripten, it built and the tests are able to run.

paugier:
Fixes #541


antocuni (approved):
LGTM

PR #594, opened 2026-06-21
unsafe: use GC_MALLOC_ATOMIC when possible

paugier:
I isolate this change in SPy core from my PR on benchmark (#565) to ease the review.

This commit is a performance optimization using GC_MALLOC_ATOMIC when possible.

---

Explanation from #565:

I propose an improvement: by using `GC_MALLOC_ATOMIC` for the allocation of types that do not contain pointer, the slowness compared to Numpy disappears.

Claude helps me for the  `GC_MALLOC_ATOMIC` part.


spy/libspy/src/unsafe.c
```diff
@@ -5,6 +5,11 @@ spy_gc_alloc(size_t size) {
     return malloc(size);
 }
 
+void *
+spy_gc_alloc_atomic(size_t size) {
+    return malloc(size);
+}
+
```
antocuni, line 7: ouch, I forgot that these functions exist 🤦‍♂️ .
I think that what happens is the following:
  - `spy_gc_alloc` and `spy_raw_alloc` (and now `spy_gc_alloc_atomic`) are always compiled inside libspy
  - when we `#define SPY_GC_BDWGC`, they are `#defined` to `spy_gc_alloc_bdwgc`, so they are effectively shadowed.

This seems very fragile.
It's probably matter for a different PR, but I think we should:
- declare that you must define one and exactly one of `SPY_GC_NONE` or `SPY_GC_BDWGC` (similar to what we do for `SPY_RELEASE`/`SPY_DEBUG`) and then do something like this:

```C
#ifdef SPY_GC_NONE
#  define spy_gc_alloc spy_nogc_alloc
#elif SPY_GC_BDWGC
#  define spy_gc_alloc spy_bwgc_alloc
#else
#  error "no GC selected"
```

And then we probaby need to modify the interpreter to explicitly call `spy_nogc_alloc` (which is what we do already now but at least it would be explicit).

spy/backend/c/cstructwriter.py
```diff
@@ -183,6 +184,11 @@ def emit_PtrType(self, fqn: FQN, w_ptrtype: W_PtrType) -> None:
         self.tbh_fwdecl.wl()
 
         memkind = w_ptrtype.memkind
+        if memkind == "gc" and not contains_gc_ptr(w_itemT):
+            # T contains no gc_ptr/gc_ref, so it's safe (and faster) to
+            # allocate it with GC_MALLOC_ATOMIC: the collector doesn't need
+            # to scan it. See spy/libspy/include/spy/unsafe.h.
+            memkind = "gc_atomic"
```
antocuni: I know that this is the term used by BDWGC but I'm not sure I like `atomic`, as it's non obvious at first and reminds me of concurrency.

The ultimate plan is to use [whippet](https://github.com/wingo/whippet), which calls this feature `pointerless`. So propose to follow whippet's naming, as it's clearer and more aligned to our future needs.

spy/libspy/include/spy/unsafe.h
```diff
@@ -43,7 +57,17 @@ spy_gc_alloc_bdwgc(size_t size) {
...
+
+   MEMKIND is one of "raw", "gc", or "gc_atomic":
+     - "raw"       -> spy_raw_alloc        (plain malloc, never collected)
+     - "gc"        -> spy_gc_alloc         (GC_MALLOC: zeroed, scanned)
+     - "gc_atomic" -> spy_gc_alloc_atomic  (GC_MALLOC_ATOMIC: not zeroed, not
+                                             scanned; only for pointer-free T)
+   PTR$alloc expands MEMKIND via spy_##MEMKIND##_alloc(...); the alias below
+   makes spy_gc_atomic_alloc resolve to spy_gc_alloc_atomic.
 */
+#define spy_gc_atomic_alloc(size) spy_gc_alloc_atomic(size)
```
antocuni: if we pass an explicit ALLOC_FUNC, then we can probably kill this weird macro

spy/vm/modules/unsafe/mem.py
```diff
@@ -43,6 +43,13 @@ def w_gc_alloc(vm: "SPyVM", w_T: W_Type) -> W_Dynamic:
     assert isinstance(w_ptrtype, W_PtrType)
     ITEMSIZE = sizeof(w_T)
 
+    # if T contains no gc_ptr/gc_ref, we can use the "atomic" allocator
+    # (GC_MALLOC_ATOMIC instead of GC_MALLOC in the bdwgc backend): the
+    # collector doesn't need to scan the result for pointers, and it's
+    # significantly faster. See spy/libspy/include/spy/unsafe.h and
+    # spy/backend/c/cstructwriter.py for the C-backend equivalent.
+    alloc_name = "spy_gc_alloc" if contains_gc_ptr(w_T) else "spy_gc_alloc_atomic"
```
antocuni: I like `pointerless` even more now. This would read as:
```
    alloc_name = "spy_gc_alloc" if contains_gc_ptr(w_T) else "spy_gc_pointerless_alloc"
```

spy/vm/modules/unsafe/misc.py
```diff
@@ -27,3 +27,45 @@ def sizeof(w_T: W_Type) -> int:
...
+    if isinstance(w_T, (W_PtrType, W_RefType)):
+        return True
+
+    if w_T is B.w_str:
+        return True
+
+    if isinstance(w_T, W_StructType):
+        return any(contains_gc_ptr(w_field.w_T) for w_field in w_T.iterfields_w())
+
+    return True
```
antocuni: I think it would be better to be more conservative here, and `assert False` or `raise NotImplemented` if we encounter an unknown type.
I understand it's a bit annoying but at the same time will help us a lot in the future.
Imagine to add e.g. the type `i16`: if you don't add it here, it would silently become "pointerful", but we won't know until we hit a weird performance problem.

spy/vm/modules/unsafe/ptr.py
```diff
@@ -50,7 +50,7 @@
     from spy.vm.vm import SPyVM
 
 
-MEMKIND = Literal["raw", "gc"]
+MEMKIND = Literal["raw", "gc", "gc_atomic"]
```
antocuni: I'm not sure this is the right level of abstraction.
From the point of view of SPy, there are only two kind of allocations, raw and gc.
Then for the GC we choose to use a faster way of allocating, but it's an implementation detail IMHO.

spy/libspy/include/spy/unsafe.h
```diff
@@ -43,7 +57,17 @@ spy_gc_alloc_bdwgc(size_t size) {
...
    array it points to, and every access is checked. The length is expressed in
    number of items, NOT size in bytes.
+
+   MEMKIND is one of "raw", "gc", or "gc_atomic":
+     - "raw"       -> spy_raw_alloc        (plain malloc, never collected)
+     - "gc"        -> spy_gc_alloc         (GC_MALLOC: zeroed, scanned)
+     - "gc_atomic" -> spy_gc_alloc_atomic  (GC_MALLOC_ATOMIC: not zeroed, not
+                                             scanned; only for pointer-free T)
+   PTR$alloc expands MEMKIND via spy_##MEMKIND##_alloc(...); the alias below
+   makes spy_gc_atomic_alloc resolve to spy_gc_alloc_atomic.
```
antocuni: see my comment below but I think that `memkind` is the wrong level of abstraction.
I think the basic structure of the macro is fine, we just need to find something different as a name.
Since memkind is used only to select the alloc functions, what about `ALLOC_FUNC`?

```
#define _SPY_PTR_FUNCTIONS_UNCHECKED(ALLOC_FUNC, PTR, T)                                  
```

where ALLOC_FUNC can be `gc_alloc`, `raw_alloc` or `gc_alloc_pointeless`?

antocuni (commented):
sorry for taking so long but as you see this is a delicate piece of machinery and I wanted to have time to review it properly.

The general idea is good and correct.
I left some comments about the actual details of how to implement it and/or naming conventions, looking forward to hear your thoughts.

spy/libspy/src/unsafe.c
paugier: I agree that it would be better. It seems that this belong to another PR though. Do you want to do it? Or should I try?

antocuni: Feel free to try :)

spy/libspy/include/spy/unsafe.h
paugier: Done. That's indeed much better like that!

spy/vm/modules/unsafe/ptr.py
paugier: Done.

spy/vm/modules/unsafe/mem.py
paugier: Now, I don't even touch this file. Anyway, the interpreter always just call `spy_nogc_alloc`.

spy/libspy/include/spy/unsafe.h
paugier: Done.

spy/backend/c/cstructwriter.py
paugier: `pointerless` is indeed nicer.

spy/libspy/src/unsafe.c
paugier: This was done in #610 (merged).

paugier:
I finally took the time to rebase and take into account your remarks. I think it is ready for another round.

paugier:
Before:
```
(.venv) pierre@gre:~/dev/spy-benchmarks/array_simple$ make benchspy 
spy build --release bench.spy
[release] build/bench 
./build/bench
spy add:# 0.247734 s
spy add inplace:# 0.0497169 s
spy empty:# 0.104048 s
spy zeros:# 0.113072 s
```

With this PR:

```
(.venv) pierre@gre:~/dev/spy-benchmarks/array_simple$ make benchspy 
spy build --release bench.spy
[release] build/bench 
./build/bench
spy add:# 0.0720551 s
spy add inplace:# 0.0333464 s
spy empty:# 0.00163627 s
spy zeros:# 0.0290363 s
```

antocuni (approved):
LGTM.
Sorry it took so long :(

+1 to merge it.

PR #598, opened 2026-06-23
Implement str.find()

shalabhc:
The `str.find()` metafunc dispatches on arity and matches the Python `find` signature.

A simple algorithm is implemented based on some benchmarks. An initial tight scan for the possible start of match followed by another loop to check the rest of the match.


stdlib/_str.spy
```diff
@@ -406,6 +445,52 @@ def _ll_isspace(ch: u8) -> bool:
...
+    if start < 0:
+        start = start + n
+        if start < 0:
+            start = 0
+    if end < 0:
+        end = end + n
+        if end < 0:
+            end = 0
+    if end > n:
+        end = n
```
JeffersGlass: Could use a comment hear about the endpoint clamping mentioned in the PR.

stdlib/_str.spy
```diff
@@ -406,6 +445,52 @@ def _ll_isspace(ch: u8) -> bool:
...
+    last = end - m
+    pos = start
+    while pos <= last:
+        candidate = self._find_byte(first, pos, last+1)
+        if candidate == -1:
+            break
+        if self._exact_match(sub, candidate):
+            result = candidate
+            break
+        pos = candidate + 1
```
JeffersGlass:
```suggestion
    while pos <= last:
        if self._exact_match(sub, pos):
            return pos
        pos += 1
```
I don't think we gain anything by using `_find_byte`, since that byte just gets rechecked by `_exact_match` anyway.

stdlib/_str.spy
```diff
@@ -396,6 +396,45 @@ class methods:
...
 
+    def _find_byte(self: str, ch: u8, start: i32, end: i32) -> i32:
+        ll = StrObject.from_str(self)
+        i = start
+        while i < end and not ll.utf8[i] == ch:
+            i = i + 1
+        result: i32 = -1
+        if i < end:
+            result = i
+        return result
```
JeffersGlass: See comment in `_str_find` below - I don't think we need this function, since its work just ends up getting duplicated by `_exact_match`

stdlib/_str.spy
```diff
@@ -396,6 +396,45 @@ class methods:
...
+        if idx + m > ll.length:
+            result = False
+        else:
+            i = 0
+            while i < m:
+                if not ll.utf8[idx + i] == lsub.utf8[i]:
+                    result = False
+                    break
+                i = i + 1
+        return result
```
JeffersGlass:
```suggestion
    def _exact_match(self: str, sub: str, idx: i32) -> bool:
        # True iff `sub` occurs in self starting exactly at idx
        ll = StrObject.from_str(self)
        lsub = StrObject.from_str(sub)
        m = lsub.length
        if idx + m > ll.length:
            return False
        else:
            i = 0
            while i < m:
                if not ll.utf8[idx + i] == lsub.utf8[i]:
                    return False
                i = i + 1
        return True
```
A little style thing, just removing named variables from places that are only used as return values.

JeffersGlass (changes requested):
This is really quite nice! A couple of style things and comment requests here.

The one more substantive change is I think that `_find_bytes` is redundant - since `_exact_match` comapres all the bytes again anyway and bails early, we don't gain anything by pre-checking the first byte separately.

stdlib/_str.spy
shalabhc: done

shalabhc:
> The one more substantive change is I think that _find_bytes is redundant

It's not redundant but it's true that just the first byte is checked twice. `_find_byte` scans looking for the start of the match and `_exact_match` checks the match itself.

But I've just rolled the entire logic in the find and removed the two util functions. Let me know if this works.

stdlib/_str.spy
shalabhc: done

JeffersGlass (approved):
This looks great! Made one tweak to a comment that referenced the old implementation - will merge as soon as tests are green again. 

Thank you for adding this!

PR #599, opened 2026-06-23
Fix operator associativity in C and SPy backends (#593)

### LINKED ISSUE #593: Wrong precedence when compiling "a - b - c" (spy backend and C backend)
`a - (b - c)` gets redshifted into `a - b - c` which is wrong.

This is a display issue only, `rs -x` produces the right result, indentical to `interp`

```python
# x.py

def f(a: i32, b:i32, c:i32) -> i32:
    return a - (b - c)

def main() -> None:
    print(f(100, 10, 5))
```

```
% spy /tmp/x.spy 
95

% spy rs -x /tmp/x.spy
95

% spy rs /tmp/x.spy 
def f(a: i32, b: i32, c: i32) -> i32:
    return a - b - c

def main() -> None:
    `_print::println[i32]`(`x::f`(100, 10, 5))
```

shalabhc:
Fixes #593

The C and SPy backends only parenthesized operands by strict precedence, ignoring associativity. As a result `a - (b - c)` was emitted as `a - b - c`, which regroups to `(a - b) - c` and computes the wrong value in C and prints the wrong expression in redshift. The redshifted ast was evaluated correctly, however.

Both backends now also parenthesize a same-precedence operand if needed. Eg

`a - (b - c)` is

```python
BinOp('-')
   left: "a" 
   right: BinOp('-')
      left: "b"
      right: "c"
```

Since precedence of root and right arg is same (both `-`), and the op is left associative, we insert `()` on the right to avoid accidental regrouping on the left.

A second bug was also fixed. Redshift converts operators to functions, eg `a - b` becomes `sub(a, b)`. This prints as expected in fqn format, but in short format it is converted back to a `BinOp` on the fly. We end up with a partially converted tree, roughly the parse equivalent of `a - sub(b, c)`. This breaks the logic because that expression doesn't need `()` but eventually it gets formatted as `a - b - c`. We now eagerly convert the subtree nodes to BinOp before applying rules.


antocuni (approved):
LGTM, thank you and sorry it took so long to review.
Unfortunately I cannot merge as it has conflicts.

@JeffersGlass @paugier could you please proceed with merging as soon as the conflicts are resolved? I won't be much online in the next days and I'd like to avoid this PR to become stale again, especially because it fixes a very nasty bug!

paugier:
I added a merge commit and merged all that in main. Thanks @shalabhc! One less bug in the universe.

PR #600, opened 2026-06-24
Update uv installation instructions in README

Viriathus1:


paugier:
@Viriathus1 Could you elaborate why you think this change is useful?

I agree that we could support a standard uv workflow. But then, I guess we should change a bit more the README and the instructions. And also version the uv.lock file?

paugier:
Also, if we choose to support a standard uv workflow, I think this works:

```
[tool.uv]
all-extras = true
```

JeffersGlass:
+1 adding the lockfiles to `gitignore`.

I think the readme should just include `uv sync --extra dev`. Devs don't necessarily needs to install `docs` and `bench` (when it exists) to get started, and those extras are or can be mentioned in their contributing sections.

Sidenote, but at some point I think we should migrate from optional dependencies to [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) - optional dependencies are for user-facing extras. That would also make the command just `uv sync` (which includes the `dev` group by default).

Viriathus1:
Hi @paugier you bring up a good point, may as well take it all the way. I'll skip the virtual environment creation and just use uv run. Did you have anything else in mind?

For the lockfile, imo we should keep only one in source control. `pixi.lock` is already committed and used by GH actions so we can use that as the source of truth, other generated lock files shouldn't be versioned unless CI depends on them.

@JeffersGlass you're right the one I suggested includes every option which isn't necessary. I believe just having `uv sync` is cleaner though so we can just do it now, what do you think?

paugier:
We have to answer to interesting questions.

- lockfile

  I don't have a very strong opinion on that but it seems to me that it would be useful and cleaner to version `uv.lock` and use it in the CI. This is the recommended practice for uv and unfortunately, uv do not really support `pylock.toml` format for `uv sync`. `pixi.lock` cannot be used by other tools and it is used only in the CI workflow checking Pixi (which is very limited).
  
- I don't really like the dev `[project.optional-dependencies]` and I agree with @JeffersGlass that we should use `[dependency-groups]`. pip even supports them now  https://pip.pypa.io/en/latest/user_guide/#dependency-groups. However, I'm not sure we can use a dependency group "from outside", and I fear that what we do in https://github.com/spylang/benchmarks is not supported:

```
dependencies = [
    "numpy>=2.5.0",
    "spylang[dev]",
]

[tool.uv]
package = false

[tool.uv.sources]
spylang = { path = "../spy", editable = true }
```

Therefore, it seems useful to keep `dev` in `[project.optional-dependencies]` (and doc can be moved in `[dependency-groups]`. But there might actually be a way to use a SPy dev `[dependency-groups]` from spylang/benchmarks. We need to investigate and find a good solution for both repositories.

Note that I don't think there will be a `bench` group in spylang/spy repo since we have now spylang/benchmarks which has its own environment.

We can also think about the experience we want for developers. I think it is reasonable to ask for uv or pixi for people who want to install SPy in developer mode in a controlled and simple way. (I mean that people can continue to only use pip but it can be slightly more complicated then.)

I think `uv sync` should install a clean and versioned environment with all the standard dependencies to develop SPy and its documentation.

Installing SPy without the dev and/or doc group is less common and useful, and then it can be done with slightly more complicated commands for example involving `pip` and `uv pip`.

On Linux,

```
# without the doc group
635M	.venv/
# with the doc group
751M	.venv/
```

It's 20% larger. Personally, I don't care about ~ 100 MB more in the SPy dev env and I prefer to have all needed in the env. However, it can be different for other people...

Last thing, currently we have this weird installation command for a Python package `make -C spy/libspy`. In the longer term, we should think about how to package SPy. `ziglang` is a build dependency for SPy (to build libspy.wasm). In contrast, I don't think `build` is a real runtime dependency for SPy. I know that it is used for the playground to build the SPy wheel. Anyway, this is clearly a different subject.

JeffersGlass:
I don't think `spylang/benchmarks` actually requires `spylang[dev]` though... as far as I can tell, the dependencies should just be:

```toml
# spylang/benchmarks :: pyproject.toml
dependencies = [
    "numpy>=2.5.0",
    "spylang",
    "pytest",
]
```

Once we have published versions of SPy, the spylang install could be editable or not, based on what the end-user wants to benchmark.

The benchmarks and pytest-based runners seem to run fine with this, and I think is cleaner - `benchmarks` needs `pytest` on its not, _not_ because SPy uses pytest, if that makes sense. So it should be a part of benchmark's dependencies, not pulled in from SPylang. And the rest of SPy's dev dependencies aren't need, as far as I can tell?

paugier:
Good catch. I'm going to try that to be sure but I agree it should work and be cleaner.

paugier:
`spylang/benchmarks` does not use `spylang[dev]` any longer so we can kill that.

Then, why not convert to `[dependency-groups]`? The disadvantage is that the installation with pip is going to be slightly less simple: I guess something like `pip install -e . --group dev` instead of `pip install -e .[dev]`. But for most people, it would just be `uv sync`.

Also, I don't understand the advantage of not versioning `uv.lock` if we start to officially adopt the standard uv workflow. Why not version this file and use uv in the CI? (except of course in pixi-checks.yml)

antocuni:
Just a few comments on my side.

I don't have any strong opinion on any of these subjects, so if there is consensus between the three of you (@paugier @JeffersGlass and @Viriathus1), I'll be happy with whatever decision you'll take w.r.t. packaging/installing.  I'm actually quite happy not to have to think about that 😅.

That said, the following are "not strong opinions" :).
First, a more general remark: either we support a tool or we don't.
If we decide that we support uv-based workflows, I think we should commit `uv.lock`.
If we decide that we support a pixi-based workflows, we should commit `pixi.lock`.

CI should decide which of the two is the "main" and use that. The other is still supported but we just do a smoke test.

The hard part is to ensure that uv.lock and pixi.lock are more-or-less synchronized. Maybe it's enough to add a test to check that `git log -1` of both is the same (i.e., they were modified together by the same commit).

I'm +0.5 to remove support for pip, if it makes the life easier and it reduces the maintenance burden.

paugier:
I'm going to propose another PR to support a uv-based workflow. I think I have a good idea of what should be done.

Viriathus1:
I initially supported just using pixi because I assumed it was decided somewhere else that it would be the main tool because the lock file was committed. Going forward, I think we should have one canonical lockfile and my preference is to make uv the primary workflow. Keeping both uv and pixi as first-class workflows increases CI overhead and creates potential lockfile synchronisation drift. We can still keep pixi in the docs as a convenience option for developers who prefer not to manage non-Python dependencies like bdw-gc manually.

Anyhow I assume @paugier wants to keep both lock files, @antocuni and @JeffersGlass what do you guys want to do?

paugier:
> Anyhow I assume @paugier wants to keep both lock files

Yes, I think there are advantages to really support these two workflows and therefore to version the two lock files and no strong disadvantages:

**Pros:**

- Pixi can be used by people that cannot install `libgc-dev` system-wide, which is not so uncommon.
- Versioning pixi.lock and uv.lock and keeping them sync with pyproject.toml is very simple (done in #611). From the point of view of SPy developers who modified the dependencies in pyproject.toml, it is just `pixi lock && uv lock` and commit.
- It is actually a very nice message from the point of view of a project supported by Anaconda to support a Pixi workflow.

**Cons:**

- CI overhead ([pixi-checks.yml](https://github.com/spylang/spy/blob/main/.github/workflows/pixi-checks.yml)) but it is already there and it is very short (typically 1m20s).
- Relative complication for contributors modifying the dependencies (They need to run `pixi lock && uv lock` and commit).

We could

- add another check in pixi-checks.yml based on git to avoid any lockfile synchronisation drift.
- add a very simple Makefile with common maintenance commands, for example

```
lock:
    pixi lock
    uv lock

format:
    ruff format
    ruff check --fix
    spy format

libspy:
    make -C spy/libspy
```

antocuni:
> * Pixi can be used by people that cannot install `libgc-dev` system-wide, which is not so uncommon.

I think this is the main point in favour of pixi.
Going forward, I expect we will have more and more spy packages which wrap C libraries, and conda/pixi offers an easy way to install them.

Unfortunately at this stage conda/pixi cannot be _the_ answer to spy packaging problems, because for any C library we would like to have at least:
1. the .a file compiled to wasm/wasi (needed for the interp and for wasi compilation)
2. the .a or .so file compiled to emscipten
3. the .a file compiled for native (optional, but needed if you want a fully static build)
4. the .so file compiled for native

AFAIK conda/pixi offer only artifact (4).

Viriathus1:
I suppose the decision was already made given the merging of #611 so I'm closing this.

PR #603, opened 2026-06-24
Basic documentation for Struct Classes

JeffersGlass:
Adds the beginnings of documentation for struct classes. ([Documentation Preview Page](https://spylang.github.io/docs/pr-preview/pr-603/howto/structs/))

I'm working on expansions to this - special methods like `__convert_to/from__`, `__getattr__`/`__getattribute__`, methods that are similar to CPython (`__eq__`, `__add__` etc), which are more like CPython's [data model documentation](https://docs.python.org/3/reference/datamodel.html). But I wanted to get this out there before it got too big to conveniently review.

### DIFF
--- a/docs/src/howto/structs.md
+++ b/docs/src/howto/structs.md
@@ -155,3 +155,7 @@ Methods may also be [blue functions](../reference/spy_builtin_functions.md#blue)
 ## Inheritance
 
 Inheriting from a base class is not currently implemented in SPy.
+
+## Structs as Arguments
+
+When used as a function argument, Structs are passed by value, meaning a copy of the struct is created for each call. The performance implications of this should be kept in mind for large structs.
--- a/docs/src/howto/structs.md
+++ b/docs/src/howto/structs.md
@@ -90,6 +90,26 @@ def main() -> None:
     print(p.name, "is", p.age[0], "years old") # Alice is 100 years old
 ```
 
+## Pointers to Structs
+
+If a struct is manually allocated using `gc_alloc` or similar, its attributes can be mutated. See the [low level memory docuementary on heap-allocated structs](../llmem.md#heap-allocated-structs) for more info.
+
+```py
+from unsafe import gc_alloc, gc_ptr
+
+@struct
+class Point:
+    x: int
+
+# spy build foo.spy
+def main() -> None:
+    p = gc_alloc[Point](1)
+    p.x = 1
+    print(p.x)
+    p.x = 2
+    print(p.x)
+```
+
 ## Constructors
 
 Structs use a default constructor which populates all their attributes in-order, and all attributes must be provided each time the default constructor is called. E.g. the example above, the constructor `p = Person('Alice', 99, [])` requires an empty list to be passed for the `books` attribute. We can call this constructor explicitly using the `__make__` method:
--- a/docs/src/howto/structs.md
+++ b/docs/src/howto/structs.md
@@ -144,6 +144,46 @@ def main() -> None:
     print(p.name, "has", len(p.books), "books") # Alice has 0 books
 ```
 
+By convention, if a struct is simply being used to wrap a lower-level type (to add constructors, methods, etc), that lower level type is denoted as `__ll__`.
+
+See the [smallpoint example](https://github.com/spylang/spy/examples/3_low_level/smallpoint.spy) for a more in-depth example and explanation. See the implementations of the [list](https://github.com/spylang/spy/stdlib/_list.spy), [dict](https://github.com/spylang/spy/stdlib/_dict.spy), and [file](https://github.com/spylang/spy/stdlib/_file.spy) objects in spy for usage of this pattern in the SPy standard library.
+
+/// info
+Eventually, the `__ll__` notation may likely be special-cased to make it a truly private attribute within the struct. Currently, it is just a naming convetion.
+///
+
+```py
+from unsafe import gc_ptr, gc_alloc
+
+@struct
+class BookData:
+    name: str
+    stars: int # out of 10
+
+@struct
+class Book:
+    __ll__: gc_ptr[BookData]
+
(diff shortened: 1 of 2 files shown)


docs/src/howto/structs.md
```diff
@@ -0,0 +1,157 @@
+title: Struct Classes
```
antocuni: I'd call them just "Structs".

The fact that we need a "class" statement is a necessary evil because of Python grammar.
If/when we will be ready to deviate more from CPython grammar, I'd like to be able to write this:
```python

type MyMetaclass Foo:
    x: int
    y: int
```

and then Foo will be created by calling `MyMetaclass` (this is already true in `ASTFrame.exec_ClassDef`).  In this scheme, `class` and `struct` will be just two builtin metaclasses, and `class Foo` would be kept as a syntax sugar for `type class`.

Anyway, that's the far future, but I think it shows why we shouldn't call them "struct classes"

docs/src/howto/structs.md
```diff
@@ -0,0 +1,157 @@
...
+    name: str
+    age: int
+    books: list[str]
+
+def main() -> None:
+    p = Person.__make__('Bob', 6, [])
+    print(p.age) # 6
+```
+
+User-facing constructors can be customized by overwriting the `__new__` method; the `__make__` must be called within to handle the initialization of the complete struct. Note that `__new__` functions like a staticmethod (it does not take a `self` parameter)
```
antocuni:
```suggestion
User-facing constructors can be customized by overwriting the `__new__` method; the `__make__` must be called within to handle the initialization of the complete struct. Note that, as in Python, `__new__` functions like a staticmethod (it does not take a `self` parameter)
```

docs/src/howto/structs.md
```diff
@@ -0,0 +1,157 @@
...
+    p = Person('Alice', 99)
+    print(p.name, "is", p.age, "years old") # Alice is 99 years old
+    
+```
+
+Struct classes may also be defined with the [generic class syntax](../howto/generics.md#generic-class-syntax), which is syntactic sugar for a generic function which defines an internal struct. See the [generics](../howto/generics.md) documentation for more info.
+
+## Attributes
+
+Attributes of struct classes do not support assignment after creation, nor can new attributes be assigned to an instance of a struct class after creation:
```
antocuni:
```suggestion
Structs are shallow immutable. Attributes cannot be set after creation, nor can new attributes be assigned to an instance of a struct class after creation:
```

antocuni (approved):
LGTM.

One note: you might want to add two things:
1. structs are passed BY VALUE, meaning that they are copied around, and thus big structs might have a penalty; `gc_ref[S]` exists, although at this time it's considered a low-level primitive and end users should not use it explicitly.  Later, we will provide better absrtractions on top of this.
2. you might want to document that if you `raw_alloc`/`gc_alloc` a struct, THEN you can mutate its content.

You might also want to mention the `__ll__: gc_ptr[S]` pattern, although this might be material for another PR.

docs/src/howto/structs.md
JeffersGlass: Makes sense! I've changed `struct class` to just `struct` everywhere.

JeffersGlass:
I've addressed all the comments - thank you for them! I'm not super happy with my description of the `__ll__ = gc_ptr[S]` pattern [here](https://github.com/spylang/spy/pull/603/changes/9ad9fbdd57669786d1a2c839a63f58238ed09a41) ... do you have any suggestions on the phrasing for why this pattern is useful?

docs/src/howto/structs.md
```diff
@@ -144,6 +144,46 @@ def main() -> None:
     print(p.name, "has", len(p.books), "books") # Alice has 0 books
 ```
 
+By convention, if a struct is simply being used to wrap a lower-level type (to add constructors, methods, etc), that lower level type is denoted as `__ll__`.
```
antocuni, line 147: What about something like this:
````
## "New type" pattern

A common idiom is to have a struct with a single field of type `T`: this creates a zero-overhead wrapper around `T`, which is a completely new type and makes it possible to enforce compile time type guarantees and/or to add methods and special methods to it.

The zero-overhead is guaranteed by the fact that structs are allocated "on the stack", passed by value, and translated into C structs by the C backend. 

This is similar and inspired by the [newtype idiom in rust](https://doc.rust-lang.org/rust-by-example/generics/new_types.html):

```python
@struct
class Kilometers:
    v: float

    def to_miles(self) -> Miles:
        return Miles(self.v * 1.609344)

@struct
class Miles:
    v: float

    def to_kilometers(self) -> Kilometers:
        return Kilometers(self.v * 1.609344)


def is_a_marathon(distance: Miles) -> bool:
    return distance.v >= 26.2

def main() -> None:
    distance = Miles(30.0)
    distance_km = distance.to_kilometers()
    print("Is a marathon?", is_a_marathon(distance))
    print("Is a marathon?", is_a_marathon(distance_km.to_miles()))
    # print("Is a marathon?", is_a_marathon(distance_km))  ## TypeError
```

## `__ll__` pattern

This is a special case of the "new type pattern".  It is used when you want to create a high-level type in terms of a low-level, possibly unsafe, implementation which is hidden to the end user.

Currently, `__ll__` is just a naming convention but in the future it might become special-cased by the compiler to make it truly private.

INSERT THE BookData example

Note that in this example, `BookData` is potentially a big struct, while `Book` is a wrapper around a *pointer* to it, so more efficient to pass around.
Normal `class`es will be (automatically) implemented using this idiom, but currently they are not supported and must be done manually.

This is also how many builtin types are implemented: `list[T]` is a pointer to `ListData[T]`, `str` is a pointer to `StrObject`, etc.  See `stdlib/*.spy` for more examples.
````

antocuni (approved):
LGTM!
I tried to write an expanded section about the "new type" idiom and the `__ll__` idiom, I hope you like it.

PR #605, opened 2026-06-24
Topologically sort type entries so all types come before their usage

seibert:
Do a topo sort of types before emitting them so types are defined before they are used in nested structures.

This solves a problem I discovered while trying to create a struct that contained a field which was a list of integers.

(This PR was created with AI assistance.)


antocuni:
@seibert as usual claude writes a solution which more or less works but it's not particularly satisfactory (at least for my taste).
I modified the PR to:
  - use `graphlib` instead of rewriting a toposort manually
  - add unit tests for what counts as a "struct dep" and for topo sort itself
  - add a couple of integration tests in the "right" places.
  
Merging as soon as it's green.

PR #606, opened 2026-06-25
Implement str.count()

shalabhc:
Algo is super simple - just uses `find()` underneath. So if `find()` handles code points correctly this will too.

Discovered an interesting case in Python's behavior: apparently you can count the empty string between characters and at the ends:
```python
>>> 'hello'.count('')
6
```

The implementation copies the behavior but has no optimization for this case. Note we can't used len as-is because `start` and `end` may be provided. Byte based indexing will eventually need to be replaced with optimized code point based indexing I think.


JeffersGlass:
This looks great to me! Thank you!

PR #607, opened 2026-06-25
Create a CFuncPtr type and allow red functions to be passed as callbacks to C libraries

seibert:
I am using the newly added external module support to wrap a C library that takes callbacks via function arguments.  I'd like to pass a red function from SPy to this module.

_Note: This is a design proposal PR to suggest a possible approach that we can discuss.  The implementation was constructed and refined with AI assistance.  Feel free to close this PR or use it as source material for a different implementation._

Overall approach:
* adds a `unsafe.c_func_ptr`: used to construct a type signature for a C function. 
   Example: `CB = c_func_ptr[i32, i32]      # type: int32_t (*)(int32_t)`
* adds a `@unsafe.c_callback` decorator which can be placed on red functions.  Initially I thought this would need to do something functional, but now it is just documentation / validation (it raises an error if you decorate a function that can't be used as a callback).  I'm open to either remove it, or have it do something.
* `W_CFuncPtrType` - the parametric function pointer type.  We cache instances of these so identical C signatures get the same type everywhere and we don't have to check structural identity.
* `W_CFuncPtr` - the value wrapper of a `W_ASTFunc` that knows how to convert from a red function and rejects blue functions.
* The C backend emits the C symbol name for the function directly when it needs to be passed as an argument.

Note that this PR _does not_ allow calling the CFuncPtr from SPy code.  It only allows passing it to C code where it can be called back.  Because of this, the PR augments the `out_of_tree` test to actually show that the callback is functional, rather than create another module to test the callback.  An example of a C callback factory function from that test looks like:

``` python
from unsafe import c_callback, c_func_ptr
from mymod import run_callback

CB = c_func_ptr[i32, i32]

@blue
def make_adder(n: i32) -> CB:
    @c_callback
    def adder(x: i32) -> i32:
        return x + n
    return adder

add5 = make_adder(5)
add10 = make_adder(10)

def run() -> i32:
    return run_callback(add5, 1) + run_callback(add10, 1)
```

Open issues:
* Should `@c_callback` be required to pass a red function as a CFuncPtr?  Right now the code accepts any red function and checks the type signature matches.
* The more logical syntax for making a CFuncPtr type is `CB = c_func_ptr[i32, (i32,)]` where the first parameter is the return type, and the second parameter is a tuple of argument types.  This ran into implementation trouble because there isn't a way to mark a tuple literal as blue.  Either the tuple constructor needs to be pure, or we need to handle tuple literals in a special way.

### DIFF
--- a/spy/backend/c/cbackend.py
+++ b/spy/backend/c/cbackend.py
@@ -14,6 +14,7 @@ from spy.fqn import FQN
 from spy.highlight import highlight_src
 from spy.vm.cell import W_Cell
 from spy.vm.function import W_ASTFunc
+from spy.vm.modules.unsafe.funcptr import W_CFuncPtr, W_CFuncPtrType
 from spy.vm.modules.unsafe.ptr import W_MemLocType
 from spy.vm.object import W_Object, W_Type
 from spy.vm.primitive import W_I32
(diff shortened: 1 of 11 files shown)


seibert:
closing this in favor of #608

PR #608, opened 2026-06-25
allow red functions to be passed as arguments and return values

seibert:
This is a redo of #607 based on out-of-band discussion with @antocuni.  In this version, we don't need a new type for C callbacks, we use the existing `W_FuncType`, and make the minimum changes required to allow red functions to be passed around as arguments.  The end goal is the same: to enable external modules to accept SPy functions as C callbacks.  Calling a function passed as an argument is still not allowed in this PR.

This PR adds a builtin `@functype` decorator to make it easier to define a new function type, so you can write stuff like:
``` python
@functype
def CB(x: i32) -> i32:
    pass

@blue
def make_adder(n: i32) -> CB:
    def adder(x: i32) -> i32:
        return x + n
    return adder
```

(This PR was created with AI assistance.)

### DIFF
--- a/spy/backend/c/cbackend.py
+++ b/spy/backend/c/cbackend.py
@@ -13,7 +13,7 @@ from spy.build.ninja import NinjaWriter
 from spy.fqn import FQN
 from spy.highlight import highlight_src
 from spy.vm.cell import W_Cell
-from spy.vm.function import W_ASTFunc
+from spy.vm.function import W_ASTFunc, W_FuncType
 from spy.vm.modules.unsafe.ptr import W_MemLocType
 from spy.vm.object import W_Object, W_Type
 from spy.vm.primitive import W_I32
@@ -145,7 +145,7 @@ class CBackend:
             modname = fqn.modname
             w_mod = self.vm.modules_w[modname]
             if w_mod.filepath is None and not isinstance(
-                w_obj, (W_MemLocType, W_StructType)
+                w_obj, (W_MemLocType, W_StructType, W_FuncType)
             ):
                 continue
 
@@ -274,19 +274,31 @@ class CBackend:
             # the forward-decl section: it needs T's own typedef to appear
             # first.
             return [w_type.w_itemT.fqn]
+        if isinstance(w_type, W_FuncType):
+            # A function-pointer typedef references its return and param types
+            # by name, so any by-value struct types must be defined first.
+            deps: list[FQN] = []
+            seen: set[FQN] = set()
+            for w_dep in [w_type.w_restype] + [p.w_T for p in w_type.params]:
+                if isinstance(w_dep, W_StructType):
+                    d = w_dep.fqn
+                    if d != fqn and d not in seen:
+                        seen.add(d)
+                        deps.append(d)
+            return deps
         if not isinstance(w_type, W_StructType) or not w_type.is_defined():
             return []
-        deps: list[FQN] = []
-        seen: set[FQN] = set()
+        struct_deps: list[FQN] = []
+        struct_seen: set[FQN] = set()
         for field in w_type.iterfields_w():
             w_fieldT = field.w_T
             if not isinstance(w_fieldT, W_StructType):
                 continue
             dep_fqn = w_fieldT.fqn
-            if dep_fqn != fqn and dep_fqn not in seen:
-                seen.add(dep_fqn)
-                deps.append(dep_fqn)
-        return deps
+            if dep_fqn != fqn and dep_fqn not in struct_seen:
+                struct_seen.add(dep_fqn)
+                struct_deps.append(dep_fqn)
+        return struct_deps
 
     def topo_sort_structdefs(
         self, content: list[tuple[FQN, W_Type]]
--- a/spy/backend/c/cstructwriter.py
+++ b/spy/backend/c/cstructwriter.py
@@ -7,6 +7,7 @@ from spy.backend.c.cffiwriter import CFFIWriter
 from spy.backend.c.context import C_Type, Context
 from spy.fqn import FQN
 from spy.textbuilder import TextBuilder
+from spy.vm.function import W_FuncType
 from spy.vm.modules.unsafe.ptr import W_PtrType, W_RefType
 from spy.vm.object import W_Type
 from spy.vm.struct import W_StructType
@@ -98,6 +99,8 @@ class CStructWriter:
             assert fqn == w_type.fqn  # sanity check
             if isinstance(w_type, W_StructType):
                 self.emit_StructType(fqn, w_type)
+            elif isinstance(w_type, W_FuncType):
+                self.emit_FuncType(fqn, w_type)
             elif isinstance(w_type, W_PtrType):
                 self.emit_PtrType(fqn, w_type)
             elif isinstance(w_type, W_RefType):
@@ -154,6 +157,15 @@ class CStructWriter:
                 tb.wl(f"{c_fieldtype} {w_field.name};")
         tb.wl("};")
         tb.wl("")
+
+    def emit_FuncType(self, fqn: FQN, w_ft: W_FuncType) -> None:
+        c_name = w_ft.fqn.c_name
+        c_ret = self.ctx.w2c(w_ft.w_restype)
+        if w_ft.params:
+            c_params = ", ".join(str(self.ctx.w2c(p.w_T)) for p in w_ft.params)
+        else:
+            c_params = "void"
+        self.tbh_fwdecl.wl(f"typedef {c_ret} (*{c_name})({c_params});")
 
     def emit_PtrType(self, fqn: FQN, w_ptrtype: W_PtrType) -> None:
         c_ptrtype = C_Type(w_ptrtype.fqn.c_name)
--- a/spy/tests/compiler/out_of_tree/mymod/__init__.py
+++ b/spy/tests/compiler/out_of_tree/mymod/__init__.py
@@ -1,7 +1,10 @@
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Annotated
 
 from spy.build.build_info import BuildInfo, BuildTarget, BuildType
+from spy.vm.b import B
+from spy.vm.function import FuncParam, W_Func, W_FuncType
+from spy.vm.primitive import W_I32
 from spy.vm.registry import ModuleRegistry
 from spy.vm.str import W_Str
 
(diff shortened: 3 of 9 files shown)


JeffersGlass:
`@functype` essentially allows up to not have to write a body for the function, yeah? I've always just used `type(CB)` in place or assigned to a variable:

```py
def CB(x: i32) -> i32:
    return 0

@blue
def make_adder(n: i32) -> type(CB):
    def adder(x: i32) -> i32:
        return x + n
    return adder
```

seibert:
Talking with @antocuni, it is likely that `@functype` works on accident right now (the `pass` in the body of CB is only working because it is a blue function and the signature isn't being checked), so we might want a different approach.

spy/backend/c/cbackend.py
```diff
@@ -274,19 +274,31 @@ def get_type_deps(self, fqn: FQN) -> list[FQN]:
...
+            # by name, so any by-value struct types must be defined first.
+            deps: list[FQN] = []
+            seen: set[FQN] = set()
+            for w_dep in [w_type.w_restype] + [p.w_T for p in w_type.params]:
+                if isinstance(w_dep, W_StructType):
+                    d = w_dep.fqn
+                    if d != fqn and d not in seen:
+                        seen.add(d)
+                        deps.append(d)
+            return deps
```
antocuni, line 288: I think this deserves its own unit test.
See e.g.:
https://github.com/spylang/spy/blob/0f2192b9ebb9b767b5d88dc5ab03c31b6a60142a/spy/tests/test_backend_c.py#L111-L140

spy/backend/c/cstructwriter.py
```diff
@@ -155,6 +158,15 @@ def emit_StructType(self, fqn: FQN, w_st: W_StructType) -> None:
...
         tb.wl("")
 
+    def emit_FuncType(self, fqn: FQN, w_ft: W_FuncType) -> None:
+        c_name = w_ft.fqn.c_name
+        c_ret = self.ctx.w2c(w_ft.w_restype)
+        if w_ft.params:
+            c_params = ", ".join(str(self.ctx.w2c(p.w_T)) for p in w_ft.params)
+        else:
+            c_params = "void"
+        self.tbh_fwdecl.wl(f"typedef {c_ret} (*{c_name})({c_params});")
```
antocuni, line 168: this works and it's good enough for this PR.
But note that it produces an incredibly ugly C type name.

This happens because function types get an FQN which depends on their "structural" shape. For example for a functype `def(i32, i32) -> f64` the fqn is:
```
FQN('builtins::def[i32, i32, f64]')
```

And the C typedef is:
```c
typedef double (*spy_builtins$def__builtins$i32_builtins$i32_builtins$f64)(int32_t, int32_t);
```

It's unclear how to improve it though (apart making `builtins` shorter, which maybe it's a good idea). The nice property of this is that if you compile two spy programs separately (e.g. a main executable and a lib) their function types are automatically compatible at the C level.

Random ideas for how to improve it:
1. give a unique numering and call them `FuncType1`, `FuncType2`, etc. They are more opaque, but complex function types will have unreadable names anyway. The problem is that separate compilation becomes harder because you need to agree on numbering.
2. do it at the spy level. Declare e.g. that `@functype` defines a NEW type with its own fqn:
```python
@functype
def BinOp(x: int, y: int) -> int:
    pass
```
Here, we would define a _fresh_ functype whose fqn is `mod::BinOp`: the advantage is that the redshift output and the C output become very clear. The new type would have conversion functions to convert to and from `def(int, int) -> int`.

I kind of like option 2, but my only doubt is that most PLs do NOT treat function types as new types. The only exception seems to be go, from what I understand.

spy/tests/compiler/out_of_tree/mymod/__init__.py
```diff
@@ -12,6 +15,10 @@
 
 MODULE = ModuleRegistry("mymod")
 
+# red def(i32) -> i32: a callback that takes one i32 and returns i32
+_w_cb_type = W_FuncType.new([FuncParam(B.w_i32, "simple")], B.w_i32)
+CB = Annotated[W_Func, _w_cb_type]
+
```
antocuni, line 21: this seems the wrong place where to put this test code, as it's not about out-of-tree at all.
I'd put it in `vm/modules/_testing_helpers.py`

spy/tests/compiler/out_of_tree/test_out_of_tree.py
```diff
@@ -37,3 +37,43 @@ def foo() -> str:
             return get_name()
         """)
         assert mod.foo() == "hello from mymod"
+
+    def test_c_callback(self):
```
antocuni, line 41: and these tests should go to `test_functype.py`

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+        @functype
+        def CB(a: i32, b: i32) -> i32:
+            pass
+        """)
+        w_CB = mod.w_mod.getattr("CB")
+        assert isinstance(w_CB, W_FuncType)
+        assert w_CB.w_restype is B.w_i32
+        assert [p.w_T for p in w_CB.params] == [B.w_i32, B.w_i32]
+        assert w_CB.color == "red"
+        assert w_CB.kind == "plain"
```
antocuni, line 21: if you want to simplify this test, you can probably just assert `w_CB.name == 'def (i32, i32) -> i32'`.

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+        mod = self.compile("""
+        @functype
+        def CB(x: i32) -> i32:
+            pass
+        """)
+        w_CB = mod.w_mod.getattr("CB")
+        w_T_direct = W_FuncType.new(
+            [FuncParam(B.w_i32, "simple")], B.w_i32, color="red", kind="plain"
+        )
+        assert w_CB is w_T_direct
```
antocuni, line 39: this seems a VERY defensive test, claude style :).
I'd just kill it as functypes are functypes are cached by construction and it's already tested by `tests/vm/test_function.py`

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+
+        def apply(cb: CB, x: i32, y: i32) -> i32:
+            return x + y
+
+        def my_add(a: i32, b: i32) -> i32:
+            return a + b
+
+        def run() -> i32:
+            return apply(my_add, 3, 4)
+        """)
```
antocuni, line 55: I don't really understand what this test is supposed to test.
_Probably_ it tests that you can use `CB` as a function argument type, but I'm not sure it warrants its own test.
I'd just put the end-to-end tests which currently live in `test_out_of_tree` and be happy.

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+        assert mod.run() == 7
```
antocuni, line 72: and this seems to be the exact same test as above?

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+            pass
+
+        def my_fn(x: i32) -> i32:
+            return x
+
+        def bad() -> CB:
+            return my_fn
+        """
+        errors = expect_errors("mismatched types")
+        self.compile_raises(src, "bad", errors)
```
antocuni, line 117: these have nothing to do with func types.
Typechecking function calls is already tested elsewhere.

Kill.

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+
+        @blue
+        def my_fn(x: i32) -> i32:
+            return x
+
+        def get_cb() -> CB:
+            return my_fn
+        """
+        errors = expect_errors("mismatched types")
+        self.compile_raises(src, "get_cb", errors)
```
antocuni, line 135: this is an interesting case, we can keep it.
But `expect_errors` should be MORE precise and check that we get a reasonable explanation/error annotation.

I have noticed multiple times that claude tends to be very bad at writing these tests.

spy/tests/compiler/test_functype.py
```diff
@@ -0,0 +1,154 @@
...
+
+        @blue
+        def make_adder(n: i32) -> CB:
+            def adder(x: i32) -> i32:
+                return x + n
+            return adder
+
+        add5: CB = make_adder(5)
+        add10: CB = make_adder(10)
+        """)
```
antocuni, line 154: kill, we already have an end-to-end test.

spy/vm/vm.py
```diff
@@ -514,10 +514,9 @@ def make_fqn_const(self, w_val: W_Object) -> FQN:
             assert w_val.fqn not in self.globals_w
 
         elif isinstance(w_val, W_Type):
-            # for now types are only builtin so they must have an unique fqn,
-            # we might need to change this when we introduce custom types
             fqn = w_val.fqn
-            assert w_val.fqn not in self.globals_w
+            if fqn in self.globals_w:
+                return fqn
```
antocuni, line 519: this is VERY wrong and it's a code smell. Bad claude 😡.
Early in this function we do a reverse lookup:
```python
        fqn = self.reverse_lookup_global(w_val)
        if fqn is not None:
            return fqn
```

so, if `w_val` is ALREADY in the globals, we returned the FQN.
If we reach this stage, it means that we need to add `w_val` to the globals, and the assert is a sanity check to ensure that for some reason the FQN doesn't already exist.
It turns out that the FQN `builtins::functype` is already taken by `W_FuncType` (see 
`function.py`):
```
W_FuncType._w = W_Type.declare(FQN("builtins::functype"))
```

So the failing assert here was catching a real bug. This means that we should probably rename the `@functype` decorator.

antocuni (changes requested):
it's mostly good, but with a lot of nitpicks here and there that overall degrade code quality and should be fixed.

The biggest thing to solve is the change to `SPyVM.make_fqn_const` which is plainly wrong, see the inline comment.

antocuni:
> `@functype` essentially allows up to not have to write a body for the function, yeah? I've always just used `type(CB)` in place or assigned to a variable:

@JeffersGlass `type(CB)` works but I think it would be good to have a way to give explicit names to function types, that's why I suggested the `@functype` decorator.

---

> Talking with @antocuni, it is likely that `@functype` works on accident right now (the `pass` in the body of CB is only working because it is a blue function and the signature isn't being checked), so we might want a different approach.

Yes, I think that now it works by chance, but we could make it working with `...`:

```python
@functype
def CB(x: int, y: int) -> float: ...
```

Speaking of which, I'm not 100% sure about the name `@functype`. What about:

```python
@def_functype
def CB(x: i32, y: i32) -> i32: ...
```

Other possibilities:
- `@define_functype`
- `@functype_alias`
- `@make_functype`
- `@get_functype`
- `@new_functype`

JeffersGlass:
My personal favorites among the names you listed @antocuni are `def_functype` and `make_functype`, but that's only based on vibes :smile:

PR #610, opened 2026-06-29
SPY_GC_NONE and spy_nogc_alloc

paugier:
Related to https://github.com/spylang/spy/pull/594

I tried to implement the improvement discussed in https://github.com/spylang/spy/pull/594/files#r3448681112

### DIFF
--- a/spy/build/flags.py
+++ b/spy/build/flags.py
@@ -26,6 +26,7 @@ BuildType = str  # "release" | "debug"
 _BASE_CFLAGS: list[str] = [
     "-fvisibility=hidden",
     "-fPIC",
+    "-DSPY_GC_NONE",
 ]
 
 # WASM-specific CFLAGS shared by wasi and emscripten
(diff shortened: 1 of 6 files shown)


spy/build/flags.py
```diff
@@ -26,6 +26,7 @@
 _BASE_CFLAGS: list[str] = [
     "-fvisibility=hidden",
     "-fPIC",
+    "-DSPY_GC_NONE",
```
antocuni, line 29: is it necessary to have it here by default and then filter it out in `config.py`?
Can't we remove it from here and add it inside `config.py` in a `if config.gc == "none"` branch?

antocuni (approved):
LGTM, but see one comment.

spy/build/flags.py
paugier: I think the libspy Makefile is not affected by config.py so `"-DSPY_GC_NONE"` needs to be in flags.py.

I thought that this "filter out" was the simplest solution.

PR #611, opened 2026-06-29
pyproject.toml: [dependency-groups] and uv

paugier:


paugier:
Good, prefix-dev/setup-pixi checks that pixi.lock is up-to-date:

```
Run prefix-dev/setup-pixi@v0.9.4
Downloading Pixi
pixi info
Restoring project cache
pixi install --locked
Error: The process '/home/runner/.pixi/bin/pixi' failed with exit code 1
```

paugier:
CI is green with uv and Pixi (both check that their respective .lock file is up-to-date).

I need to update at least:

- [x] README.md
- [x] CONTRIBUTING.md
- [x] docs/src/contributing.md

Note for me: `pip install -e . --group dev` seem to work.

paugier:
I think this PR is ready.

uv and Pixi are both supported as first class workflows.

pip continues to be usable but not with `pip install -e .[dev]`. One needs to run with a recent pip: `pip install -e . --group dev`.

pixi.lock and uv.lock are both versioned. There are mechanisms in the CI to check if these files are up-to-date with pyproject.toml and approximately in sync (modified together in the same PR). It seems to me that it's enough.

.github/workflows/pixi-checks.yml
```diff
@@ -23,6 +23,24 @@ jobs:
...
+        run: |
+          BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          UV_CHANGED=$(git diff --quiet $BASE_SHA -- uv.lock && echo no || echo yes)
+          PIXI_CHANGED=$(git diff --quiet $BASE_SHA -- pixi.lock && echo no || echo yes)
+          echo "uv.lock changed in this PR:   $UV_CHANGED"
+          echo "pixi.lock changed in this PR: $PIXI_CHANGED"
+          if [ "$UV_CHANGED" != "$PIXI_CHANGED" ]; then
+            echo "::error::Only one of uv.lock/pixi.lock was modified in this PR. Run both 'pixi lock' and 'uv lock' and commit any resulting changes."
+            exit 1
+          fi
```
antocuni, line 43: this is good enough for this PR.
Idea for a follow-up (only if it's easy to do, there is no point in wasting too much time on it else): we could put this logic into a script, and have a pre-commit hook to check it?

.github/workflows/playground-deploy.yml
```diff
@@ -23,38 +23,33 @@ jobs:
...
+        uses: astral-sh/setup-uv@v8.1.0
         with:
-          python-version: "3.13"
+          enable-cache: true
 
-      - name: Install ziglang
-        run: python -m pip install ziglang==0.13.0
+      - name: Install venv
+        run: |
+          # to be able to call `python -m spy.build.flags`
```
antocuni:
```suggestion
          # we need to have spy installed BEFORE the "build libspy" pass because the Makefile calls `python -m spy.build.flags`
```

.github/workflows/tests.yml
```diff
@@ -18,7 +18,7 @@ on:
 
 jobs:
   test:
-    name: test-${{ matrix.os }}-${{ matrix.python-version}}
+    name: test-${{ matrix.os }}
```
antocuni, line 21: I think we no longer do `matrix.python-version` because now the python version is managed by uv in `pyproject.toml`, is it correct?
What will we need to do in case we decide to run CI agains multiple supported python version?

docs/src/contributing.md
```diff
@@ -19,28 +19,29 @@ Next, clone the repo to your machine:
 
 Create a virtual environment for your project, and install both the SPy dependencies and the docs-specific dependencies. Any package installer and virtual environment will work; here are a couple of options:
```
antocuni:
```suggestion
Create a virtual environment for your project, and install both the SPy dependencies and the docs-specific dependencies. Any package installer and virtual environment will work, but `uv` is preferred as it's the one tested in CI. Here are a couple of options:
```

antocuni: @paugier what about this wording?

antocuni (approved):
LGTM.
I added a few comments but they are not merge blockers.

Feel free to merge!
Thanks

.github/workflows/tests.yml
paugier: - extend the range of supported Python versions in pyproject.toml.
- add `matrix.python-version` and use it as an input parameter for uv.

.github/workflows/pixi-checks.yml
paugier: I agree we could do that.

Very bad answer: I'm not very motivated to work on pre-commit since I don't use Git locally so pre-commit doesn't have effect on me :smiley: 

More generally, I don't love this principle of changes automatically included in a commit.

Moreover, I expect that this won't be so disturbing for the few developers that need to modify the dependencies. Most contributors won't be affected.

docs/src/contributing.md
paugier: I included that in the last commit.

PR #619, opened 2026-07-18
Fix path to `examples/1_high_level/hello.spy` in `README.md`

touilleMan:
`hello.spy` has been moved last month by commit cf0d731f60ef6e9902c8698663c8b7c8f86bddd2, but README hasn't been updated accordingly ;-)

### DIFF
--- a/examples/1_high_level/hello.spy
+++ b/examples/1_high_level/hello.spy
@@ -1,9 +1,1 @@
-"""
-Things to notice:
-  - Every SPy program defines `main()` as its entry point.
-    SPy libraries do not need a `main()` function.
-"""
-
-
-def main() -> None:
-    print("Hello world!")
+../hello.spy
--- a/examples/hello.spy
+++ b/examples/hello.spy
@@ -1,0 +1,9 @@
+"""
+Things to notice:
+  - Every SPy program defines `main()` as its entry point.
+    SPy libraries do not need a `main()` function.
+"""
+
+
+def main() -> None:
+    print("Hello world!")
--- a/spy/tests/test_cli.py
+++ b/spy/tests/test_cli.py
@@ -337,8 +337,8 @@ class TestMain:
     def test_execute_pyodide(self):
         # pyodide under node cannot access /tmp/, so we cannot try to execute
         # files which we wrote to self.tmpdir. Instead, let's try to execute
-        # examples/1_high_level/hello.spy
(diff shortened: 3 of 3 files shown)


paugier:
Don't you think that a symlink in `examples` would be better for this? This would keep the README a bit simpler.

antocuni:
@paugier I like the symlink idea.
@touilleMan do you want to do it?

touilleMan:
I can updated the PR to use a simlink
However this might create an issue on Windows (by default, symlinks become regular text files containing their target paths on Windows) are you okay with this ?

paugier:
On Windows, the symlink will be cloned as a copy, which will be fine to run the commands in the README, so yes it is fine.

antocuni:
> On Windows, the symlink will be cloned as a copy, which will be fine to run the commands in the README, so yes it is fine.

I think that on windows you get a file whose _content_ is the path to the symlinked file (at least with some git config). This would break the readme.
But what we can do is the following:
1. the "real" file is in examples/hello.spy
2. examples/1_high_level/hello.spy is a symlink to ../hello.spy

This should always work because the playground is built on linux.

paugier:
Yes, the symlink in examples/1_high_level looks like a good solution.

and sorry for my bad understanding...

touilleMan:
@antocuni I have updated the PR so that `examples/hello.spy` is the real file and `examples/1_high_level/hello.spy` the symlink :+1:

antocuni:
@touilleMan thanks! Tests are failing though :(
It should be as easy as fixing the path here:
https://github.com/spylang/spy/blob/86d1647813709314d409ffae17ce4889de1ed087/spy/tests/test_cli.py#L335-L344

I suspect that for some reason inside pyodide we are unable to resolve the symlink correctly.

touilleMan:
@antocuni I have updated the PR with your suggestion, it seems to work fine :+1: 

(the macOS CI seems not have some hiccup though, but it is related to Github Action, not me :sweat_smile: )

antocuni:
> @antocuni I have updated the PR with your suggestion, it seems to work fine 👍
> 
> (the macOS CI seems not have some hiccup though, but it is related to Github Action, not me 😅 )

yes, I tried to rerun that job but it seems that github actions are having issues. Given that it works on ubuntu and the change is easy, I'll be brave and merge anyway :).

Thanks for doing that!

PR #620, opened 2026-07-18
Document in README that pip install requires PEP 735 Dependency Groups support

touilleMan:


paugier:
Thanks @touilleMan. I'm responsible for this issue so happy about this fix!

PR #623, opened 2026-07-18
Add repr support for floating-point types

filiplajszczak:
f32 and f64 previously inherited the identity-based object repr. That lacked value in the interpreter and generated calls to missing symbols in C builds. Both types got value-based repr implementations matching their str output. f32 got the missing C formatter.


spy/tests/compiler/test_str.py
```diff
@@ -320,6 +324,20 @@ def repr_u32(x: u32) -> str:
...
+        """)
+        assert mod.repr_f64(3.14) == "3.14"
+
+    def test_repr_f32(self):
+        mod = self.compile("""
+        def repr_f32(x: f32) -> str:
+            return repr(x)
+        """)
+        assert mod.repr_f32(3.14) == "3.14"
+
```
antocuni: this is a bit of nitpick but it might be a better idea to have a `test_repr_numbers` (similar to `test_str_number`). This way we avoid proliferation of tests which all look the same, and it will also be faster as we avoid compiling lots of small modules.

antocuni (approved):
LGTM

spy/tests/compiler/test_str.py
filiplajszczak: Makes sense.

filiplajszczak:
Updated the implementation to use vendored Ryu for shortest-round-trip f32 formatting, then normalize its output to the intended CPython-style representation. The same C formatter is now shared by the interpreter and generated C backend.

spy/tests/compiler/test_str.py
```diff
@@ -315,12 +319,83 @@ def repr_u8(x: u8) -> str:
...
+        assert mod.repr_f32(1e-5) == "1e-05"
+        # CPython switches to fixed notation at exponent -4.
+        assert mod.repr_f32(1e-4) == "0.0001"
+        # CPython keeps exponent 15 in fixed notation.
+        assert mod.repr_f32(1e15) == "1000000000000000.0"
+        # CPython switches to scientific notation at exponent 16.
+        assert mod.repr_f32(1e16) == "1e+16"
+        # str and repr share one formatter, including for values needing more digits.
+        x = 0.5000000596046448
+        assert mod.str_f32(x) == mod.repr_f32(x)
```
antocuni: these tests are excellent, but since they are checking all the possible cases of f32 formatting, maybe they belong more to `test_float`?
We could keep a simple smoke test for f32 here in test_str, and move all the more specific tests to test_float.

spy/tests/compiler/test_str.py
```diff
@@ -315,12 +319,83 @@ def repr_u8(x: u8) -> str:
...
+
+    def test_repr_f32_roundtrip(self):
```
antocuni: same as above, I think it belongs to test_float

spy/libspy/src/str.c
```diff
@@ -108,6 +110,115 @@ spy_builtins$f64$__str__(double x) {
     return spy_str_from_format("%g", x);
 }
 
+/* Ryu finds the shortest digits but always renders them in scientific notation.
+   Reformat those digits to match CPython's repr conventions without changing the
+   round-trippable value. */
```
antocuni: I'm confused. Ryu readme says this:
```
IEEE Type	Supported Output Formats
[...]
32 Bit (float)	Shortest
```

And then:
```
The %f format prints the full decimal part of the given floating point number, and then appends as many digits of the fractional part as specified using the precision parameter.

The %e format prints the decimal number in scientific notation with as many digits after the initial digit as specified using the precision parameter.

The %g format prints either %f or %e format, whichever is shorter.
```

So I seem to understand that "shortest" means that the implement "%g", and thus sometimes it might actually return something in non-scientific notation?

Who is wrong here? Your assumption, the README or my understanding? :)

spy/libspy/Makefile
```diff
@@ -10,7 +10,10 @@
 #
 EMCC_AVAILABLE := $(shell which emcc 2>/dev/null)
 
-SRCS = src/str.c src/bytes.c src/builtins.c src/debug.c src/unsafe.c src/operator.c src/posix.c
+SRCS = src/str.c src/bytes.c src/builtins.c src/debug.c src/unsafe.c src/operator.c src/posix.c \
+       vendored/ryu/ryu/f2s.c
+
+RYU_CFLAGS := -I $(CURDIR)/vendored/ryu -DRYU_FLOAT_FULL_TABLE
```
antocuni, line 17: what does `-DRYU_FLOAT_FULL_TABLE` do?
I couldn't find any docs for it, maybe it's worth putting a comment to explain it.

antocuni (commented):
Thank you @filiplajszczak.
I think the PR is almost ready to be merged.
See below for a nitpick suggestion about tests (not a merge blocker, feel free to ignore), plus a couple of comments/questions about ryu

spy/tests/compiler/test_str.py
filiplajszczak: Makes sense. I'll apply that.

spy/tests/compiler/test_str.py
filiplajszczak: applied

spy/libspy/src/str.c
filiplajszczak: “Shortest” refers to the number of significant decimal digits needed to round-trip to the same `f32`, rather than to choosing the shorter of fixed and scientific notation as `%g` does.

In the `f2s_buffered_n` implementation we vendor, `to_chars()` unconditionally appends `E` and an exponent for every finite nonzero value. The normalization step keeps Ryu’s shortest round-trippable digits but changes their presentation to CPython’s fixed/scientific notation rules.

So the README’s description of `%g` is correct, but `f2s_buffered_n` is not a `%g` formatter. I’ll clarify the comment.

spy/libspy/Makefile
filiplajszczak: It makes Ryu use the dedicated f32 powers-of-five lookup tables from `f2s_full_table.h` (precomputed constants used to convert efficiently between binary and decimal exponents). Without it, the f32 implementation depends on either Ryu’s d2s tables or its size-optimized calculation path.

We only vendor the f32 subset here, so this option lets `f2s.c` remain independent of the larger d2s tables. This is documented in `vendored/ryu/README.md`, but I agree it is not obvious at the Makefile call site. I’ll add a short comment there.

antocuni: oh I see now. I didn't realize that we only vendor a _subset_ of ryu.
Thanks for the explanation!

filiplajszczak:
With extra/edited comments it could be fine now.

antocuni (approved):
thank you very much @filiplajszczak! 🙏 
LGTM, merging.

PR #624, opened 2026-07-18
docs: fix path to hello example in README.md

KuligKamil:
Examples in README.md didn't work.


paugier:
Thanks. It's useful but it seems to me that it is the same as https://github.com/spylang/spy/pull/619 :smiley:

PR #625, opened 2026-07-18
typed var declaration in loops is not allowed in interpreted mode

6502:
Original code was working by chance because the \xHH escape was exercised only once in the tests.
Removing the type makes it ok.


antocuni (approved):
LGTM.
I approved test run just now, will merge as soon as they are green.

PR #627, opened 2026-07-19
Add unicode_codepoint iterator

Seppli11:
This PR adds a the `unicode_codepoint` iterator in the `_str` module, allowing to iterate over the individual codepoints of a utf-8 string.

### DIFF
--- a/spy/tests/compiler/test_str.py
+++ b/spy/tests/compiler/test_str.py
@@ -1,6 +1,8 @@
 import re
 from textwrap import dedent
 
+import pytest
+
 from spy.errors import SPyError
 from spy.tests.support import CompilerTest, only_interp, skip_backends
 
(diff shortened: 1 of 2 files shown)


spy/tests/compiler/test_str.py
```diff
@@ -669,3 +671,61 @@ def foo(s: str, target: str) -> bool:
...
+        assert get_codepoint(input, 0) == 104
+        assert get_codepoint(input, 1) == 101
+        assert get_codepoint(input, 2) == -1
+
+    def test_unicode(self, get_codepoint):
+        input = "hé桁🨀"
+        # h (1 utf byte)
+        assert get_codepoint(input, 0) == 104
+        # é (2 utf bytes)
+        assert get_codepoint(input, 1) == 233
```
antocuni: maybe we can use something like `ord('é')` instead of hardcoding the number? (same for the other lines)

spy/tests/compiler/test_str.py
```diff
@@ -669,3 +671,61 @@ def foo(s: str, target: str) -> bool:
...
+            it = unicode_codepoints(StrObject.from_str(s))
+            j = 0
+            for codepoint in it:
+                if j == i:
+                    return codepoint
+                j += 1
+            return -1
+        """
+        mod = self.compile(src)
+        return mod.get_codepoint
```
antocuni: I'd suggest to:
1. don't create a separate `TestUnicodeCodePoints` class
2. add a single `test_unicode_codepoints` method to the existing class
3. keep a separate test_unicodede_codepoints_invalid`

Having separate "small" tests is very nice, but at the same time by doing that we force compiling the same code multiple times, so on the long run if we do this for all the tests we will get a very slow testsuite.

Also, I prefer to have a single-class-per-test-file (I don't think it's written anywhere, sorry for that :)).

antocuni (commented):
LGTM, but see some (small) comments

Seppli11:
Thx for the review.
I consolidated all the tests (except for `test_unicodede_codepoints_invalid`) into one test and used `ord("...")` in the assertions. I've re-requested a review

antocuni (approved):
LGTM, thanks!
I'll merge as soon as tests are green (currently running)

PR #628, opened 2026-07-19
Handle signed zero and non-finite f32 constants

filiplajszczak:
The interpreter's IEEE-754 division fallback selected infinity from the numerator sign alone, so division by negative zero returned the wrong sign. Combine both operand signs with copysign while retaining NaN for zero divided by zero.

Redshift can fold `f32` overflow into infinity. The C writer previously appended an `f` suffix to Python's spelling and emitted invalid identifiers such as `inff` and `nanf`; emit the C99 `INFINITY` and `NAN` macros instead.

The signed-zero regression constructs -0.0 inside SPy, so it tests signed-zero behavior independently of Python-to-SPy argument conversion. The constant-emission regression overflows the largest finite `f32` deliberately, ensuring the C writer receives a folded non-finite constant instead of a runtime division call.

It was possible thanks to the help of gpt-5.6-sol.

### DIFF
--- a/spy/backend/c/cwriter.py
+++ b/spy/backend/c/cwriter.py
@@ -1,3 +1,4 @@
+import math
 from types import NoneType
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


antocuni (approved):
LGTM. Will merge as soon as tests turn green

PR #630, opened 2026-07-19
Make examples work when compiled

redorlik:
Added a way to test if the examples work when compiled. Due to a problem of getting the correct return code with the "spy bull -x", the building process is separate from the test of the executable.

There are four examples that are not compilable at this point in time, and these examples are marked as XFAIL.

AI was not used in this pull request

### DIFF
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -54,6 +54,15 @@ def _run(spy_file: Path) -> subprocess.CompletedProcess:
         check=False,
     )
 
+def _build_run(spy_file: Path) -> subprocess.CompletedProcess:
+    cmd = ["spy","build","-x", str(spy_file)]
+    print( "cmd >>>>>>>"," ".join(cmd))
+    return subprocess.run(
+        cmd,
+        capture_output=True,
+        text=True,
+        check=False,
+    )
 
 def test_all_examples_have_expected_output(request) -> None:
     """Catch the case where a new .spy file was added without a saved expected output.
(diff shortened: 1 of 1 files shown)


examples/test_examples.py
```diff
@@ -107,3 +117,14 @@ def test_example(spy_file: Path, request) -> None:
...
+
+@pytest.mark.parametrize("spy_file", _spy_files, ids=lambda f: f.stem)
+def test_example_interp(spy_file: Path, request) -> None:
+    run_example(spy_file,request)
+
+@pytest.mark.parametrize("spy_file", _spy_files, ids=lambda f: f.stem)
+def test_example_build(spy_file: Path, request) -> None:
+    # There are four examples which fails in the c compiling stage - so they are marked as XFAIL
+    if spy_file.stem in ['collections','annotated','unroll_nested_loops','convert']:
+        pytest.xfail()
```
antocuni, line 129: I'm not sure about this approach.
The biggest issue is that it doesn't solve the problem that you encountered during the sprint, i.e. that you opened an example, tried to compile and see it was broken.

I think that a better way is the following:
1. add a special "marker" at the beginning of the docstring of examples which are known not to compile
2. detected the marker here (could be something as simple as `if MARKER in spy_file.read()`).

What to use as a marker is open to discussion, but I propose something like this in the first line of the docstring:
```
"""
SPY BUILD: xfail

Things to notice:
  - Tuples, dicts, and lists work as in Python.
[...]
"""

def main() -> None:
    ...
```

antocuni (commented):
LGTM but see the comment on how to detect xfails.

Also, I see that tests fail early due to `ruff check`. You might want to install/enable `pre-commit`, see `CONTRIBUTING.md`.

PR #631, opened 2026-07-22
Merge Assign and UnpackAssign ast classes.

rlamy:
Simple and unpacking assignments are now distinguished by the type of their target.

Fixes buggy scope analysis when assigning to `a, b` when `a` exists in an outer scope.

For context, this was discussed with @antocuni during the EuroPython sprint as a prerequisite to #455. The next step would be to desugar complex assignments into a sequence of simple ones, but this refactoring appears to be already useful on its own.


antocuni (approved):
LGTM, thank you for doing this! Do you plan to continue working on #455?

PR #639, opened 2026-07-30
Wip

hoodmane:


antocuni:
for more context and to the benefit of whoever stumbles upon here.
This is a PR which I and @hoodmane started when physically together, then hood continued on his own. It aims to make pyodide a "first class" supported platform, which means that we will be more confident about the playground.

I'll review this PR ASAP (might take some days)

hoodmane:
Okay CI is green so I'll close this, clean up the branch, and make a real PR.

PR #642, opened 2026-08-05
Fix HostModule on Emscripten

hoodmane:


spy/libspy/include/spy.h
```diff
@@ -26,14 +35,17 @@
...
+#if defined(SPY_TARGET_EMSCRIPTEN)
+#  include "emscripten.h"
 
-#if defined(SPY_GC_NONE) + defined(SPY_GC_BDWGC) != 1
-#  error "You must define either SPY_GC_NONE or SPY_GC_BDWGC"
+#  define IMPORT_STUB_TRAPPING(ret, name, rest...)                                     \
+      EM_JS(ret, name, rest, { _spy_panic_import_stub_js(#name); } name.stub = true)
+
+#else
+#  define IMPORT_STUB_TRAPPING(ret, name, rest...) ret WASM_IMPORT(name) rest
```
antocuni: I'd like a comment to explain what's going on here, because I'm not sure to fully understand

spy/libspy/src/emscripten_pre.js
```diff
@@ -0,0 +1,9 @@
+Module.instantiateWasm = (imports, successCallback) => {
```
antocuni, line 21: a comment here would be appreciated as well.

antocuni (commented):
I'm not sure to understand what's going on.
I was about to ask an explanation in the PR, but I think it's probably a better idea to add comment in the source code for posterity as well.

hoodmane:
Yeah, makes sense this is a bit obscure. I added detailed comments.

spy/libspy/src/emscripten_pre.js
```diff
@@ -0,0 +1,24 @@
+// Custom instantiateWasm hook that also calls Module.adjustWasmImports(imports)
+// if it is defined. This allows us to insert HostModule-defined symbols into
+// imports.
+//
+// instantiateWasm is the only hook that Emscripten exposes after the imports
+// object is created but before wasm instantiation. My attempts to add an
+// adjustWasmImports hook upstream have not succeeded:
+// https://github.com/emscripten-core/emscripten/pull/23794
```
antocuni, line 8:
```suggestion
// instantiateWasm is the only hook that Emscripten exposes after the imports
// object is created but before wasm instantiation. A previous attempt to add an
// adjustWasmImports hook upstream has not succeeded:
// https://github.com/emscripten-core/emscripten/pull/23794
```

spy/libspy/include/spy.h
```diff
@@ -26,14 +35,32 @@
...
+// defined in a host module. If it's ever called at runtime, we panic.
+//
+// TODO: Maybe drop the panic in release mode?
 
-#if defined(SPY_GC_NONE) + defined(SPY_GC_BDWGC) != 1
-#  error "You must define either SPY_GC_NONE or SPY_GC_BDWGC"
+#  include "emscripten.h"
+
+#  define IMPORT_STUB_TRAPPING(ret, name, rest...)                                     \
+      EM_JS(ret, name, rest, { _spy_panic_import_stub_js(#name); } name.stub = true)
```
antocuni: so this is calling the JS function `_spy_panic_import_stub_js`, right?
But I don't understand, where is it defined? I cannot find it anywhere

spy/libspy/src/emscripten_pre.js
```diff
@@ -0,0 +1,24 @@
+// Custom instantiateWasm hook that also calls Module.adjustWasmImports(imports)
+// if it is defined. This allows us to insert HostModule-defined symbols into
+// imports.
```
antocuni, line 3: aaaah, it took a while to understand but now I think I get it.
We need to do this dance so that `adjustWasmImports` is called.
We **already** define `adjustWasmImports` elsewhere (in `llwasm/emscripten.py:link_and_instantiate`), that's why we don't see any definition in ths PR.

This hook is already used by the playground, but for that we rely on a pyodide-specific feature which calls `adjustWasmImports`: this is a more general solution which works even when we compile non-pyodide standalone modules.

Is it correct?

spy/libspy/include/spy.h
antocuni: I think that `WASM_IMPORT` without `IMPORT_STUB_TRAPPING` doesn't make sense when targeting emscripten, right?

If so, I wonder whether this should just be called `WASM_IMPORT`. The idea is that:
- on wasi targets, `WASM_IMPORT` is just an extern declaration with a special `__attribute__` recognized by the linker
- on emscripten targets, `WASM_IMPORT` defines a stub which is later overridden by `adjustImports`.

This probably means that we need to change the signature of `WASM_IMPORT` to have return type, name and rest separated, but I think it's good enough.
What do you think?

antocuni (commented):
thank you for the comments! I have a much better understanding now, but this raised _more_ questions, see below :)

spy/libspy/src/emscripten_pre.js
hoodmane: Well we were defining it before but it was dead code. There was at some point a patch to Emscripten that made this work but it got deleted at some point, probably when you put the xfail into the HostModule test.

spy/libspy/include/spy.h
hoodmane: > so this is calling the JS function _spy_panic_import_stub_js, right?
> But I don't understand, where is it defined? I cannot find it anywhere

Oops, it was defined in my wip branch but I forgot to move it here.

hoodmane: > I think that WASM_IMPORT without IMPORT_STUB_TRAPPING doesn't make sense when targeting emscripten, right?

Right. On my WIP branch, I had `IMPORT_STUB_TRAPPING()` and `IMPORT_STUB_FALLBACK()`. The `FALLBACK()` version doesn't panic because it is expected that the callee can handle the situation where it is not provided. For instance by either calling an imported panic function for tests or by debug printing some stuff and aborting.

If we just want to call it `WASM_IMPORT` I'm not sure what the best way to handle the fallback case is.

antocuni: > The `FALLBACK()` version doesn't panic because it is expected that the callee can handle the situation where it is not provided. For instance by either calling an imported panic function for tests or by debug printing some stuff and aborting.

is this feature actually needed by anything in SPy?
If not, I'd probably prefer to just drop the support for the fallback case and rename `IMPORT_STUB_TRAPPING` to `WASM_IMPORT`.

hoodmane: Well it's needed by the panic handler and debug printer -- there's one version of these things for the tests and a different version that runs otherwise. But I can make something ad hoc.

hoodmane: Let's do as you suggest for now and we can deal with anything else in a followup.

antocuni: > Well it's needed by the panic handler and debug printer -- there's one version of these things for the tests and a different version that runs otherwise. But I can make something ad hoc.

I *think* I see what you mean, but I think it's broken also for WASI builds and should probably be handled differently to what we do now, but in the same way for WASI and emscripten builds.

Currently, WASI standalone builds are broken:

```
❯ cat /tmp/x.spy 
def foo(x: float, y: float) -> float:
    return x / y

def main() -> None:
    print(foo(1, 1))

❯ spy build --target wasi /tmp/x.spy 
[debug] /tmp/build/x.wasm 

❯ wasmtime /tmp/build/x.wasm 
Error: failed to run main module `/tmp/build/x.wasm`

Caused by:
    0: failed to instantiate "/tmp/build/x.wasm"
    1: unknown import: `env::spy_debug_log` has not been defined
```

We even have this comment in `debug.h`:
```
/* Debug utilities:
  - In emscripten and native mode, these are implemented in debug.c
  - In WASI mode, they must be provided by the host.

TODO: ideally, we want TWO different WASI modes:
  - for tests, we want the imports
  - for standalone executables, we want debug.c
*/
```

I think we never caught this in tests because in tests we compile/run `hello.spy` for WASI: the problem arises only if the generated C code calls (directly or indirectly) `spy_panic` (and float division does).

So *probably* what we want to fix that TODO. I started to look into it but it seems a bit more nuanced than I thought.
I'll try to fix it soon and then we can continue from there?

antocuni: oh, I see you committed something in the meantime.
I'm also happy to merge this ASAP and fix the WASI thing later.

hoodmane: Well I can fix this if you like. But probably in a followup to this PR?

hoodmane: So the trick about changing `WASM_IMPORT` like I tried is that now it defines the symbol. I think this is why I originally made a separate macro for `IMPORT_STUB_TRAPPING`.

antocuni: > Well I can fix this if you like. But probably in a followup to this PR?

if you feel like doing that, would be very appreciated :).
I looked a bit at the code, I think that what I want is:
1. modify `build.config.OutputKind`: currently it's `Literal["exe", "lib", "py-cffi"]`, but "lib" is essentially "what we need for tests" and it's not used elsewhere. So I'd rename `lib` to `testlib`.
2. There is already a check that `outputkind == 'lib'` works only for emscripten and wasi. Currently the check in in `ninja.py` (likely for historical reasons) but probably it should be moved to `config.py`
3. add a `SPY_OUTPUT_KIND` macro visible to C; 
4. fix the TODO in `debug.h` by dispatching on `SPY_OUTPUT_KIND`.

antocuni:
thank you @hoodmane ! Merged

PR #643, opened 2026-08-05
Run `prek run -a` on main branch

hoodmane:
For some reason my local precommit doesn't agree with the results in CI. Maybe CI precommit will accept the results of my local run?


antocuni (approved):
I think this happens because we added formatting for .spy and .[ch] files later, and then we have been running it only on files which were modified since then.

+1 for this

PR #644, opened 2026-08-08
Rename lib OutputKind to testlib

hoodmane:
Since it's only used for tests. I also moved the definition of the OutputKind type into build_info.py to go next to BuildType and BuildTarget.


spy/build/flags.py
```diff
@@ -15,12 +16,12 @@
 from typing import Optional
 
 import spy
+from spy.build.build_info import BuildType
 
 _LIBSPY = spy.ROOT.join("libspy")
 _INCLUDE = _LIBSPY.join("include")
 _BUILD = _LIBSPY.join("build")
 
-BuildType = str  # "release" | "debug"
```
antocuni, line 23: nice catch :)

antocuni (approved):

antocuni:
thanks @hoodmane !

PR #650, opened 2026-08-10
For Emscripten and WASI build both libspytest.a and libspy.a

hoodmane:
libspy.a defines the debug helpers in debug.c whereas libspytest.a imports them from wasm imports. libspytest.a is designed for using with llwasm, maybe we should have a different name that reflects this more accurately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,10 @@ ways:
     statically linked to any spy executable
 
   - `make -C spy/libspy` creates a `libspy.a` for each supported target, which
-    currently are `native`, `emscripten` and `wasi`
+    currently are `native`, `emscripten` and `wasi`. For `emscripten` and `wasi`
+    it also creates a `libspytest.a`, which is the flavor used by testlibs.
+    `libspytest.a` expects the WebAssembly host to provide debug helpers as WASM
+    imports whereas `libspy.a` implements them in `debug.c`.
 
   - `spy/libspy/__init__.py` contains some support code to be able to load the
     WASM version of libspy in the interpreter.
--- a/README.md
+++ b/README.md
@@ -258,9 +258,9 @@ ways:
 
   - `make -C spy/libspy` creates a `libspy.a` for each supported target, which
     currently are `native`, `emscripten` and `wasi`. For `emscripten` and `wasi`
-    it also creates a `libspytest.a`, which is the flavor used by testlibs.
-    `libspytest.a` expects the WebAssembly host to provide debug helpers as WASM
-    imports whereas `libspy.a` implements them in `debug.c`.
+    it also creates a second `libspy.a` used by llwasm which expects the
+    WebAssembly host to provide debug helpers as WASM imports. The normal
+    `libspy.a` implements them in `debug.c`.
 
   - `spy/libspy/__init__.py` contains some support code to be able to load the
     WASM version of libspy in the interpreter.
--- a/spy/build/config.py
+++ b/spy/build/config.py
@@ -6,7 +6,7 @@ from typing import Literal, Optional
 
 import spy.libspy
 from spy.build.build_info import BuildTarget, BuildType, OutputKind
-from spy.build.flags import get_cc, get_cflags, get_ldflags, get_libdir
+from spy.build.flags import get_cc, get_cflags, get_ldflags, get_libdir, get_libname
 from spy.errors import WIP
 
 GCOption = Literal["none", "bdwgc"]
@@ -59,7 +59,7 @@ class CompilerConfig:
 
         self.CC = get_cc(flags_target)
         self.cflags += get_cflags(
-            flags_target, config.build_type, config.warning_as_error
+            flags_target, config.build_type, config.kind, config.warning_as_error
         )
         self.cflags += EXTRA_CFLAGS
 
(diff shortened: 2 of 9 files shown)


antocuni:
I agree it's a problem that we need to solve somehow.
I don't dislike the libspy/libspytest solution, but let's try to explore more opportunities, because I have some vague feeling that we might be adding the split on the wrong axis of the matrix.

Currently, this is the matrix:
- TARGET: native, wasi, emscripten
- TYPE: release, debug
- OUTPUT: exe, testlib, py-cffi

Here we are adding a new axis; let's call it:
- TEST: test, notest

Let's ignore py-cffi for a while, as it is a bit special anyway and might disappear in the future.

The main combinations which are allowed are:
- standalone executables (what you get from `spy build`): `TARGET/TYPE/exe/notest`
- snippets for test (what you get from `self.compile(src)`): `TARGET/debug/testlib/test`

Some notes:
  - `TARGET/release/testlib/test` is not used/supported now but it could be in the future
  - `native/debug/testlib/test` is not supported not but it will be if/when I do my `linuxtime` experiment

In this PR, you project this matrix on the filesystem in this way (emscripten not shown because I didn't have emcc available on this machine). Let's call is the **libspytest option**:
```
❯ fd -I libspy
native/debug/libspy.a
native/release/libspy.a
wasi/debug/libspy.a
wasi/debug/libspy.wasm
wasi/debug/libspytest.a
wasi/release/libspy.a
wasi/release/libspy.wasm
wasi/release/libspytest.a
```

I think we have at least two different ways to project this.
**Nested outputkind option**: `TARGET/TYPE/OUTPUTKIND/libspy.*`
So for example we would have:
```
native/debug/testlib/libspy.a
native/debug/exe/libspy.a
native/release/testlib/libspy.a
native/release/exe/libspy.a
...
wasi/debug/testlib/libspy.a    # what tests need
wasi/debug/exe/libspy.a        # for standalone wasi exes
wasi/release/testlib/libspy.a  # not really used as of now
wasi/release/exe/libspy.a      # for standalone wasi exes
emscripten/...
```

**flattened OUTPUTKIND option**: here we conflate TYPE and TEST into a single variable; e.g `release, debug, debug-testlib`. So we would have:
```
native/debug/libspy.a
native/debug-testlib/libspy.a
native/release/libspy.a
wasi/debug/libspy.a.          # for standalone wasi exes
wasi/debug-testlib/libspy.a   # what tests need
wasi/release/libspy.a         # for standalone wasi exes
emscripten/...
```

Some pros and cons:
- `libspytest` PRO is that it makes it clear that it's "the libspy for tests", especially if we find a better name like `libspy_for_tests.a`
- the biggest CON IMHO is that we use two different names for the same library, which might be confusing. Also, it requires the `get_libname` complication in the build system

- "nested OUTPUTKIND" PRO: just another axis, no special explanation needed
- CON: very nested, easier to get lost, create combinations which are not actually used, 

- "flattened OUTPUTKIND" PRO: more flatten, very obvious that `debug-testlib` is the version for debug testlibs :)
- CON: less "pure" than "nested OUTPUTKIND", feels like a quick hack.

I honestly don't know what's the best. I think I'm +0.1 towars "flattened OUTPUTKIND" because it seems the simplest way to get the job done, but it's a very mild preference.

Happy/interested to hear your thoughts :)

hoodmane:
> we use two different names for the same library

Well the question is whether it's the same library or a different library. I was feeling that it was a different library because it has different symbols and can't really be used exchangably. But the differences are pretty trivial so maybe it's the same library. 

I'm happy with any of these options I just figured it was easier to pick something somewhat arbitrarily for definiteness and then if you wanted to bike shed it's easier to do once there's a specific proposal.

hoodmane:
So let's go with the flattened output kind.

antocuni:
> Well the question is whether it's the same library or a different library. I was feeling that it was a different library because it has different symbols and can't really be used exchangably. But the differences are pretty trivial so maybe it's the same library.

yes I agree, it's kind of a philosophical question, that's why I wanted your opinion :).
You could also think of two libraries in this sense:
- libspy-core which needs `spy_debug_log` and `spy_debug_set_panic_message`
- libspy-debug which implements them using libc

Then you can say that `testlib` depends on `libspy-core`, and `exe` depends on both `libspy-core` and `libspy-debug`.
But as you say, the differences are very trivial so I don't think it's worth introducing too much complication.

> So let's go with the flattened output kind.

works for me! Thanks 🎉

hoodmane:
One small benefit of "separate library with separate name" is that it is a bit easier to clearly refer to it.

hoodmane:
testlib is also slightly odd, maybe it should be testexe? Since we link a wasi/emscripten executable and use that from the tests, so we're not really using it as a library.

antocuni:
> testlib is also slightly odd, maybe it should be testexe? Since we link a wasi/emscripten executable and use that from the tests, so we're not really using it as a library.

in the wasi case it's not an executable, it's a reactor, so it fits more the definition of "lib" than "exe" IMHO.
Would be nice to have a reactor-style also for emscripten but from what I understand?

antocuni (approved):
Thank you! I like this solution, going to merge it soon

Issue #653, opened 2026-08-21
Explicit vectorization (`SIMD`) in SPy — design plan


paugier:
I got a interesting feedback about my project to work on NumSPy: without vectorization and parallelism, SPy is not the right language for this task. So I thought about these two aspects. I think explicite vectorization is indeed a must have even to start the project and that autovectorization via the C compiler is not sufficient enough (not reliable), in particular in comparison with Mojo.

I produced with Claude a design plan before implementing anything. It seems that explicit vectorization in SPy is reachable with few relativelly simple PRs. I open this issue and share this plan to get feedback before starting to work on it.

---

Status: draft / discussion document

Scope: introduce a hardware-shaped `SIMD[dtype, size]` type in `unsafe`, plus the pointer load/store primitives and stdlib helpers needed to write explicitly vectorized loops in SPy.

## 1. Motivation

C compilers can autovectorize loops, but it's a heuristic that fails silently: no error, just a scalar loop 4-8x slower than it could be. It's especially unreliable on code the compiler can't prove doesn't alias, or with non-trivial trip counts (stencil boundaries), or reductions.

Explicit SIMD types make vectorization a property of the type system instead of a compiler mood: if it compiles, it's vectorized. Two main consumers:

- **DNS/LES-style kernels** (stencils, elementwise combine-and-store, reductions), which today depend entirely on autovectorization firing.
- **NumSPy**: array elementwise operations have no way to express "operate on `W` elements at a time." Explicit SIMD is a natural lowering target for elementwise ufunc-style ops on `ndarray` — instead of interpreter-level per-element dispatch, elementwise ops could compile to exactly the loop shapes below. NumSPy is the primary consumer this plan targets.

## 2. Key technical finding

SPy already has a homogeneous-fixed-size-collection type, `tuple[...]` (`stdlib/_tuple.spy`), built on `struct` (`spy/vm/struct.py`). It **cannot** be extended to support runtime indexing, and this isn't a small gap:

- `W_Struct` stores fields as a **named** dict (`values_w: dict[str, W_Object]`), each with a fixed offset from `calc_layout()`.
- The C backend (`cstructwriter.py`) emits this as a C struct with named members. **C cannot index a named struct field with a runtime value** — which is exactly why `_tuple.spy`'s `__getitem__` restricts indices to blue (compile-time) values.

SIMD fundamentally needs runtime-indexable, contiguous, homogeneous storage — a real array, not N named fields.

### The C-level trick: GCC/Clang vector extensions

[...]

PR #654, opened 2026-08-24
Add interp memory read/write for i8, u32, i64, u64, f32 and corresponding tests

paugier:
This just completes the interp memory read/write for few types currently unsupported.

Weakly related to #653.


spy/tests/compiler/test_struct_ptr.py
```diff
@@ -0,0 +1,95 @@
...
+
+class TestStructByPtr(CompilerTest):
+    """
+    Read/write struct fields through `gc_ptr[Struct]` for every primitive
+    dtype, plus a whole-struct-by-value read.
+
+    See https://github.com/spylang/spy/issues/653 (PR0).
+    """
+
+    def test_ptr_struct_fields_all_dtypes(self):
```
antocuni: Having a whole file just for this test seems very overweight. I vote for just moving it to `test_ptr.py`

spy/tests/compiler/test_struct.py
```diff
@@ -27,7 +27,7 @@ class TestStructOnStack(CompilerTest):
     Test for structs allocated on the stack, passed around by value as
     primitive types.
 
-    There are additional tests for ptr-to-structs in test_unsafe.py.
+    There are additional tests for ptr-to-structs in tests/compiler/unsafe and test_struct_ptr.py.
```
antocuni: reminder to kill this if we move the test to test_ptr.py

spy/tests/compiler/unsafe/test_ptr.py
```diff
@@ -718,3 +718,58 @@ def get_byte(s: str, i: i32) -> u8:
...
+            """
+        )
+        assert mod.rt_i32(42) == 42
+        assert mod.rt_u8(200) == 200
+        assert mod.rt_f64(1.5) == 1.5
+        assert mod.rt_i8(-5) == -5
+        assert mod.rt_u32(42) == 42
+        assert mod.rt_i64(42) == 42
+        assert mod.rt_u64(42) == 42
+        assert mod.rt_f32(7) == 7.0
```
antocuni: maybe we can use values which are representable only in the desired type (e.g. 2**31 + 1 for `i64`)? This way we ensure that we use the correct format for struct.pack/unpack.

antocuni (approved):
+1, feel free to merge after resolving the (small) comments

spy/tests/compiler/test_struct.py
paugier: Done. Note that test_unsafe.py does not exit anymore.

spy/tests/compiler/unsafe/test_ptr.py
paugier: Yes, it is much better. Done.

spy/tests/compiler/test_struct_ptr.py
paugier: Done.

PR #669, opened 2026-09-03
Add initial implementation of input() keyword support

kc611:
As titled, 

This PR aims to add support for the `input()` keyword allowing user input from terminal.

### DIFF
--- a/docs/src/reference/python_builtins.md
+++ b/docs/src/reference/python_builtins.md
@@ -55,6 +55,10 @@ The following built-in functions work similarly to their equivalents in CPython;
 
 :   The `int` type is currently an alias to `i32`. In the future, `int` will alias preferred individual types for specific platforms, but currently it is always `i32`.
 
+### __input__(prompt: str = "") { data-toc-label='input()' }
+
+:   Writes `prompt` to standard out (and flushes it), then reads a line from standard input. The trailing newline is stripped from the result. Raises `EOFError` if standard input is exhausted before any character is read.
+
 ### __len__(object) { data-toc-label='len()' }
 
 :   Return the length (the number of items) in a container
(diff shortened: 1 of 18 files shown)


antocuni:
@kc611 I'm not sure this is the right approach.

What you are doing is essentially to implement most of the logic in C and expose a single function `input` which does the equivalent of `sys.stdin.readline()`. This is not the "spy approach".
The "spy approach" is to write as little C as possible, and to write as much as possible in spy itself.

I propose the following plan:
1. implement/expose `sys.stdin`, `sys.stdout` and `sys.stderr`
2. implement `input()` in spy (e.g. inside `stdlib/_builtins.spy`) on top of `sys.stdin.readline`. 

The good news is that we already have all the machinery needed to implement (1). Look at `stdlib/_file.spy`:
https://github.com/spylang/spy/blob/8ef0a1b7d8796a6c1025fad0d0ac96d52948386a/stdlib/_file.spy#L1-L17

You can see that a `file` is a ptr to `_FileData`, which is in turn a small wrapper around libc `FILE*` type (exposed by the `posix` module).
`posix` is implemented at interp-level in `spy/vm/modules/posix`.

The easiest way is probably to implement helper functions such as `posix._get_stdin` which return a 
`W__FILE` (which corresponds to `_FILE` at app-level), and then wrap it inside a proper applevel `stdlib._file.file` object.

We don't have a `sys` module yet, so as a first approximation I think it's fine to implement `_file.get_stdin()` as a normal function. We will understand how to write that into `sys` later.

If you don't fully follow what I'm saying it's perfectly normal because it cross-cuts a lot of spy layers, so it's normal to be a bit confused about it initially. Happy to talk more about it.

---

**NOTE**

Normally the standard approach to implement `input()` and similar is to link against `libreadline` (or `libedit` on mac) which gives you line editing, history, etc.
But for that we need a bit more machinery on the spy side, because we need to understand how we want to distribute/install/discover compile-time dependencies.

PR #670, opened 2026-09-03
Add initial implementation for str.strip()

kc611:
As titled, 

This PR aims to add initial support for `str.split()`.

### DIFF
--- a/stdlib/_str.spy
+++ b/stdlib/_str.spy
@@ -418,6 +418,15 @@ class methods:
         return True
 
     @blue.metafunc
+    def strip(m_self, *args_m):
+        nargs = len(args_m)
+        if nargs == 0:
+            return OpSpec(_strip_whitespace)
+        elif nargs == 1:
+            return OpSpec(_strip_chars)
+        return OpSpec.NULL
+
+    @blue.metafunc
     def find(m_self, *args_m):
         # str.find(sub[, start[, end]]) -> i32
         nargs = len(args_m)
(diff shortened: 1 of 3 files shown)


examples/1_high_level/str_strip.spy
```diff
@@ -0,0 +1,7 @@
+def main() -> None:
+    text = "    Something important with whitespaces.    "
+    stripped = text.strip()
+    print("Original: ")
+    print(text)
+    print("Stripped: ")
+    print(stripped)
```
antocuni, line 7: I don't think this example is actually needed, else we would need to add one for each str/list/dict methods and it would be too much.

stdlib/_str.spy
```diff
@@ -543,6 +552,62 @@ def _count_3(self: str, sub: str, start: i32, end: i32) -> i32:
...
+    start = 0
+    end = length
+    while start < end:
+        found = False
+        for i in range(chars_len):
+            if ll.utf8[start] == llchars.utf8[i]:
+                found = True
+                break
+        if not found:
+            break
```
antocuni, line 590: what does CPython do in this case? Does it do a linear search among `chars` all the time or it uses some more clever approach?

antocuni (commented):
LGTM, thank you!
The only thing I'd like to understand is what CPython do for the `chars` case. Your current implementation is probably OK if `len(chars)` is small (which I expect to be the common case), but probably very slow in the degenerate case in which `len(chars)` is very big.

stdlib/_str.spy
kc611: So CPython does something called a bloom filter beforehand but it can give out false positives so the linear scan is still done: 

https://github.com/python/cpython/blob/1f25c333a280561e86cc0af69af307b53ddbaaac/Objects/unicodeobject.c#L12675-L12684

https://github.com/python/cpython/blob/main/Objects/unicodeobject.c#L891

I have tried implementing a similar thing in the latest commit.

antocuni (commented):
oh wow, I didn't expect CPython to be so advanced.
I'm fine to copy the bloom filter from CPython but we should probably document it better.
I found myself spending quite a bit of time reading CPython source code to understand how it works, but I don't thin it's a good long-term strategy.

What I would like is:
1. put bloom filter logic into `bloom_*` functions
2. add unit tests for them
3. have some comment which explain how they work.

This PR is becoming much more advanced/time consuming that I originally anticipated though.
If you want, we can merge the "simple" implementation first, and leave the bloom filter as a follow-up. Or if you prefer, you can keep working on this.

kc611:
Happy to split the PR. That'll make things easier on the review side.

antocuni (approved):
good idea! This PR is good, merging as soon as tests are green

PR #672, opened 2026-09-05
Generic metafunc

paugier:
A metafunc that has to be called with [].

Potentially useful for `gc_ptr[T, N]` / `gc_ptr[T]`.

@antocuni do you think it makes sense to add this?

### DIFF
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -64,7 +64,7 @@ LoweringStage = typing.Literal[
 
 
 ClassKind = typing.Literal["class", "struct"]
-FuncKind = typing.Literal["plain", "generic", "metafunc"]
+FuncKind = typing.Literal["plain", "generic", "metafunc", "generic_metafunc"]
 FuncParamKind = typing.Literal["simple", "var_positional"]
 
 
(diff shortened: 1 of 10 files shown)


antocuni:
it makes sense but it might add unnecessary complexity.
Metafunc are useful to do blue-time dispatch of red opimpl.
On the other hand, generic funcs are supposed to be called a blue time where we have less constraints on what we accept and how we deal with arguments.
E.g., the typical use case for metafunc is overloading on the number of arguments or type of arguments, but with a `blue.generic` func we should be able to achieve the same effect with "python style" dynamic dispatch, something like this:

```python
@blue.generic
def gc_ptr(T, *args):
    if len(*args) == 0:
        align = alignof(T)
    elif len(*args) == 1:
        align = args[0]
    else:
        raise TypeError
    ...
```

or, as we discussed on discord, in this specific case default args are even better:
```python
@blue.generic
def gc_ptr(T, align=-1):
    if align == -1:
        align = alignof(T)
    ...
```

Unless we have a real need to a generic metafunc, I'd vote for closing this issue and reconsider it if/when we need it

paugier:
Yes, it adds additional complexity. With something like #673, I think this it not needed.

antocuni:
@paugier should we close it?

paugier:
Yes, I close it.

PR #673, opened 2026-09-06
Optional arg for generic function at interp level

paugier:
I think this is needed for `gc_ptr[T, N]`.

### DIFF
--- a/spy/tests/compiler/operator/test_itemop.py
+++ b/spy/tests/compiler/operator/test_itemop.py
@@ -4,7 +4,7 @@ from spy.vm.opspec import W_MetaArg, W_OpSpec
 from spy.vm.primitive import W_I32
 from spy.vm.registry import ModuleRegistry
 from spy.vm.vm import SPyVM
-from spy.vm.w import W_Object
+from spy.vm.w import W_Object, W_Type
 
 
 class W_MyClass(W_Object):
@@ -128,3 +128,27 @@ class TestItemop(CompilerTest):
         assert mod.set_and_get(0, 2, 24) == 24
         assert mod.set_and_get(2, 1, 99) == 99
         assert mod.get_default(1, 2) == 0
+
+    def test_generic_builtin_func_with_optional_arg(self):
+        EXT = ModuleRegistry("ext")
+
+        @EXT.builtin_func(color="blue", kind="generic")
+        def w_scaled(vm: "SPyVM", w_T: W_Type, w_N: W_I32 | None = None) -> W_I32:
+            if w_N is None:
+                n = 100
+            else:
+                n = vm.unwrap_i32(w_N)
+            return vm.wrap(n)
+
+        self.vm.make_module(EXT)
+        mod = self.compile("""
+        from ext import scaled
+
+        def with_default() -> i32:
+            return scaled[i32]
+
+        def with_explicit_n() -> i32:
+            return scaled[i32, 42]
+        """)
+        assert mod.with_default() == 100
+        assert mod.with_explicit_n() == 42
--- a/spy/ast.py
+++ b/spy/ast.py
@@ -65,7 +65,7 @@ LoweringStage = typing.Literal[
 
 ClassKind = typing.Literal["class", "struct"]
 FuncKind = typing.Literal["plain", "generic", "metafunc"]
-FuncParamKind = typing.Literal["simple", "var_positional"]
+FuncParamKind = typing.Literal["simple", "var_positional", "optional"]
 
 
 @extend(py_ast.AST)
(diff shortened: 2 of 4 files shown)


spy/tests/compiler/operator/test_itemop.py
```diff
@@ -128,3 +128,27 @@ def get_default(i: i32, j: i32) -> i32:
...
+    def test_generic_builtin_func_with_optional_arg(self):
+        EXT = ModuleRegistry("ext")
+
+        @EXT.builtin_func(color="blue", kind="generic")
+        def w_scaled(vm: "SPyVM", w_T: W_Type, w_N: W_I32 | None = None) -> W_I32:
+            if w_N is None:
+                n = 100
+            else:
+                n = vm.unwrap_i32(w_N)
+            return vm.wrap(n)
```
antocuni, line 141: this test seems a bit weird: `w_T` is ignored, and `scaled` seems a strange name (what does it scale?)

Maybe you can declare something like `inc[n, delta=1]`?

antocuni (commented):
I don't understand what/why is needed

e.g. why do you need to introduce FuncParamKind.optional? I don't remember HOW default arguments are implemented for plain functions, but if it's not needed for them why is it needed for generic funcs?

antocuni:
I fear it's more complicated than expected :(

Look for example here:
https://github.com/spylang/spy/blob/f3be6ecd407e61ff0c20b2f7d7efbae1bd7c333c/spy/vm/function.py#L190-L202

here:
https://github.com/spylang/spy/blob/f3be6ecd407e61ff0c20b2f7d7efbae1bd7c333c/spy/vm/vm.py#L941-L942

and here:
https://github.com/spylang/spy/blob/f3be6ecd407e61ff0c20b2f7d7efbae1bd7c333c/spy/vm/typechecker.py#L104-L138

There is the assumption that we can typecheck a builtin function by iterating over its `all_params()`, but if you introduce default arguments then all this logic is completely broken.

I think it's related to something which I got wrong at the beginning and fixed it mid way, but probably not completely.
Currently, we use `ast.Call` for two different things:
- a "low-level" C-like call where we have only positional arguments and that it can be easily translated to C
- a "high-level" Python-like call with keyword arguments (eventually), default arguments, and all possible python's whistle&bells

There is already the concept of `fastcall` which more or less maps to the low-level call I mention above, but I think it's not completely consistent.

A possible scenario could be this:
- the `ast.Call()` node which is generated by DopplerFrame to call opimpl should be an `ast.FastCall`.
- `ast.FastCall` can ONLY call builtin functions and AST functions, it's not possible to override by the user (basically, if something has a FastCall, it must have an equivalent C function) 
- In this world, `W_FuncType` would refer to the *fastcall* signature (I think it's already implicitly the case).
- for builtin default arguments, the `W_FuncType` would NOT take default args into account, and it's the job for `W_BuiltinFunc.__call__` to insert them correctly (as it happens for ASTFuncs).
It might be that this works, but as far as I know it might also open a can of worms.

Sorry for driving you in this rabbit hole :).

---

Btw, another comment on this PR which should be automatically solved by the proposal above:

Let's take an app-level func with default args:
```python
def foo(x: int, y: int = 1) -> int:
    return x + y
```

Currently, the `FuncParamKind` of `y` is `simple`. Your proposal introduces `optional` but ONLY for interp-level functions, which is a bit weird.

paugier:
I see that it is much more complicated than I thought.

This seems like a prerequisite for what I'd like to do on `gc_ptr[Y, N]` and then SIMD. But to be honest, the plan with FastCall seems too big for me. It does not seem like a small fix but more like relatively deep changes to parts of SPy that I don't understand.

antocuni:
> I see that it is much more complicated than I thought.
> 
> This seems like a prerequisite for what I'd like to do on `gc_ptr[Y, N]` and then SIMD. But to be honest, the plan with FastCall seems too big for me. It does not seem like a small fix but more like relatively deep changes to parts of SPy that I don't understand.

yes I agree and I'm happy to do that (eventually :)).
For `gc_ptr[T, N]`, I suggest to use the trick I showed on discord:
```python
@UNSAFE.builtin_func(color="blue", kind="generic")
def w_gc_ptr(vm: "SPyVM", w_T: W_Type, *args_w: W_Dynamic) -> W_Dynamic:
    """
    The gc_ptr[T] generic type
    """
    if len(args_w) == 0:
        alignment = calc_default_aligment()
    elif len(args_w) == 1:
        # XXX typecheck
        alingment = vm.unwrap_i32(args_w[0])
    else:
        n = len(args_w)
        raise SPyError("W_TypeError", f"gc_ptr accepts 1 or 2 arguments, got {n}")

    fqn = FQN("unsafe").join("gc_ptr", [w_T.fqn])  # unsafe::gc_ptr[i32]
    w_ptrtype = W_PtrType.from_itemtype(fqn, "gc", w_T)
    return w_ptrtype
```

If you don't mind, I'm going to close this PR and open an issue (assigned to me)

PR #675, opened 2026-09-06
Add shortest-round-trip formatting for f64

filiplajszczak:
The previous C backend implementation formatted `f64` values using `%g` with its default precision of six significant digits. Some results therefore did not round-trip to the original binary64 value and differed from interpreter formatting.

## Changes

- Vendor Ryu's binary64 shortest-format implementation.
- Generalize the existing SPy-owned Ryu normalizer to support binary64's 17 significant digits and three-digit exponents.
- Add an exported `spy_f64_to_str` formatter around `d2s_buffered_n`.
- Route `f64` `str` and `repr` through the same libspy implementation in the interpreter, Doppler, and generated C backends.
- Preserve signed zero, non-finite values, float type markers, notation boundaries, subnormals, and values requiring full binary64 precision.
- Use Ryu's portable 64-bit d2s path to avoid Zig WASM multivalue ABI conflicts in compiler-runtime helpers.


spy/tests/compiler/test_str.py
```diff
@@ -263,10 +263,6 @@ def type_str() -> str:
         assert mod.type_str() == "<spy type 'i32'>"
 
     def test_str_numbers(self):
-        # NOTE: float2str produces slightly different results in Python vs C
-        # backend: e.g. str(0.0) == '0' in Python, '0.0' in the C backend.
-        # Eventually, we want to port the formatting code from CPython, but
-        # for now we just allow both results and keep the C backend simple.
```
antocuni, line 269: love that we could finally kill this 🎉

spy/tests/compiler/test_float.py
```diff
@@ -91,6 +91,41 @@ def f32_equal(x: f32, y: f32) -> bool:
...
+        """)
+
+        def assert_roundtrip(x: float, expected: str):
+            result = mod.repr_f64(x)
+            assert result == expected
+            assert float(result) == x
+
+        assert_roundtrip(1.0000000000000002, "1.0000000000000002")
+        assert_roundtrip(5e-324, "5e-324")
+        assert_roundtrip(1.7976931348623157e308, "1.7976931348623157e+308")
```
antocuni, line 127: note that I slightly modified this test as I dislike "loops inside a test", and I think this pattern is already used elsewhere in the codebase.

antocuni (approved):
LGTM, thank you.

PR #676, opened 2026-09-07
Aligned pointers and pointer reinterpret-cast

paugier:
- [x] blue `alignof`
- [x] `gc_ptr[T, N=alignof(T)]`
- [x] `gc_alloc[T, N=alignof(T)]`
- [x] `spy_alloc_aligned_impl` C helper function
- [x] load/store with `__builtin_memcpy` when `alignment < alignof(T)`
- [ ] ~~`cast[DstItemT](ptr)`~~
- [ ] ~~`align_cast[N](ptr)`~~
- [ ] ~~`align_offset(ptr, N, n)`~~

It seems to me that it reasonable to not include cast functions in this first PR and to focus on the core of the problem.

The principles and the API have been discussed on Discord.

### DIFF
--- a/spy/tests/compiler/unsafe/test_align_alloc.py
+++ b/spy/tests/compiler/unsafe/test_align_alloc.py
@@ -10,8 +10,6 @@ over-allocate, round the pointer up, and discard the base pointer.
 import pytest
 
 from spy.tests.support import CompilerTest
-from spy.tests.wasm_wrapper import WasmPtr
-from spy.vm.modules.unsafe.ptr import W_Ptr
 
 # Alignment value guaranteed to exceed SPY_BASE_ALIGNMENT on all targets
 # (16 on native-64, 8 on wasm).
@@ -24,30 +22,19 @@ def memkind(request):
 
 
 class TestOverAllocAlignment(CompilerTest):
-    def _get_addr(self, p: W_Ptr | WasmPtr) -> int:
-        """Extract the raw address from a returned pointer."""
-        if self.backend in ("interp", "doppler"):
-            assert isinstance(p, W_Ptr)
-            return p.addr
-        else:
-            assert isinstance(p, WasmPtr)
-            return p.addr
-
     def test_ptr_address_is_aligned(self, memkind):
         k = memkind
         mod = self.compile(
             """
-        from unsafe import {k}_alloc as k_alloc, {k}_ptr as k_ptr
+        from unsafe import {k}_alloc as k_alloc, {k}_ptr as k_ptr, ptr_to_addr
 
         def alloc() -> k_ptr[i32, {N}]:
-            return k_alloc[i32, {N}](4)
+            p = k_alloc[i32, {N}](4)
+            assert ptr_to_addr(p) % {N} == 0
+            return p
         """.format(k=k, N=OVER_ALIGNMENT)
         )
-        p = mod.alloc()
-        addr = self._get_addr(p)
-        assert addr % OVER_ALIGNMENT == 0, (
-            f"expected {OVER_ALIGNMENT}-byte aligned address, got {addr}"
-        )
+        mod.alloc()
 
     def test_over_alloc_roundtrip(self, memkind):
         k = memkind
@@ -107,10 +94,12 @@ class TestOverAllocAlignment(CompilerTest):
         # properly aligned (not just the first one)
         mod = self.compile(
             """
-        from unsafe import gc_alloc, gc_ptr
+        from unsafe import gc_alloc, gc_ptr, ptr_to_addr
 
         def alloc() -> gc_ptr[i32, {N}]:
-            return gc_alloc[i32, {N}](1)
+            p = gc_alloc[i32, {N}](1)
+            assert ptr_to_addr(p) % {N} == 0
+            return p
 
         def foo() -> i32:
             a = alloc()
@@ -123,11 +112,9 @@ class TestOverAllocAlignment(CompilerTest):
         """.format(N=OVER_ALIGNMENT)
         )
         assert mod.foo() == 6
-        # verify that each individual allocation is aligned
+        # a few more independent allocations, for extra confidence
         for _ in range(3):
-            p = mod.alloc()
-            addr = self._get_addr(p)
-            assert addr % OVER_ALIGNMENT == 0
+            mod.alloc()
 
     def test_over_alloc_f64(self):
         mod = self.compile(
@@ -146,14 +133,12 @@ class TestOverAllocAlignment(CompilerTest):
     def test_over_alloc_address_is_aligned_f64(self):
         mod = self.compile(
             """
-        from unsafe import gc_alloc, gc_ptr
+        from unsafe import gc_alloc, gc_ptr, ptr_to_addr
 
         def alloc() -> gc_ptr[f64, {N}]:
-            return gc_alloc[f64, {N}](2)
+            p = gc_alloc[f64, {N}](2)
+            assert ptr_to_addr(p) % {N} == 0
+            return p
         """.format(N=OVER_ALIGNMENT)
         )
-        p = mod.alloc()
-        addr = self._get_addr(p)
-        assert addr % OVER_ALIGNMENT == 0, (
-            f"expected {OVER_ALIGNMENT}-byte aligned address, got {addr}"
-        )
+        mod.alloc()
--- a/spy/tests/compiler/unsafe/test_align_alloc.py
+++ b/spy/tests/compiler/unsafe/test_align_alloc.py
@@ -1,0 +1,159 @@
+"""
+Tests for the spy_alloc_aligned_impl C helper function.
+
+These tests exercise the *over-allocation* path: when the requested
+alignment exceeds SPY_BASE_ALIGNMENT (16 on native-64, 8 on wasm), the
+allocator can no longer rely on the base allocator's guarantee and must
+over-allocate, round the pointer up, and discard the base pointer.
+"""
+
+import pytest
+
+from spy.tests.support import CompilerTest
+from spy.tests.wasm_wrapper import WasmPtr
+from spy.vm.modules.unsafe.ptr import W_Ptr
+
+# Alignment value guaranteed to exceed SPY_BASE_ALIGNMENT on all targets
+# (16 on native-64, 8 on wasm).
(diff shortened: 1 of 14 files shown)


spy/tests/compiler/unsafe/test_alignof.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_all_i32_fields(self):
+        src = """
+            @struct
+            class Point:
+                x: i32
+                y: i32
+            """
+        mod = self.compile(src)
+        w_Point = mod.w_mod.getattr("Point")
+        assert alignof(w_Point) == 4
```
antocuni: these tests looks a bit weird.
I'd write them this way:
```
        src = """
        @struct
        class Point:
            x: i32
            y: i32

        def alignof_Point() -> i32:
            return alignof(Point)            
        """
        mod = self.compile(src)
        assert mod.alignof_Point() == 4
```

no need to test `alignof` at interp-level, it's already tested by `test_alignof_primitives` and implicitly by these compiler tests.

Also, very nitpicking but worth flagging while I'm here: the convention is that the spy code is aligned _at the same_ level as `src`:
```
src = """
# GOOD: spy code here
"""

src = """
    # BAD: spy code indented
"""
```

spy/tests/compiler/unsafe/test_alignof.py
```diff
@@ -0,0 +1,108 @@
...
+            from unsafe import alignof
+            from __spy__ import COLOR
+
+            def foo() -> i32:
+                N = alignof(i32)
+                assert COLOR(N) == "blue"
+                return N
+            """
+        mod = self.compile(src)
+        assert mod.foo() == 4
```
antocuni: this is also very weird. It basically tests that `@UNSAFE.builtin_func(color="blue")` is actually blue, which we already test extensively everywhere else.

spy/tests/compiler/unsafe/test_ptr.py
```diff
@@ -835,3 +835,89 @@ def rt_struct_byval_f32(v: f32) -> f32:
...
+    @only_interp
+    def test_explicit_alignment_in_fqn(self):
+        w_ptrtype = self.vm.fast_call(UNSAFE.w_gc_ptr, [B.w_i32, self.vm.wrap(8)])
+        assert repr(w_ptrtype) == "<spy type 'unsafe::gc_ptr[i32, 8]'>"
+
+    @only_interp
+    def test_different_alignments_are_different_types(self):
+        w_8 = self.vm.fast_call(UNSAFE.w_gc_ptr, [B.w_i32, self.vm.wrap(8)])
+        w_16 = self.vm.fast_call(UNSAFE.w_gc_ptr, [B.w_i32, self.vm.wrap(16)])
+        assert w_8 is not w_16
```
antocuni: I think these should either become "real compiler tests" (using self.compile) or be moved outside of the class

spy/tests/compiler/unsafe/test_ptr.py
```diff
@@ -835,3 +835,89 @@ def rt_struct_byval_f32(v: f32) -> f32:
...
+        src = """
+            from unsafe import gc_alloc, gc_ptr
+
+            def foo() -> i32:
+                p: gc_ptr[i32] = gc_alloc[i32, 4](1)
+                p[0] = 7
+                return p[0]
+            """
+        mod = self.compile(src)
+        assert mod.foo() == 7
```
antocuni: this is a VERY weird way of testing it. Can'y we just `return gc_ptr[i32] is gc_ptr[i32, 4]`?

spy/tests/compiler/unsafe/test_ptr.py
```diff
@@ -835,3 +835,89 @@ def rt_struct_byval_f32(v: f32) -> f32:
...
+        src = """
+            from unsafe import gc_alloc, gc_ptr
+
+            def foo() -> i32:
+                p: gc_ptr[i32, 8] = gc_alloc[i32, 8](1)
+                p[0] = 42
+                return p[0]
+            """
+        mod = self.compile(src)
+        assert mod.foo() == 42
```
antocuni: I'm not sure this is testing anything, but I might be wrong

spy/tests/compiler/unsafe/test_under_align.py
```diff
@@ -0,0 +1,77 @@
+"""
+Tests for load/store with __builtin_memcpy when alignment < alignof(T).
+"""
```
antocuni: why these are not part of `test_ptr`?

spy/vm/modules/unsafe/mem.py
```diff
@@ -10,18 +10,44 @@
...
 if TYPE_CHECKING:
     from spy.vm.vm import SPyVM
 
 
+# The base alignment that the interp-level allocators (spy_raw_alloc,
+# spy_nogc_alloc, which are just malloc) already guarantee.  This must
+# match the C-side SPY_BASE_ALIGNMENT in spy/libspy/include/spy/unsafe.h.
+# On wasm32/wasm64, malloc returns 8-byte-aligned pointers; on native
+# 16-byte alignment.
+SPY_BASE_ALIGNMENT_INTERP = 8
```
antocuni: just a note for the future: ideally, we woule like to avoid magic numbers.
We might be able to use the same technique that we use for `_spy_StrObject_layout()` to make sure that this number matches the one used by the C compiler.

But, I think that `sizeof()` suffers of the same problem, so probably this must be fixed by some othe PR

spy/vm/modules/unsafe/ptr.py
```diff
@@ -54,25 +52,52 @@
...
+def w_raw_ptr(vm: "SPyVM", w_T: W_Type, *args_w: W_Dynamic) -> W_Dynamic:
+    """
+    The raw_ptr[T] / raw_ptr[T, N] generic type
+    """
+    if len(args_w) == 0:
+        # raw_ptr[T] is just the common-case spelling of
+        # raw_ptr[T, alignof(T)]. Type identity across calls comes purely
+        # from the blue-call cache, which keys on (func, args_w). So we
+        # recurse through the 2-arg call, landing on the same cache entry
+        # and therefore the same W_PtrType object.
```
antocuni:
```suggestion
        # raw_ptr[T] is just a shortcut for raw_ptr[T, alignof(T)]
```

Recent claude models tend to produce *very* verbose comments in which they "dump" their reasoning.

spy/vm/modules/unsafe/misc.py
```diff
@@ -73,3 +81,74 @@ def contains_gc_ptr(w_T: W_Type) -> bool:
...
+    elif isinstance(w_T, W_StructType):
+        if not w_T.is_defined():
+            # not-yet-defined struct (e.g. a struct that (transitively)
+            # points to itself, or one of the special bootstrapping
+            # struct types like _str::StrObject that this function may be
+            # asked about before its fields are populated -- see the
+            # analogous comment in W_MemLocType.from_itemtype). We can't
+            # look at fields that don't exist yet, so fall back to 1
+            # rather than crashing;
+            return 1
```
antocuni: this sounds *very* bad.
Because of how blue caching works it means that in this case:

```
@struct
class Point:
    x: int
    y: int
    next: gc_ptr[Point]
```

`gc_ptr[Point]` will fallback to `gc_ptr[Point, 1]`, and because of blue caching *it will stay it forever*.

This seems just wrong?

I don't have an easy solution for this problem though :(

spy/vm/modules/unsafe/misc.py
```diff
@@ -73,3 +81,74 @@ def contains_gc_ptr(w_T: W_Type) -> bool:
...
+
+def parse_optional_alignment(
+    vm: "SPyVM", w_T: W_Type, args_w: tuple, funcname: str
+) -> int:
+    """
+    Shared arg-parsing for the optional, defaulted alignment type param on
+    {raw,gc}_ptr[T, N=alignof(T)] / {raw,gc}_alloc[T, N=alignof(T)].
+    `args_w` is whatever extra positional blue args were
+    passed after `T`: zero (use the default) or one (an i32 `N`).
+    """
```
antocuni: a bit of nitpick, but factoring this logic into its own function seems overkill, to avoid duplicating very little logic.
Not a merge blocker though, just my personal comment, feel free to leave it like now

spy/vm/modules/unsafe/ptr.py
```diff
@@ -54,25 +52,52 @@
...
+    """
+    unsafe::{raw,gc}_ptr[T] when `alignment` is T's natural alignment
+    (the default) else unsafe::{raw,gc}_ptr[T, N], so that an explicit,
+    non-default alignment is part of the type's identity
+    (gc_ptr[T,N] and gc_ptr[T,M] for N != M are distinct types).
+    """
+    qualifiers: list = [w_T.fqn]
+    if alignment != alignof(w_T):
+        qualifiers.append(str(alignment))
+    return FQN("unsafe").join(funcname, qualifiers)
```
antocuni: same comment as above. Do we really need a 18-line function to save 2 lines?

spy/vm/modules/unsafe/ptr.py
```diff
@@ -179,6 +211,11 @@ def from_itemtype(cls, fqn: FQN, memkind: MEMKIND, w_itemT: W_Type) -> Self:
             assert False
         w_T.memkind = memkind
         w_T.w_itemT = w_itemT
+        if alignment is None:
+            from .misc import alignof
+
+            alignment = alignof(w_itemT)
```
antocuni: I think `alignof` is already imported?

spy/tests/compiler/unsafe/test_align_alloc.py
```diff
@@ -0,0 +1,137 @@
+"""
+Tests for the spy_alloc_aligned_impl C helper function.
+
+These tests exercise the *over-allocation* path: when the requested
+alignment exceeds SPY_BASE_ALIGNMENT (16 on native-64, 8 on wasm), the
+allocator can no longer rely on the base allocator's guarantee and must
+over-allocate, round the pointer up, and discard the base pointer.
+"""
```
antocuni: if we want to unit-test `spy_alloc_aligned_impl`, it might be a better idea to write a `CTest` and test it directly.

If we want to write spy compiler tests, then same comment as `test_under_align`: I think they should move to `test_ptr.py`

PR #677, opened 2026-09-07
Add value-based repr for bool

filiplajszczak:
Boolean values previously had inconsistent `repr` behavior:

- the interpreter inherited `object`’s identity-based representation;
- C builds generated a call to a missing generic `repr` symbol.

Define `bool.__repr__` to share `bool.__str__` formatting, giving `True` and `False` the same canonical representation on every backend.


antocuni (approved):
LGTM, thank you!

PR #678, opened 2026-09-07
Better scoping rules

antocuni:
This is a giant PR which heavily refactors and improves scoping rules.
From the user perspective, the high-level overview is this:
  - `strict_scoping` rules are similar to what you find in other programming languages: every variable must be declared, each block opens a nested scope, variables declared in inner scopes can shadow the ones declared in outer scopes
  - `pythonic_scoping` rules are similar (but not identical!) to what you have in Python: implicit declaration via assignment, a limited form of type inference and the illusion of a flat per-function namespace

`pythonic_scoping` is implemented as syntax sugar on top of `strict_scoping`, which means that subsequent stages of the pipeline don't need to worry about it; it is the default, and you can enable strict scoping by doing `from __spy__ import strict_scoping` at the beginning of a module.

The full rules are described in docs/src/reference/scoping.md

A breaking change w.r.t. the old rules is that to modify a global variable you need `global x`, as in Python.

Also, the old `spy symtable` command is now `spy scopes`.

---

From the implementation point of view, the main points are these.
We no longer have `SymTable`: it has been replaced by `Scope`, which represents a nested tree of symbols and it's used only during scope analysis time, and `FrameInfo`, which represents the flat layout of local variables used at runtime.

`Symbol` now has two "names": `src_name` is what's written in the source, and `slot_name` is the unique id in the flatten `FrameInfo`, used e.g. to identify a local variable. Because of this, local vars are now mangled as e.g. `x$0`, `x$1`, etc.

`astcompile` and `force_inline` can reuse the same mechanism to get unique slots for temporary variables: this is nicer and kills the ad-hoc ways to get uniquely mangled names.

`SPdb` and `astcompile_interactive` required a big refactoring to be able to inspect the locals of a function: the user types a `src_name`, but we need to know how to map it to the corresponding `slot_name`. To do that, we need the full tree of `Scope` as built by `ScopeAnalyzer`. We save it in the .spyc as some kind of "debug info".

At the AST level, we introduce `ast.Block` as the container of "a list of statements which introduces a scope"; `ast.Block` is not rendered in dumps to keep the output more readable.

In the old system, we had `var_storage = "NameError"` to signal a lazy error. This is now replaced by a more general notion of `ast.PoisonExpr`, which can be used in place of any expression and results in a lazy error at runtime.


docs/src/reference/scoping.md
```diff
@@ -0,0 +1,1019 @@
...
+
+### `[scope.loop-fresh]` A block-local is fresh on each loop iteration { #scope-loop-fresh }
+
+```python
+def f() -> None:
+    for i in range(3):
+        var x: i32
+        if i == 0:
+            x = 1
+        print(x)         # when i == 1: `x` is unassigned
```
paugier, line 342: It's not very clear what happens in this case in practice: compile error or runtime error?

docs/src/reference/scoping.md
```diff
@@ -0,0 +1,1019 @@
...
+([`[py.scope-lifting]`](#py-scope-lifting)).
+
+```python
+COND = True
+
+def main() -> None:
+    if COND:
+        def g() -> i32:
+            return 42
+    print(g())           # ERROR: `g` is local to the `if` body
```
paugier: Wouldn't it be clearer to introduce lifting first before `[py.def-class]`. Also, it is not clear why lifting does not work in this precise case.

docs/src/reference/scoping.md
```diff
@@ -0,0 +1,1019 @@
...
+
+A write-only shadow stays silent. We are aware of it and accept it for now;
+real usage will tell us whether it deserves a diagnostic (see
+[`[future.shadow-guard]`](#future-shadow-guard)).
+
+```python
+var COUNT: i32 = 0
+
+def reset() -> None:
+    COUNT = 0            # a dead local; the global is untouched
```
paugier, line 851: Could be slightly better with `COUNT = 1`?

paugier (commented):
I just read the new doc and just have few questions.

Is it true that in strict scoping, all variables must be declared (with an explicit MODIFIER) except loop targets? This is what I understand from the doc but it is not written explicitly.

Note also that I was first tempted to think that in script scoping, the type was mandatory in the declaration (or `auto`).

docs/src/reference/scoping.md
```diff
@@ -0,0 +1,1019 @@
...
+`MODIFIER` can be:
+
+  - `const`: the name is assigned only once
+  - `var`: the name can be reassigned
+
+In strict scoping, `MODIFIER` is mandatory. Under pythonic scoping it can be
+omitted and inferred from the number of assignments, see
+[`[py.constness]`](#py-constness).
+
+If `TYPE` is `auto` or omitted, the type is inferred, see
```
paugier, line 54: Using `auto` is exactly equivalent to no type, isn't it?

docs/src/reference/scoping.md
```diff
@@ -0,0 +1,1019 @@
...
+MODIFIER name: TYPE = initializer
+```
+
+`MODIFIER` can be:
+
+  - `const`: the name is assigned only once
+  - `var`: the name can be reassigned
+
+In strict scoping, `MODIFIER` is mandatory. Under pythonic scoping it can be
+omitted and inferred from the number of assignments, see
```
paugier, line 51: "inferred from the number of assignments, see `[py.constness]`" <- this formulation is a bit too restricted to local variables.

docs/src/reference/scoping.md
antocuni: what happens when reading uninitialized vars is explained by `[decl.initializer]`: runtime error in the interp, undefined behavior when compiled.

This is not ideal of course: the idea is to fix it by implementing `[future.definite-assignment]`, but this PR was already big enough that I didn't want to do it now (also, in the old system we had the same problem)

docs/src/reference/scoping.md
antocuni: agreed. I move it after all the lifting rules, see 3971c493

docs/src/reference/scoping.md
antocuni: agreed, done by 5c0d5efe

docs/src/reference/scoping.md
antocuni: yes. Do you think there could be a better wording?

docs/src/reference/scoping.md
antocuni: why?

antocuni:
> Is it true that in strict scoping, all variables must be declared (with an explicit MODIFIER) except loop targets? This is what I understand from the doc but it is not written explicitly.

colloquially speaking, yes. Formally speaking not because there are many syntactical constructs which introduce new names, so we can consider them "explicit" from this POV.
E.g.:
```
for NAME in x: ...
with x as NAME: ...
except Exception as NAME:
def NAME(): ...
class NAME: ...
```

> Note also that I was first tempted to think that in script scoping, the type was mandatory in the declaration (or `auto`).

improvements to the doc are VERY welcome.
I have been immersed in this docs/branch for weeks now, so it's very hard to me to read the docs with "fresh eyes" and I'm not surprised that it can be improved.

I propose to proceed with merging (because I'm really fed up with this by now 😂) and then work on improved docs on a follow-up PR.

@paugier also, are you going to read/review the rest of the branch? I was about to press "merge", but if you are doing the review, I'm happy to wait.

Unfortunately it's a very big PR and as such it's not easy to review: the problem is that I couldn't find a good self-contained intermediate state where I could split. In retrospect, I should have done a stack of PRs, but it's too late.

paugier:
Congratulation for this huge and useful work. I understand that you are now fed up with it. Please merge as it is and we will be able to improve the doc if and when needed.

PR #679, opened 2026-09-07
Add value-based repr for complex128

filiplajszczak:
Complex values previously inherited `object`’s identity-based representation in the interpreter, while C builds generated a call to a missing generic `repr` symbol.

Implement `complex128.__str__` and `complex128.__repr__` once in app-level SPy, exposed through lazy attributes like the existing `str` and `bytes` methods. The formatter composes canonical `f64` strings according to CPython’s complex formatting rules, so interpreter, Doppler, and C builds share the same behavior.

Adapt CPython’s tests for ordinary values, infinities, NaNs, pure imaginary values, and the full signed-zero matrix.

### DIFF
--- a/spy/tests/compiler/test_complex.py
+++ b/spy/tests/compiler/test_complex.py
@@ -19,6 +19,44 @@ class TestComplex(CompilerTest):
         """)
         assert mod.foo() == 12.3j
 
+    def test_repr(self):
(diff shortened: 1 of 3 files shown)


antocuni (approved):
Good idea to use lazy methods for this! Merging

PR #682, opened 2026-09-08
Add repr support for range

filiplajszczak:
`range` previously inherited `object`’s identity-based representation, and compiled backends could fail while lowering that generic path.

Implement `range.__repr__` in app-level SPy using CPython’s normalized spelling: always include `start` and `stop`, and include `step` only when it differs from one. `str(range)` inherits the same representation through the existing object fallback.


antocuni (approved):
LGTM

PR #684, opened 2026-09-09
Add length support for range

filiplajszczak:
Implement `range.__len__` in app-level SPy using CPython’s arithmetic for positive and negative steps. Empty and direction-mismatched ranges return zero; non-empty ranges use constant-time integer calculation rather than iteration.

Cover forward, reverse, empty, negative-bound, and mismatched-direction cases.


antocuni (approved):
LGTM

PR #685, opened 2026-09-09
Add integer indexing for range

filiplajszczak:
Implement constant-time `range[index]` lookup in app-level SPy using the range start, step, and computed length.

Normalize negative indices like CPython and raise `IndexError` for positive overflow, overly negative indices, and empty ranges.

### DIFF
--- a/spy/tests/stdlib/test__range.py
+++ b/spy/tests/stdlib/test__range.py
@@ -62,6 +62,31 @@ class TestRange(CompilerTest):
         assert mod.range_len(5, 5, 1) == 0
         assert mod.range_len(-10, -1, 2) == 5
 
+    def test_getitem(self):
+        mod = self.compile("""
+        from _range import range
+
+        def getitem(start: int, stop: int, step: int, index: int) -> int:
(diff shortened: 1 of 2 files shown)


spy/tests/stdlib/test__range.py
```diff
@@ -62,6 +62,32 @@ def range_len(start: int, stop: int, step: int) -> int:
...
         assert mod.range_len(-10, -1, 2) == 5
 
+    @pytest.fixture
+    def getitem_mod(self):
+        return self.compile("""
+        from _range import range
+
+        def getitem(start: int, stop: int, step: int, index: int) -> int:
+            return range(start, stop, step)[index]
+        """)
```
antocuni: this seems a bit overengineered, but also we need to recompile the same stuff again and again for all the tests.
I think it's fine to have a single `test_getitem` where you do all the checks (maybe you can use comments to split the asserts into logical sections0

filiplajszczak: Fair enough. I'll fold it in.

antocuni (approved):

PR #687, opened 2026-09-10
Add initial support for string .join() method

kc611:
As titled, 

This PR intends to add support for `str.join()` method.

Note: I did not use blue metafunc because the right now `join()` implementation has a fixed signature: `join(self, items: list[str])`. If we want to support something like `join(*items)` I can change it into a blue metafunc as required.

### DIFF
--- a/examples/1_high_level/string_manipulation.spy
+++ b/examples/1_high_level/string_manipulation.spy
@@ -1,0 +1,6 @@
+def foo(s: str) -> None:
+    print(" ".join(s.split()))
+
+
+def main() -> None:
+    foo("hello world from spy")
--- a/examples/expected_output/string_manipulation.txt
+++ b/examples/expected_output/string_manipulation.txt
@@ -1,0 +1,1 @@
+hello world from spy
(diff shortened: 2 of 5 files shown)


examples/1_high_level/string_manipulation.spy
```diff
@@ -0,0 +1,6 @@
+def foo(s: str) -> None:
+    print(" ".join(s.split()))
+
+
+def main() -> None:
+    foo("hello world from spy")
```
antocuni, line 6: I don't think we need an example just for `join`, let's kill this

examples/expected_output/string_manipulation.txt
```diff
@@ -0,0 +1 @@
+hello world from spy
```
antocuni, line 1: same, kill

antocuni (approved):
@kc611 sorry for taking so long.
The PR is good, just kill the example and we are good to go.

I see a conflict in `_str.spy`: this is very likely caused by PR #678, because while doing it I also run `spy format` on it.
Fixing the conflict should be as easy as running `spy format` on the file on your branch.

antocuni (approved):

PR #691, opened 2026-09-11
Support mod.metafunc() call

paugier:
SPy fails to call a metafunc like this `mymod.foo(5)`.

Using `W_Func.op_METACALL`  in module.py fixes this issue.


antocuni (approved):
LGTM.
Should we also do the same for `@blue.generic` functions?

paugier:
I added another failing test for generic functions called as `mod.func[i32](...)`.

module `__getattribute__` as a metafunc fixes the test. Is it correct?

PR #692, opened 2026-09-11
testing: SPyTuple in Python

paugier:
This improves the usability for testing of tuples returned in Python.

It would be convenient for SIMD testing (and much cleaner than what I did in my related PRs).

Would it be acceptable to have that?

Note that to support all backends, it's easier to define SPyTuple in spy/vm/struct.py (i.e. next to `UnwrappedStruct`).

### DIFF
--- a/spy/tests/compiler/test_struct.py
+++ b/spy/tests/compiler/test_struct.py
@@ -8,7 +8,7 @@ from spy.vm.b import B
 from spy.vm.modules.unsafe import UNSAFE
 from spy.vm.object import W_Type
 from spy.vm.registry import ModuleRegistry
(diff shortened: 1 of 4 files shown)


antocuni (commented):
I don't dislike the general idea.
But I wonder whether we actually _need_ the extra complexity of introducing `SPyTuple`: maybe we could just return a plain python tuple, without bothering to pretend to be an `UnwrappedStruct`?

After all, we already do a similar thing for `list[i32]`, `list[str]` and `dict[i32, i32]`

paugier:
Indeed it is simpler and better without `SPyTuple` and with just `tuple`. I had to slightly adapt two tests.

PR #695, opened 2026-09-11
Add conda based non-static bdw-gc support during compiler initialization

kc611:
When trying to install `spy` using `conda` on `macOS`. I ran into problems with `bwd-gc` because the path for `bwd-gc` lives in the `conda` environment so `cc` couldn't find `gc.h` (nor `-lgc` at link time).

Steps I followed (in `spy` root directory): 

```
conda create env -n spylang
conda activate spylang
pip install -e .
make -C spy/libspy
```

Hence I had to use the following fix so that `bwd-gc` could be detected. If it seems useful please do consider adding this fix to `spy`.

### DIFF
--- a/spy/build/config.py
+++ b/spy/build/config.py
@@ -1,3 +1,4 @@
+import os
 import shutil
 import subprocess
 import sys
@@ -126,9 +127,19 @@ class CompilerConfig:
                 self.ldflags += ["-L", f"{gc_prefix}/lib", "-lgc"]
             else:
                 self.ldflags += ["-lgc"]
-                # On macOS, Homebrew installs bdw-gc outside the default
-                # compiler search paths
-                if sys.platform == "darwin" and shutil.which("brew"):
+                conda_prefix = os.environ.get("CONDA_PREFIX")
+                if conda_prefix and os.path.exists(
+                    os.path.join(conda_prefix, "include", "gc.h")
+                ):
+                    # bdw-gc installed in a conda environment
+                    self.cflags += ["-I", f"{conda_prefix}/include"]
+                    self.ldflags += ["-L", f"{conda_prefix}/lib"]
+                    # conda-forge libgc has an @rpath install name: without an
+                    # rpath entry the executable won't find it at runtime
+                    self.ldflags += [f"-Wl,-rpath,{conda_prefix}/lib"]
+                elif sys.platform == "darwin" and shutil.which("brew"):
+                    # On macOS, Homebrew installs bdw-gc outside the default
+                    # compiler search paths
                     prefix = subprocess.run(
                         ["brew", "--prefix", "bdw-gc"],
                         capture_output=True,


antocuni (commented):
thank your for investigating this.

So, from what I understand the problem is that libraries installed in a conda env are not automatically visible to the compiler, so we need to pass the right flags to locate it.

That's fine but it seems a more general fix for all conda-managed libraries, so probably we need to do it unconditionally, not only for gc.h.

But now my question is different: why does it work at all in linux then?
It's totally possible that it works by chance on my machine, and it picks up the system-installed libgc.

paugier:
Did you try with Pixi (also conda based), which should officially be supported?

Note that we have in pyproject.toml

```toml
[tool.pixi.activation.env]
# to be able to link libgc (bdw-gc)
C_INCLUDE_PATH = "$PIXI_PROJECT_ROOT/.pixi/envs/default/include:$C_INCLUDE_PATH"
LIBRARY_PATH = "$PIXI_PROJECT_ROOT/.pixi/envs/default/lib:$LIBRARY_PATH"
# only useful for macOS
DYLD_LIBRARY_PATH = "$PIXI_PROJECT_ROOT/.pixi/envs/default/lib:$DYLD_LIBRARY_PATH"
```

antocuni:
(for reference, Kaustubh is an Anaconda colleague and we hit this problem when I was trying to onboard him).

> Did you try with Pixi (also conda based), which should officially be supported?

@paugier that was also my first reaction: we test pixi, why it doesn't work?
It turns out that at least on mac, this doesn't seem to work; this happens on both my Mac and @kc611 Mac (using pixi):

```
❯ spy build -x examples/hello.spy
[debug] examples/build/hello
dyld[77840]: Library not loaded: @rpath/libgc.1.dylib
  Referenced from: <0BAAF8DB-04AA-3A4B-BB13-997425F753F5> /Users/antocuni/anaconda/spy/examples/build/hello
  Reason: no LC_RPATH's found
```

What I haven't understood is whether this is a mac-specific problem, or it's also a linux problem which for some reason it's not triggered.
Do we test `build -x` in pixi tests?

paugier:
We do test `build -x` with Pixi (https://github.com/spylang/spy/blob/main/.github/workflows/pixi-checks.yml#L56) on both Linux and macOS. In the CI, libgc is provided by the conda-forge package.

antocuni:
> We do test `build -x` with Pixi (https://github.com/spylang/spy/blob/main/.github/workflows/pixi-checks.yml#L56) on both Linux and macOS. In the CI, libgc is provided by the conda-forge package.

I did some investigation with claude: `hello.spy` doesn't do any GC call, gcc defaults to `--as-needed` and thus doesn't even try to link`libgc`.
I pushed 2fde439f which disables the fix and tries to improve the smoke test, let's see if it fails

antocuni:
so, pixi tests still pass. Claude theory was wrong.
I think that the fix is correct, but I'd like to have a failing test first, else we will run into the same issue again eventually.
I don't have access to my mac now.
@kc611 are you able to investigate why pixi works on CI but not on your machine and to push a test which fails in CI before applying your fix?

antocuni:
@kc611 thank you for the fix! I edited the logic slightly, add a pixi test which actually fails without the fix and add some comments but the actual fix is still yours :pray: .

@paugier I had to solve a conflict with #699 , and while I was at it I took the liberty to improve the explanation of _why_ `LDFLAGS` must be last.

Merging!

PR #697, opened 2026-09-11
Add variadic construction for slice

filiplajszczak:
Accept the one-, two-, and three-argument `slice` forms supported by CPython.

Normalize `slice(stop)` to omitted start and step values, and `slice(start, stop)` to an omitted step, while preserving the existing specialization for integer and `None` components.


antocuni (approved):
LGTM.

PR #700, opened 2026-09-12
Upgrade development and runtime dependencies

filiplajszczak:
Update Wasmtime from 8.0.1 to 48.0.0 and refresh the Ninja, pytest, pre-commit, Pyodide build, CFFI, setuptools, and types-pexpect dependencies.


paugier:
Note that Wasmtime 48.0.0 is the new LTS version (https://docs.wasmtime.dev/stability-release.html), so it really makes sense to update!

Just a remark: I'm a bit worried about potential malicious changes in such PRs (big changes in lock files with a lot of hashes and addresses). I quickly checked by eyes and I didn't see anything suspicious but of course I didn't check the hashes and it's easy to miss a wrong address. I don't think it is a real problem for this specific PR but just sharing this concern.

PR #701, opened 2026-09-14
Add support for <, >, <= and >= operators for string and bytes types

kc611:
As titled, 

This PR adds support for <, >, <= and >= operators for bytes and string types


antocuni (approved):
LGTM.

@kc611 same comment as on #687 : the _str.spy conflict is likely caused by `spy format`, running it should be enough to fix.

